PR #224, opened 2022-10-07
add responsibleai text and vision images to azureml-assets

imatiach-msft:
Add image files for ResponsibleAI Text and Vision scenarios for large language models (ie GPT3) and computer vision models.
Also adds tests which use RAITextInsights and RAIVisionInsights in the environment definitions.

### DIFF
--- a/responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/tests/src/main.py
+++ b/responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/tests/src/main.py
@@ -1,0 +1,107 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Validate running RAITextInsights using responsibleai text docker image."""
+
+import pandas as pd
+
+import datasets
+import zipfile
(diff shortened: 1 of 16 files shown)


responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/tests/src/main.py
```diff
@@ -0,0 +1,107 @@
...
+    train_data = pd_data[NUM_TEST_SAMPLES:]
+    test_data = pd_valid_data[:NUM_TEST_SAMPLES]
+
+    model = retrieve_dbpedia_model()
+
+    # load the model and tokenizer
+    tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
+
+    device = -1
+    if device >= 0:
```
tongyu-microsoft, line 77: where do we set `device` to be greater than 0? Will this be added later?

imatiach-msft: yes we will add it once we have GPU images, so test can be run for both

PR #233, opened 2022-10-18
add text and vision dpv2 components, jobs and tests to azureml-assets

imatiach-msft:
Add text and vision dpv2 components, jobs and tests to azureml-assets.

This includes:
1.) The RAI Text Insights component and the DBPedia dataset for multiclass text classification
2.) The RAI Vision Insights component and the Fridge dataset for multiclass image classification

This is required to onboard to the azureml preview registry for RAI text and vision components.

### DIFF
--- a/responsibleai/vision/components/rai_vision_insights/spec.yaml
+++ b/responsibleai/vision/components/rai_vision_insights/spec.yaml
@@ -1,0 +1,54 @@
+$schema: http://azureml/sdk-2-0/CommandComponent.json
+name: rai_vision_insights
+display_name: RAI Vision Insights
+version: 0.0.1.preview
+type: command
+inputs:
+  task_type:
+    type: string
+    enum: ['image_classification']
+  model_input: # mlflow model name:version
+    type: mlflow_model
+    optional: false
+  model_info:
+    type: string # model name:version
+    optional: false
+  train_dataset:
+    type: mltable
+  test_dataset:
+    type: mltable
+  target_column_name:
+    type: string
+  maximum_rows_for_test_dataset:
+    type: integer
+    default: 5000
+  classes:
+    type: string # Optional[List[str]]
+    default: '[]'
+  precompute_explanation:
+    type: boolean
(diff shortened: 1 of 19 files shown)


responsibleai/vision/jobs/prepare.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from azure.ai.ml.entities import AmlCompute, PipelineJob
+from azure.ai.ml import dsl, load_component, MLClient
+from azure.identity import DefaultAzureCredential
+from azureml.core import Workspace
```
kicha0, line 7: I think azureml.core does not exist in the test environment, you could either find dpv2 equivalent or update the requirments.txt (can be done after merging this)

https://msdata.visualstudio.com/DefaultCollection/Vienna/_git/azureml-asset?path=/registry/requirements.txt

responsibleai/vision/components/rai_vision_insights/spec.yaml
```diff
@@ -0,0 +1,54 @@
+$schema: http://azureml/sdk-2-0/CommandComponent.json
+name: rai_vision_insights
+display_name: RAI Vision Insights
+version: 2
```
kicha0, line 4: for component in private / public preview, should we use semvar versioning and add preview at the end, for example 0.0.1.preview

kicha0 (commented):
please update versioning of the component.

responsibleai/vision/components/rai_vision_insights/spec.yaml
imatiach-msft: done!

responsibleai/vision/jobs/prepare.py
imatiach-msft: I added some try/except logic, this should not be an issue in the builds

kicha0 (approved):

PR #251, opened 2022-10-27
Adding Model Support to asset_publish.py

haritha-bendapudi:
Adding Model Publish Support to asset_publish.py.

Model are saved as folder structure:

-model
    -model.yml
    -asset.yml
    -payload.json
    -output_schema.json
We fetch information from asset.yml and model.yml to clone the model from github , from a commit, and convert it into a mlflow model. The model is then pushed to registry.
Folder model_publish_utils in release directory contains the script to fetch, convert and lastly delete the model files saved to disk


haritha-bendapudi:
@praveenmathamsetty , @adurbhaka , @DHRUV-EDDI , @susridhar , @elesel , @quchuyuan

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -123,12 +124,19 @@ def _str2bool(v: str) -> bool:
                                                              asset_name=asset.name,
                                                              version=final_version)
         # Handle specific asset types
-        if asset.type in [assets.AssetType.COMPONENT, assets.AssetType.ENVIRONMENT]:
+        if asset.type in [assets.AssetType.COMPONENT, assets.AssetType.ENVIRONMENT, assets.AssetType.MODEL]:
+
+            if asset.type == assets.AssetType.MODEL:
+                model_publish_utils = model_publish_utils(str(asset.spec_with_path))
```
elesel, line 130: Please preserve `Path` objects where possible. String-based path handling isn't fun.

elesel (changes requested):
I know this is a draft PR but thought I'd provide some comments at this point.

scripts/requirements.txt
```diff
@@ -1,3 +1,12 @@
 GitPython==3.1.27
 pyyaml==5.4
-requests==2.27.1
\ No newline at end of file
+requests==2.27.1
+tqdm==4.63.0
+azure-ai-ml
+azure-identity
+transformers[torch]
+wget
```
elesel, line 8: Where is `wget` used?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
```diff
@@ -0,0 +1,116 @@
...
+    def _get_model_task_name(self) -> str:
+        with open(self.asset_yaml_path) as f:
+            asset_file = yaml.safe_load(f)
+        return asset_file["task"]
+
+    @classmethod
+    def _git_clone_model(self) -> None:
+        if self.validate_model_and_task():
+            cmd = f'git clone {self.model_uri} {self.model_dir_name}'
+            os.system(cmd)
```
elesel, line 60: Please use `subprocess.run` instead.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -123,12 +124,19 @@ def _str2bool(v: str) -> bool:
                                                              asset_name=asset.name,
                                                              version=final_version)
         # Handle specific asset types
-        if asset.type in [assets.AssetType.COMPONENT, assets.AssetType.ENVIRONMENT]:
+        if asset.type in [assets.AssetType.COMPONENT, assets.AssetType.ENVIRONMENT, assets.AssetType.MODEL]:
```
wayliums, line 127: since model is already being handled differently now, can we take it out and have another ifelse?

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -151,6 +159,11 @@ def _str2bool(v: str) -> bool:
             if results.returncode != 0:
                 logger.log_warning(f"Error creating {asset}")
                 failure_list.append(asset)
+            
+            if asset.type == assets.AssetType.MODEL:
+                model_publish_utils.delete_model_artifact()
```
wayliums, line 164: can you add comment why is this needed?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
```diff
@@ -0,0 +1,116 @@
...
+                os.chdir(f'{self.model_dir_name}')
+                os.system(f'git reset --hard {self.model_commit}')
+
+    # Convert Model to ISV FLavoured MLFlow model and save it.
+    @classmethod
+    def _save_mlflow_model(self):
+        config = AutoConfig.from_pretrained(self.model_name)
+        misc_conf = {"task_type": self.model_task_name}
+        task_model_mapping = {
+            "multiclass": AutoModelForSequenceClassification,
```
wayliums, line 71: these are too specific to your scenario, can you make it generic? This toolset directory will be used by lots of folks (both internal and external) to reuse and see how to publish assets, we don't want **any** asset specific content here.

scripts/requirements.txt
```diff
@@ -1,3 +1,12 @@
...
\ No newline at end of file
+requests==2.27.1
+tqdm==4.63.0
+azure-ai-ml
+azure-identity
+transformers[torch]
+wget
+pandas
+mlflow
+https://scorestorageforgeneric.blob.core.windows.net/libs/azureml_metrics-0.1.0.72521629-py3-none-any.whl
```
wayliums, line 11: This repo is public facing, are we ok sharing these? who's going to update these?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
wayliums: for example, in code here, you should handle conversion in a generic way, the yaml config could specify the transformation needed

scripts/requirements.txt
haritha-bendapudi: Shipra's team owns these wheel files. They are planning on pushing these to pypi in the near future. Till then we will manually update these in the requirements.txt as and when we receive updates from Shipra's team

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+                url: https://huggingface.co/bert-base-uncased
+                commit_hash: 5546055f03398095e385d7dc625e636cc8910bf2
+        publish:
+            type: mlflow_model # could be one of (custom_model, mlflow_model, triton_model)
+            flavor: hftransformers # flavors should be specificed only for mlflow_model and model would be published in this flavor
+            tags: # Key value pairs which will be published as tags in the registry 
+                isv : huggingface
+                task: fill-mask
+    """
+
```
elesel, line 174: Create `__init__` and `_validate` functions similar to `EnvironmentConfig`.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+    @property
+    def name(self) ->str: 
+        return self._yaml.get("name")
+
+    @property
+    def path(self) -> object:
```
elesel, line 180: Consider returning a `Path` object unless `path` is a dict. In the latter case, return `None` since other properties are used to retrieve `url` and `commit_hash`. Look at `EnvironmentConfig.image_*` functions.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self._yaml.get("path")
+
+    @property
+    def package(self) -> Dict[str,str]:
```
elesel, line 184: Make private (`_package`) as a helper function used by `package_url` and `package_commit_hash`.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self.path.get("package") if type(self.path()!="str") else None
+        
+    @property
+    def commit_hash(self) -> str:
```
elesel, line 188:
```suggestion
    def package_commit_hash(self) -> str:
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self.package.get("commit_hash") if self.package() !=None else None
+
+    @property
+    def url(self) -> str:
```
elesel, line 192:
```suggestion
    def package_url(self) -> str:
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self.package.get("url") if self.package() !=None else None
+
+    @property
+    def publish(self) -> Dict[str,object]:
```
elesel, line 196: Make this private (`_publish`) since it shouldn't be used outside of the class.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self._yaml.get("publish")
+
+    @property
+    def type(self) -> str:
```
elesel, line 200: Make an `Enum` for this.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self.publish.get("type")
+    
+    @property
+    def flavor(self) -> str:
```
elesel, line 204: Make an `Enum` for this too.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+        return self.publish.get("flavor") if self.publish.get("flavor") else None
+
+    @property
+    def tags(self) -> Dict[str,str]:
+        return self.publish("tags") if self.publish.get("tags") else {}
```
elesel, line 209:
```suggestion
        return self.publish.get("tags", {})
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+            type: mlflow_model # could be one of (custom_model, mlflow_model, triton_model)
+            flavor: hftransformers # flavors should be specificed only for mlflow_model and model would be published in this flavor
+            tags: # Key value pairs which will be published as tags in the registry 
+                isv : huggingface
+                task: fill-mask
+    """
+
+    @property
+    def name(self) ->str: 
+        return self._yaml.get("name")
```
elesel, line 177: Please use single quotes around key values. Also applies to the rest of the changes in this PR.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+    @property
+    def flavor(self) -> str:
+        return self.publish.get("flavor") if self.publish.get("flavor") else None
+
+    @property
+    def tags(self) -> Dict[str,str]:
+        return self.publish("tags") if self.publish.get("tags") else {}
+    
+    @property
+    def task_name(self) -> str:
```
elesel, line 212: Would you be able to add an example for this to the class?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+            package: # if package not specificed , path would be be used like os path object.
+                url: https://huggingface.co/bert-base-uncased
+                commit_hash: 5546055f03398095e385d7dc625e636cc8910bf2
+        publish:
+            type: mlflow_model # could be one of (custom_model, mlflow_model, triton_model)
+            flavor: hftransformers # flavors should be specificed only for mlflow_model and model would be published in this flavor
+            tags: # Key value pairs which will be published as tags in the registry 
+                isv : huggingface
+                task: fill-mask
+    """
```
elesel, line 173: Use similar indentation style as other classes in this file.
```suggestion
    Example:

    name: bert-base-uncased
    path: # should be string or should contain package details
        package: # if package not specificed , path would be be used like os path object.
            url: https://huggingface.co/bert-base-uncased
            commit_hash: 5546055f03398095e385d7dc625e636cc8910bf2
    publish:
        type: mlflow_model # could be one of (custom_model, mlflow_model, triton_model)
        flavor: hftransformers # flavors should be specificed only for mlflow_model and model would be published in this flavor
        tags: # Key value pairs which will be published as tags in the registry 
            isv : huggingface
            task: fill-mask
    """
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +154,139 @@ def release_paths(self) -> List[Path]:
...
+    @property
+    def tags(self) -> Dict[str,str]:
+        return self.publish("tags") if self.publish.get("tags") else {}
+    
+    @property
+    def task_name(self) -> str:
+        return self.tags.get("task") if self.tags.get("task") else None
+
+    @property
+    def model_dir(self) -> str:
```
elesel, line 216: Please use `Path` objects instead of strings for file/directories, and move this function back to `asset_publish.py`.

scripts/requirements.txt
```diff
@@ -1,3 +1,15 @@
...
+https://scorestorageforgeneric.blob.core.windows.net/libs/azureml_evaluate_mlflow-0.1.0.75053207-py3-none-any.whl
+
+
+
+
+
+
```
elesel, line 15:
```suggestion
https://scorestorageforgeneric.blob.core.windows.net/libs/azureml_evaluate_mlflow-0.1.0.75053207-py3-none-any.whl
```

elesel (changes requested):
Thanks for starting `ModelConfig`. Please see my comments for additional changes required and let me know if you have any questions.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -5,7 +5,20 @@
...
+from mlflow.models import ModelSignature
+from sqlalchemy import true
+from transformers import  AutoTokenizer, AutoConfig, pipeline
+from transformers import (
+    AutoModelForSequenceClassification,
+    AutoModelForMaskedLM,
+    AutoModelForTokenClassification,
+    AutoModelForQuestionAnswering,
+    AutoModelWithLMHead
+)
```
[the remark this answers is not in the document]
haritha-bendapudi: instead of adding these functions to asset_publish.py directly, I plan to have a separate class for these utilities and call that function from asset_publish.py. Is that fine?

scripts/azureml-assets/azureml/assets/config.py
haritha-bendapudi: Moved to asse_publish.

scripts/azureml-assets/azureml/assets/config.py
haritha-bendapudi: We have seperated path into local_path and package

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
haritha-bendapudi: Extracted it to a new class called model_publsih_utils

scripts/requirements.txt
```diff
@@ -1,3 +1,12 @@
...
+requests==2.27.1
+tqdm==4.63.0
+azure-ai-ml
+azure-identity
+transformers[torch]
+wget
+pandas
+mlflow
+https://scorestorageforgeneric.blob.core.windows.net/libs/azureml_metrics-0.1.0.72521629-py3-none-any.whl
+https://scorestorageforgeneric.blob.core.windows.net/libs/azureml_evaluate_mlflow-0.1.0.74451692-py3-none-any.whl
```
skasturi: Could we please not use static package urls here? We could expose a feed or wait till this gets pushed to PyPi? 

@ShipraJain01 What do you think?

skasturi (changes requested):
Changed requested

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +142,76 @@ def release_paths(self) -> List[Path]:
...
+        local_path: "../models/bert-base-uncased" # the local path to the model
+        package:
+            url: https://huggingface.co/bert-base-uncased
+            commit_hash: 5546055f03398095e385d7dc625e636cc8910bf2
+    publish:
+        type: mlflow_model # could be one of (custom_model, mlflow_model, triton_model)
+        flavor: hftransformers # flavor should be specificed only for mlflow_model
+        tags: # tags published to the registry
+            isv : huggingface
+            task: fill-mask
```
skasturi, line 161: I assume this aligns with the spec we discussed? 

@novaturient95

haritha-bendapudi: This is the asset.yaml file

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +142,76 @@ def release_paths(self) -> List[Path]:
...
+    def path_local_path(self) -> object:
+        return self._path.get('local_path')
+
+    @property
+    def _path_package(self) -> Dict[str,str]:
+        return self._path.get("package") if type(self._path()!="str") else None
+        
+    @property
+    def package_commit_hash(self) -> str:
+        return self._path_package.get("commit_hash") if self._path_package() !=None else None
```
skasturi, line 188: >!=N [](http://example.com/codeflow?start=77&length=3)

(nit) spacing

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
```diff
@@ -0,0 +1,116 @@
...
+        if self.validate_model_and_task():
+            cmd = f'git clone {self.model_uri} {self.model_dir_name}'
+            os.system(cmd)
+            if self.model_commit:
+                os.chdir(f'{self.model_dir_name}')
+                os.system(f'git reset --hard {self.model_commit}')
+
+    # Convert Model to ISV FLavoured MLFlow model and save it.
+    @classmethod
+    def _save_mlflow_model(self):
```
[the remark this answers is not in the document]
haritha-bendapudi: @adurbhaka

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -141,6 +142,76 @@ def release_paths(self) -> List[Path]:
...
+    def _path(self) -> object:
+        return self._yaml.get("path")
+
+    @property
+    def path_local_path(self) -> object:
+        return self._path.get('local_path')
+
+    @property
+    def _path_package(self) -> Dict[str,str]:
+        return self._path.get("package") if type(self._path()!="str") else None
```
skasturi, line 184: >)!=" [](http://example.com/codeflow?start=60&length=4)

Could you please run black on the file?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    AutoModelForMaskedLM,
+    AutoModelForTokenClassification,
+    AutoModelForQuestionAnswering,
+    AutoModelWithLMHead
+)
+
+class MLFlowModelUtils:
+
+    def __init__(self, name, task_name, model_dir):
+        self.name = name
```
skasturi, line 21: @wayliums @elesel Wondering if we are using type hints in this repo?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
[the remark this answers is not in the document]
haritha-bendapudi: During our last conversation with Shipra and team, we got to know that pushing packages, metrics, evaluate-flow to PyPi would take a lot of effort and would be targeted only after private preview. So, till they are pushed to public, we have decided to use the packages from storage. We will be updating them as and when we recieve new versions.

scripts/requirements.txt
haritha-bendapudi: Same as above.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_utils.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation.
```
skasturi, line 1: (nit) rename to utils.py since model is already captured in the parent directory name?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) Microsoft Corporation.
```
skasturi, line 1: (nit) rename to mlflow_utils.py?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+# Licensed under the MIT License.
+
+import logging
+import os
+import azureml.evaluate.mlflow as mlflow
```
skasturi, line 6: >mlflow [](http://example.com/codeflow?start=34&length=6)

(nit) rename to azureml_mlflow to avoid conflation with native mlflow.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+            model = AutoModelWithLMHead.from_pretrained(self.name, config=config)
+        else:
+            logging.error("Invalid Task Name")
+        tokenizer = AutoTokenizer.from_pretrained(self.name, config=config)
+        sign_dict = {"inputs": '[{"name": "input_string", "type": "string"}]', "outputs": '[{"type": "string"}]'}
+        if self.task_name == "question-answering":
+            sign_dict["inputs"] = '[{"name": "question", "type": "string"}, {"name": "context", "type": "string"}]'
+        signature = ModelSignature.from_dict(sign_dict)
+        self.mlflow_model_dir = self.name + "-mlflow"
+        mlflow.hftransformers.save_model(model, f"{self.mlflow_model_dir}", tokenizer, config, misc_conf, signature=signature)
```
skasturi, line 51: > mlflow.hftransformers.save_model [](http://example.com/codeflow?start=7&length=33)

Could we please add exception handling for this? 

Such as disk space not available, MemoryError etc? 

We are already seeing such issues locally

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+            "ner": AutoModelForTokenClassification,
+            "question-answering": AutoModelForQuestionAnswering,
+            "summarization": AutoModelWithLMHead,
+            "text-generation": AutoModelWithLMHead,
+            "text-classification": AutoModelForSequenceClassification
+        }
+        if self.task_name in task_model_mapping:
+            model = task_model_mapping[self.task_name].from_pretrained(self.name, config=config)
+        elif "translation" in self.task_name:
+            model = AutoModelWithLMHead.from_pretrained(self.name, config=config)
```
skasturi, line 42: @novaturient95 @ghyadav Is this the step where we saw `MemoryError` on the agents?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        tokenizer = AutoTokenizer.from_pretrained(self.name, config=config)
+        sign_dict = {"inputs": '[{"name": "input_string", "type": "string"}]', "outputs": '[{"type": "string"}]'}
+        if self.task_name == "question-answering":
+            sign_dict["inputs"] = '[{"name": "question", "type": "string"}, {"name": "context", "type": "string"}]'
+        signature = ModelSignature.from_dict(sign_dict)
+        self.mlflow_model_dir = self.name + "-mlflow"
+        mlflow.hftransformers.save_model(model, f"{self.mlflow_model_dir}", tokenizer, config, misc_conf, signature=signature) 
+
+    def _convert_to_mlflow_package(self):
+        return None
```
skasturi, line 54: What is this method used for?

scripts/azureml-assets/azureml/assets/config.py
elesel: Note that [Black](https://pypi.org/project/black/) isn't a requirement for the repo, but [flake8](https://flake8.pycqa.org/en/latest/) is used in one of the checks to ensure proper formatting.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
elesel: We are. They're not required, but they certainly help.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
praveenmathamsetty: hf_mlflow

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def _convert_to_mlflow_package(self):
+        return None       
+
+    def covert_into_mlflow_model(self):
+        if self.flavor == "hftransformers":
+            self._convert_to_mlflow_hftransformers()
+        #TODO add support for pyfunc. Pyfunc requires custom env file.
+        else :
+            self._convert_to_mlflow_package()
```
praveenmathamsetty, line 61: Currently only two flavors supported pyfunc & HF. In case of other flavor we need to throw an exception.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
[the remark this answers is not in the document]
ShipraJain01: @skasturi : Haritha's point is correct. We are in process of publishing azureml-metrics as py-pi (with internal consumption only but that would be post PrP. We arent planning to publish azureml-evaluate-mlflow to py-pi rite now since the changes within are strictly internal.

Please let us know if you see this as blocker.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
ShipraJain01: Since azureml.mlflow is Py-Pi released packaged for external consumption, we cant merge the changes of azureml.evaluate.mlflow to the package since changes within are strictly internal. Also we are working on Pyfunc proposal for OSS based on which next steps would be clear on enabling this package as part of azureml-mlflow or thru OSS mlflow package directly.

elesel:
@haritha-bendapudi please merge the latest from the main branch into yours to pick up all of the documentation I added to config.py. It looks like the automated checks are complaining to you about tech debt I created!

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -151,6 +191,43 @@ def _str2bool(v: str) -> bool:
...
+                "--tags", model_config.tags
+                ]
+            if resource_group:
+                cmd.extend(["--resource-group", resource_group])
+            if workspace:
+                cmd.extend(["--workspace", workspace])
+            print(cmd)
+
+            # Run command
+            if debug_mode:
```
wayliums, line 215: for this part can you refactor to a common code?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
haritha-bendapudi: As of now we only support hf_flavor mlflow transform. We could add pyfunc, but the constraint with Pyfunc is that we would be dealing with only huggingface with pyfunc, not generic pyfunc. This is because pyfunc expects the environment to be defined for building the wrapper. So, if a user wants to convert a model to mlflow using pyfunc, he would have to provide the necessary packages as well, which we donot support as of now

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
novaturient95: yes. In case of disk space related errors, it fails while downloading/cloning model

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -52,6 +52,45 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+    """
+    Prepares the model. Downloads the model if required and converts the models to specified
+    publish type.
+
+    Return: returns the local path to the model.
+    """
+    if (model_config.path_local_path):
+        return None
+
+    model_dir = '/tmp/' + model_config.name
```
novaturient95, line 73: not sure if this is handled, agent storage will not be sufficient for models we are going to download.
@wayliums @elesel can you please share specs for agents we are using currently for component and environment. I propose using a different pool of agents for model publishing

@skasturi

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -52,6 +52,45 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+    model_utils = ModelUtils(model_url= model_config.package_url, model_commit_hash= model_config.package_commit_hash, model_dir=model_dir)
+    model_utils.download_model()
+    update_spec_file(spec_file, model_dir)
+    if model_config.type.value is 'mlflow_model':
+        mlflow_dir = model_config.name  +'mlflow'
+        mlflow_utils = MLFlowModelUtils(name = model_config.name, task = model_config.task_name, flavor = model_config.flavor.value, model_dir = model_dir)
```
novaturient95, line 79: nit: utils classes should generally be static

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -151,6 +191,43 @@ def _str2bool(v: str) -> bool:
...
                 logger.log_warning(f"Error creating {asset}")
                 failure_list.append(asset)
+            
+        elif asset.type == assets.AssetType.MODEL:
+
+            model_config = asset.model_config_as_object()
+            model_dir = model_prepare(model_config, asset.spec_with_path)
+            # Assemble command
+            cmd = [
+                shutil.which("az"), "ml", asset.type.value, "create",
```
novaturient95, line 201: @skasturi I think this is where you mentioned we should have favors as per @praveenmathamsetty suggestion.
@haritha-bendapudi can this be incorporated?

novaturient95: Also, how we are enforcing semver here?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    def _convert_to_mlflow_hftransformers(self):
+        config = AutoConfig.from_pretrained(self.name)
+        misc_conf = {"task_type": self.task_name}
+        task_model_mapping = {
+            "multiclass": AutoModelForSequenceClassification,
+            "multilabel": AutoModelForSequenceClassification,
+            "fill-mask": AutoModelForMaskedLM,
+            "ner": AutoModelForTokenClassification,
+            "question-answering": AutoModelForQuestionAnswering,
+            "summarization": AutoModelWithLMHead,
```
novaturient95, line 35: for `AutoModelWithLMHead`, we get below deprecation warning: 

```
/home/azureuser/miniconda3/envs/mp/lib/python3.8/site-packages/transformers/models/auto/modeling_auto.py:806: FutureWarning: The class `AutoModelWithLMHead` is deprecated and will be removed in a future version. Please use `AutoModelForCausalLM` for causal language models, `AutoModelForMaskedLM` for masked language models and `AutoModelForSeq2SeqLM` for encoder-decoder models.
```

can we just go with just with [AutoModel](https://huggingface.co/transformers/v3.0.2/model_doc/auto.html) for all task types? 

cc: @ghyadav

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        elif "translation" in self.task_name:
+            model = AutoModelWithLMHead.from_pretrained(self.name, config=config)
+        else:
+            logging.error("Invalid Task Name")
+        tokenizer = AutoTokenizer.from_pretrained(self.name, config=config)
+        sign_dict = {"inputs": '[{"name": "input_string", "type": "string"}]', "outputs": '[{"type": "string"}]'}
+        if self.task_name == "question-answering":
+            sign_dict["inputs"] = '[{"name": "question", "type": "string"}, {"name": "context", "type": "string"}]'
+        signature = ModelSignature.from_dict(sign_dict)
+        self.mlflow_model_dir = self.name + "-mlflow"
```
novaturient95, line 50: I name going to contain `/`? For eg. in case of `microsoft/deberta-base`.
Or are publishers expected to share valid model names? Should we add validations for spec info as well?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class ModelUtils:
+    def __init__(self, model_url, model_commit_hash, model_dir):
+        self.model_url = model_url
+        self.model_commit_hash = model_commit_hash
+        self.model_dir = model_dir
+
+    def download_model(self) -> None:
+        cmd = f"git clone {self.model_url} {self.model_dir}"
```
skasturi, line 14: > cmd = f"git clone {self.model_url} {self.model_dir}" [](http://example.com/codeflow?start=7&length=53)

How do we ensure that LFS is enabled?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
susridhar: The output of an AutoModel will be raw hidden states without any specific head on top. Our current scoring script uses HF pipelines, which need a specialized model(with the appropriate head); they don't work with AutoModels. I think @ghyadav's library also needs a model with a specialized head, as a few of the tasks are implemented using pipelines. We'll need to go ahead with this approach itself.
As for the warning, I think it can be avoided by using AutoModelForSeq2SeqLM/AutoModelForCausalLM instead of AutoModelWithLMHead. We'll test that change out and update the PR once it's done.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_utils.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import logging
+import azureml.evaluate.mlflow as hf_mlflow
```
skasturi, line 5: (nit) Usual order or imports: 

<System imports>

<Third party package imports>

<Internal package imports>

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -0,0 +1,21 @@
...
+        run(cmd, shell=True)
```
skasturi, line 15: >run(cmd, shell=True) [](http://example.com/codeflow?start=8&length=20)

Exception handling.

scripts/requirements.txt
skasturi: Could we please split this PR into two? 

1. Supporting Model publishing
2. MLFlow HF flavor?

@novaturient95

skasturi:
MLFlow HF flavor is not yet pushed to PyPi and only then we can officially include in this repo.

Therefore, could we please split this PR into two?

1. Supporting Model publishing
2. MLFlow HF flavor?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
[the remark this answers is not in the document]
skasturi: Thanks @haritha-bendapudi , @ShipraJain01. 

I don't consider this a blocker. I am thinking that we should separate out the public and non-public stuff and create separate PRs for them. 

Model configuration etc can be part of this PR. HF Flavor in a separate PR & we keep iterating based on dev indices till the package pushed to PyPi. 

@wayliums , @elesel

scripts/azureml-assets/azureml/assets/release/asset_publish.py
elesel: Please ask this question via Teams and we'll discuss there.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/model_publish_utils.py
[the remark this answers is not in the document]
haritha-bendapudi: Made changes to this PR to make it model publish without HFTransformers flavor.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_model_utils.py
haritha-bendapudi: when the name contains `/`, it only affects the az ml model create step. The error thrown when spec.yaml has / in the name is in accordance to the error one faces when locally creating a model using az cli. Since we are following az-cli schema for the spec.yaml file, the user should make sure that their spec file is inline with az cli requirements

haritha-bendapudi:
> Thanks for starting `ModelConfig`. Please see my comments for additional changes required and let me know if you have any questions.

Added Additional changes

haritha-bendapudi:
> Thanks for starting `ModelConfig`. Please see my comments for additional changes required and let me know if you have any questions.

Additional Changes added

scripts/azureml-assets/azureml/assets/release/asset_publish.py
haritha-bendapudi: Moving this conversation to teams

inference/mlflow-huggingface/src/models/bert-base-uncased/spec.yaml
```diff
@@ -0,0 +1,3 @@
+name: bert-base-uncased
+path: C:\Users\Administrator\azureml-assets\scripts\azureml-assets\tmp\bert-base-uncased
```
skasturi, line 2: >C:\Users\Administrator\azureml-assets\scripts\azureml-assets\tmp\bert-base-uncased [](http://example.com/codeflow?start=6&length=82)

Could you please put in relative path?

skasturi: Or may be use URL here?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -106,6 +106,8 @@ def _validate_exists(property_name: str, property_value: object):
     def _validate_enum(property_name: str, property_value: object, enum: Enum, required=False):
         """Ensure an enum value is set and is expected.
 
+        "model.flavor", self.flavor, ModelFlavor, True)
```
skasturi, line 109: >"model.flavor", self.flavor, ModelFlavor, True) [](http://example.com/codeflow?start=8&length=47)

Is this the set of enums expected? Could you please elaborate?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -179,42 +182,42 @@ def _validate(self):
         Raises:
             ValidationException: If validation failed.
         """
-        Config._validate_exists('name', self.name)
-        Config._validate_exists('version', self.version)
+        Config._validate_exists("name", self.name)
```
skasturi, line 185: >("n [](http://example.com/codeflow?start=31&length=3)

(nit) These could be in single quotes may be?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -232,36 +235,149 @@ def release_paths(self) -> List[Path]:
         return release_paths
 
 
+class ModelType(Enum):
+    """Enum for the Model Types accepted in ModelConfig."""
+
+    MLFLOW = "mlflow_model"
+    CUSTOM = "custom_model"
+    TRITON = "triton_model"
```
skasturi, line 243: >triton_model [](http://example.com/codeflow?start=14&length=12)

@wayliums @elesel Could you please provide more details on what Triton model is?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -232,36 +235,149 @@ def release_paths(self) -> List[Path]:
...
+
+    MLFLOW = "mlflow_model"
+    CUSTOM = "custom_model"
+    TRITON = "triton_model"
+
+
+class ModelFlavor(Enum):
+    """Enum for the Flavors accepted in ModelConfig."""
+
+    HFTRANSFORMERS = "hftransformers"
```
skasturi, line 249: >HFTRANSFORMERS = "hftransformers [](http://example.com/codeflow?start=4&length=32)

Could we add this in the azureml-asset repo instead?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_utils.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""MLFlow Operations Class."""
+
+
+class MLFlowModelUtils:
```
skasturi, line 6: Could we remove this file since there does not seem to be any functionality?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        return result.returncode
+
+    def _delete_downloaded_artifacts(self) -> None:
+        """Delete the downloaded artifacts in case of failure."""
+        logger.print("Deleting the incomplete model artifacts")
+        delete_cmd = f"rm -rf {self.model_dir}"
+        self._run(delete_cmd)
+
+    def download_model(self) -> bool:
+        """Download the Model."""
```
skasturi, line 45: >Download [](http://example.com/codeflow?start=11&length=8)

Where/How do we capture metadata that the model is available through `git lfs` and not through the usual http download?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        if result.returncode != 0:
+            logger.print(f"Failed with error {result.stdout}.")
+        else:
+            logger.print(f"Successfully executed! Output: \n{result.stdout}")
+        return result.returncode
+
+    def _delete_downloaded_artifacts(self) -> None:
+        """Delete the downloaded artifacts in case of failure."""
+        logger.print("Deleting the incomplete model artifacts")
+        delete_cmd = f"rm -rf {self.model_dir}"
```
skasturi, line 41: > delete_cmd = f"rm -rf {self.model_dir}" [](http://example.com/codeflow?start=7&length=40)

`rm -rf` could have catastrophic effects.

Could we please create a temporary dir and mark it to be deleted instead?

skasturi:
Could you please elaborate on why the folder structure is "inference/mlflow-huggingface/src/models/bert-base-uncased"?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
haritha-bendapudi: We are not currently storing any data regarding the model availability through git lfs/http. We will include it in the next iteration.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
haritha-bendapudi: using temp directory.

scripts/azureml-assets/azureml/assets/config.py
haritha-bendapudi: The double quotes seem to be a formatting change. All the properties in the config file are in double quotes.

scripts/azureml-assets/azureml/assets/config.py
haritha-bendapudi: We should provide support for all the fields in model.yaml file. Removing these would mean we cannot handle 'flavor' field in model.yaml.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/mlflow_utils.py
haritha-bendapudi: It is a failure case for type mlflow_model in model.yaml

inference/mlflow-huggingface/src/models/bert-base-uncased/asset.yaml
```diff
@@ -0,0 +1,5 @@
+name: bert-base-uncased
+version: 1
+type: model
+spec: spec.yaml
+extra_config: model.yaml
```
skasturi, line 5: (nit) add new line at the end

inference/mlflow-huggingface/src/models/bert-base-uncased/spec.yaml
skasturi: Are we going to retain this change?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        self._run(lfs_cmd)
+        clone_cmd = f"git clone {self.model_url} {self.model_dir.name}"
+        result = self._run(clone_cmd)
+        if result != 0:
+            self._delete_downloaded_artifacts()
+            return False
+        if self.model_commit_hash:
+            cmd = f"git reset --hard {self.model_commit_hash}"
+            self._run(cmd, cwd=self.model_dir.name)
+        self._run("rm -rf .git", cwd=self.model_dir.name)
```
skasturi, line 62: >"rm -rf .git [](http://example.com/codeflow?start=18&length=12)

I was suggesting to use the following, @haritha-bendapudi 

[https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)

inference/mlflow-huggingface/src/models/bert-base-uncased/model.yaml
```diff
@@ -0,0 +1,14 @@
+name: bert-base-uncased                                               #Required
+path:                                                                 #Required
+  # local_path: "./local-path-to-model-files"                         #Required  Either package or local_path should be passed. if both are given, local_path is given preference  
+  package:                                                            #Required. Either package or local_path should be passed. if both are given, local_path is given preference
+    commit_hash: 5546055f03398095e385d7dc625e636cc8910bf2             #Optional
+    url: https://huggingface.co/bert-base-uncased                     #Required
```
skasturi, line 6: > url: https://huggingface.co/bert-base-uncased [](http://example.com/codeflow?start=3&length=46)

We need to add metadata here that captures that we need to do `git lfs` and not a http/ftp based download.

@novaturient95 Could you please add in your PR?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
haritha-bendapudi: I have used temporary directory to download the model and clean the files after moving it to the registry. I am using rm -rf .git here due to the following reason: When we download the model files from git, the .git folder is downloaded too. We do not want this to be published to the model registry since it adds to its size, which in turn will result in a bigger SKU for inferencing the model.

inference/mlflow-huggingface/src/models/bert-base-uncased/model.yaml
haritha-bendapudi: We could add that metadata to model.yaml. In that case we can conditionally install LFS. I'll inform my team regarding this and get back to you. All the huggingface stock models require LFS support, hence we went with this generalization.

elesel (changes requested):

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
skasturi: As discussed offline, let us serialize & write the mlflow-model to a sub-directory and push that to the registry rather than having to delete the .git folder.

inference/mlflow-huggingface/src/models/bert-base-uncased/asset.yaml
```diff
@@ -0,0 +1,5 @@
+name: bert-base-uncased
+version: 1
```
skasturi, line 2: >1 [](http://example.com/codeflow?start=9&length=1)

auto

haritha-bendapudi: @adurbhaka should I change this to auto in this iteration?

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -62,25 +147,39 @@ def _str2bool(v: str) -> bool:
...
-    parser.add_argument("-d", "--debug", type=_str2bool, nargs='?', const=True, default=False, help="debug mode")
+    parser.add_argument(
+        "-d",
+        "--debug",
+        type=_str2bool,
+        nargs="?",
+        const=True,
+        default=False,
+        help="debug mode",
+    )
```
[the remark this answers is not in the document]
haritha-bendapudi: @elesel Could you recommend a formatter? I am using BlackFormatter. All the formating changes are a result of that.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -232,36 +233,149 @@ def release_paths(self) -> List[Path]:
...
+
+    @property
+    def name(self) -> str:
+        """Model Name."""
+        return self._yaml.get("name")
+
+    @property
+    def _path(self) -> object:
+        """Model Path."""
+        return self._yaml.get("path")
```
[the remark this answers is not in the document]
haritha-bendapudi: Implemented the change.

inference/mlflow-huggingface/src/models/bert-base-uncased/model.yaml
haritha-bendapudi: Hello Sasidhar. Huggingface document on downloading models says : https://huggingface.co/docs/hub/models-downloading 
```
git lfs install
git clone <MODEL URL>
```
Hence, to simplify this, we can install LFS for all models.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -142,27 +250,57 @@ def _str2bool(v: str) -> bool:
...
-            else:
-                results = run(cmd)
-
-            # Check command result
-            if results.returncode != 0:
-                logger.log_warning(f"Error creating {asset}")
-                failure_list.append(asset)
+            results = run_command(cmd, failure_list, debug_mode)
+
+        elif asset.type == assets.AssetType.MODEL:
```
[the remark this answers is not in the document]
haritha-bendapudi: Since I have to run the run_command within the context_manager for the model, I will create function for preparing and executing the command and run it within the if else.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        """Download the Model."""
+        """
+        Initialize LFS
+        Clones the model from the URL into the Model Directory.
+        Deletes the incomplete model artifact in case of failure.
+        Sets the HEAD to the commit hash.
+        Deletes the .git folder in the downloaded artifacts.
+        """
+        lfs_cmd = "git lfs install"
+        self._run(lfs_cmd)
```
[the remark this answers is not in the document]
haritha-bendapudi: For this to be moved outside the block, considering that multiple models will be published, it should be moved outside the for loop over assets.

scripts/requirements.txt
```diff
@@ -1,3 +1,7 @@
 GitPython==3.1.27
 pyyaml==5.4
-requests==2.27.1
\ No newline at end of file
+requests==2.27.1
+tqdm==4.63.0
+torch==1.12.1
+pandas
+mlflow
```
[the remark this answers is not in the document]
haritha-bendapudi: We do not need these directories for this PR, since we do not have MLFlow support yet. Removing these dependencies for now. For future mlflow model support, we can discuss how to structure the files to reduce unnecessary increase in overall package size

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -40,18 +46,126 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+                          sort_keys=False)
+
+
+def update_spec_file(spec_file: Path, path: Path):
+    """Update the yaml file after getting the model has been prepared."""
+    with open(spec_file) as f:
+        model_file = yaml.safe_load(f)
+    model_file['path'] = path
+    with open(spec_file, "w") as f:
+        yaml.dump(model_file, f)
```
elesel, line 67: You could add this functionality into scripts/azureml-assets/azureml/assets/update_spec.py instead. Again, another PR would be fine too.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -40,18 +46,126 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+    model_config: assets.ModelConfig, spec_file: Path, model_dir: str
+) -> Dict[str, Path]:
+    """Prepare Model."""
+    """
+    Prepare the model. Download the model if required.
+    Convert the models to specified publish type.
+
+    Return: returns the local path to the model.
+    """
+    result = {"download_success": False, "model_dir": None}
```
elesel, line 80: nit:
```suggestion
    result = {'download_success': False, 'model_dir': None}
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -232,6 +232,147 @@ def release_paths(self) -> List[Path]:
...
+
+    @property
+    def _path(self) -> object:
+        """Model Path."""
+        return self._yaml.get('path', {})
+
+    @property
+    def path_local(self) -> Path:
+        """Model Local Path."""
+        return self._path.get('local_path')
```
elesel, line 309: Return an actual Path object:
```suggestion
        local_path = self._path.get('local_path')
        return Path(local_path) if local_path else None
```

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -40,18 +46,126 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+
+    if model_config.path_local:
+        update_spec_file(spec_file, os.path.abspath(
+            model_config.path_local))
```
elesel, line 84:
```suggestion
        update_spec_file(spec_file, model_config.path_local.absolute())
```

inference/mlflow-huggingface/src/models/bert-base-uncased/spec.yaml
elesel: The full path should be removed to minimize confusion. How about this?
```suggestion
path: will-be-replaced-when-model-is-published
```

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -40,18 +46,126 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+    if model_config.type == assets.ModelType.CUSTOM:
+        model_utils = ModelUtils(
+            model_url=model_config.remote_uri,
+            model_commit_hash=model_config.remote_commit_hash,
+            model_download_type=model_config.remote_type.value,
+            model_dir=model_dir,
+        )
+        validate_download = model_utils.download_model()
+        if validate_download:
+            update_spec_file(spec_file, model_dir+"/model_dir")
```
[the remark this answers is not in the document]
haritha-bendapudi: Removed model_dir by using shutil.rmtree to remove .git

inference/mlflow-huggingface/src/models/bert-base-uncased/spec.yaml
elesel: That's fine!

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -232,6 +232,147 @@ def release_paths(self) -> List[Path]:
...
+
+    def __init__(self, file_name: Path):
+        """Initialize object for the Model Properties extracted from extra_config model.yaml."""
+        super().__init__(file_name)
+        self._validate()
+
+    def _validate(self):
+        """Validate the yaml file."""
+        Config._validate_exists('model.path', self._path)
+        Config._validate_exists('model.publish', self._publish)
```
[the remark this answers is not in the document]
haritha-bendapudi: Will add it in a separate PR. Might be adding more fields based on renewed requirements for MLFlow HFflavour

scripts/azureml-assets/azureml/assets/release/asset_publish.py
haritha-bendapudi: Next PR it is.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -40,18 +46,126 @@ def test_files_preprocess(test_jobs, asset_ids: dict):
...
+    with open(spec_file) as f:
+        model_file = yaml.safe_load(f)
+    model_file['path'] = path
+    with open(spec_file, "w") as f:
+        yaml.dump(model_file, f)
+
+
+def model_prepare(
+    model_config: assets.ModelConfig, spec_file: Path, model_dir: Path
+) -> Dict[str, Path]:
```
elesel, line 72: `Path` wouldn't apply to the `download_success` boolean. Also consider raising exceptions when things break instead of passing booleans around.
```suggestion
) -> Dict[str, object]:
```

wayliums (approved):

elesel (approved):
Final nits and then I think it's ready for merge.

elesel (approved):

PR #274, opened 2022-11-18
Support model download from public blobstore endpoints

novaturient95:
Add support for downloading model files from azure blobstore

### DIFF
--- a/.github/actions/create-azure-resources/action.yaml
+++ b/.github/actions/create-azure-resources/action.yaml
@@ -27,7 +27,11 @@ runs:
     - name: Create Azure resources
       shell: bash
       run: bash ${{ inputs.scripts-setup-dir }}/create_azure_resources_daily.sh
-    
+
+    - name: Setup azcopy
+      shell: bash
+      run: bash ${{ inputs.scripts-setup-dir }}/install_azcopy.sh
+
     - name: Create additional environment variables
       shell: bash
       run: echo "subscription_id=${{ inputs.subscription-id }}" >> $GITHUB_ENV
--- a/scripts/azureml-assets/azureml/assets/config.py
+++ b/scripts/azureml-assets/azureml/assets/config.py
@@ -266,6 +266,7 @@ class ModelDownloadType(Enum):
     FTP = 'ftp'
     GIT = 'git'
     HTTP = 'http'
+    AZURE_BLOBSTORE: "azure"
 
 
 class ModelConfig(Config):
--- a/scripts/azureml-assets/azureml/assets/release/asset_publish.py
+++ b/scripts/azureml-assets/azureml/assets/release/asset_publish.py
@@ -95,13 +95,16 @@ def model_prepare(
             update_spec_file(spec_file, model_dir)
 
     elif model_config.type == assets.ModelType.MLFLOW:
-        mlflow_utils = MLFlowModelUtils(
-            name=asset.name,
-            task_name=model_config.task_name.value,
-            flavor=model_config.flavor.value,
-            mlflow_model_dir=model_dir,
+        # TODO: udpate this once we start consuming git based model
+        model_utils = ModelUtils(
+            model_url=model_config.remote_uri,
+            model_commit_hash=model_config.remote_commit_hash,
+            model_download_type=model_config.remote_type,
+            model_dir=model_dir,
         )
-        validate_download = mlflow_utils.convert_into_mlflow_model()
+        validate_download = model_utils.download_model()
+        if validate_download:
+            update_spec_file(spec_file, model_dir)
 
     else:
         print(model_config.type.value, assets.ModelType.MLFLOW)
--- a/scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
+++ b/scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
@@ -71,11 +71,23 @@ class ModelUtils:
         shutil.rmtree(git_path, onerror=_onerror)
         return True
 
+    def _download_blobstore_artifacts(self) -> bool:
+        """Download model files from blobstore"""
+        az_copy_cmd = f"azcopy cp --recursive=true {self.model_url} {self.model_dir}"
+        result = self._run(az_copy_cmd)
+        # exit for non-zero error
+        if result != 0:
+            logger.log_warning(f"Failed to download model files with URL: {self.model_url}")
+            return False
+        return True
+
     def download_model(self) -> bool:
         """Prepare the Download Environment."""
         if self.model_download_type == ModelDownloadType.GIT:
             download_success = self._download_git_model()
             return download_success
+        if self.model_download_type == ModelDownloadType.AZURE_BLOBSTORE:
+            return self._download_blobstore_artifacts()
         else:
             logger.print('Unsupported Model Download Method.')
         return False
--- a/scripts/setup/install_azcopy.sh
+++ b/scripts/setup/install_azcopy.sh
@@ -1,0 +1,7 @@
+echo "Downloadng azcopy ...."
+wget https://aka.ms/downloadazcopy-v10-linux 
+tar -xvf downloadazcopy-v10-linux
+sudo rm /usr/bin/azcopy
+sudo cp ./azcopy_linux_amd64_*/azcopy /usr/bin/
+alias azcopy=/usr/bin/azcopy
+echo "Installed azcopy ...."


.github/actions/create-azure-resources/action.yaml
```diff
@@ -27,7 +27,11 @@ runs:
     - name: Create Azure resources
       shell: bash
       run: bash ${{ inputs.scripts-setup-dir }}/create_azure_resources_daily.sh
-    
+
+    - name: Setup azcopy
+      shell: bash
+      run: bash ${{ inputs.scripts-setup-dir }}/install_azcopy.sh
```
elesel, line 33: I don't think these should be here, but you could add them to .github\workflows\assets-test.yaml if azcopy will be used to support asset testing.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -266,6 +266,7 @@ class ModelDownloadType(Enum):
     FTP = 'ftp'
     GIT = 'git'
     HTTP = 'http'
+    AZURE_BLOBSTORE = 'azure'
```
elesel, line 269: Unless there are other types of Azure storage that we'll need to support, there's no need to distinguish with using the `_BLOBSTORE` suffix. I'd just use `AZURE`.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -67,15 +67,27 @@ def _download_git_model(self) -> bool:
             if commit_exists != 0:
                 # TODO: Error handling incase of incorrect commit hash.
                 logger.log_warning("Commit hash doesn't exist. Using model from latest HEAD.")
-        git_path = os.path.join(self.model_dir, '.git')
+        git_path = os.path.join(self.model_dir, ".git")
         shutil.rmtree(git_path, onerror=_onerror)
         return True
 
+    def _download_blobstore_artifacts(self) -> bool:
```
elesel, line 74: Consider using `_download_azure_artifacts` instead of `_download_blobstore_artifacts`.

scripts/setup/install_azcopy.sh
```diff
@@ -0,0 +1,7 @@
+echo "Downloadng azcopy ...."
+wget https://aka.ms/downloadazcopy-v10-linux 
+tar -xvf downloadazcopy-v10-linux
+sudo rm /usr/bin/azcopy
+sudo cp ./azcopy_linux_amd64_*/azcopy /usr/bin/
+alias azcopy=/usr/bin/azcopy
+echo "Installed azcopy ...."
```
elesel, line 7: Please move this script one directory up. The scripts/setup directory focuses on Azure resources.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -67,15 +67,27 @@ def _download_git_model(self) -> bool:
             if commit_exists != 0:
                 # TODO: Error handling incase of incorrect commit hash.
                 logger.log_warning("Commit hash doesn't exist. Using model from latest HEAD.")
-        git_path = os.path.join(self.model_dir, '.git')
+        git_path = os.path.join(self.model_dir, ".git")
         shutil.rmtree(git_path, onerror=_onerror)
         return True
 
+    def _download_azure_artifacts(self) -> bool:
+        """Download model files from blobstore."""
```
skasturi, line 75: >blobstore [](http://example.com/codeflow?start=37&length=9)

(nit) could this be any url out there?

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -67,15 +67,27 @@ def _download_git_model(self) -> bool:
...
+        az_copy_cmd = f"azcopy cp --recursive=true {self.model_url} {self.model_dir}"
```
skasturi, line 76: > {self.model_dir} [](http://example.com/codeflow?start=67&length=17)

Is this the target registry storage dir?

skasturi:
> class ModelUtils:

Is there a reason this class needs to be stateful? 

---
Refers to: scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py:24 in 5198b05. [](commit_id = 5198b05ad3992cfbcd0ea67d4c4ac1c1929c3065, deletion_comment = False)

skasturi:
>     def _run(self, cmd, cwd: Path = "./") -> int:

Could we please make this a static method? 

---
Refers to: scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py:34 in 5198b05. [](commit_id = 5198b05ad3992cfbcd0ea67d4c4ac1c1929c3065, deletion_comment = False)

scripts/install_azcopy.sh
```diff
@@ -0,0 +1,7 @@
+echo "Downloadng azcopy ...."
```
skasturi, line 1: >Downloadng [](http://example.com/codeflow?start=6&length=10)

Downloading

scripts/install_azcopy.sh
```diff
@@ -0,0 +1,7 @@
...
+wget https://aka.ms/downloadazcopy-v10-linux
```
skasturi, line 2: Add verbose flags?

skasturi: Also, please add a download folder/target file name.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
novaturient95: `self.model_url` is target storge dir. `model_dir` is where it is going to download on the agents

.github/actions/create-azure-resources/action.yaml
novaturient95: Done

scripts/setup/install_azcopy.sh
novaturient95: Done

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
novaturient95: for blobstore: It has to follow: 
https://<storage_account>.blob.core.windows.net/<container>/<path>

scripts/install_azcopy.sh
```diff
@@ -0,0 +1,7 @@
+echo "Downloadng azcopy ...."
+wget https://aka.ms/downloadazcopy-v10-linux 
+tar -xvf downloadazcopy-v10-linux
+sudo rm /usr/bin/azcopy
```
skasturi, line 4: >/usr/bin/ [](http://example.com/codeflow?start=8&length=9)

Why does azcopy need to be in /usr/bin? 

Can we not execute it from the downloaded folder itself?

skasturi: Also, please prefer `/opt/`

scripts/install_azcopy.sh
skasturi: What's the working directory in which this script is run?

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -95,13 +94,16 @@ def model_prepare(
             update_spec_file(spec_file, model_dir)
 
     elif model_config.type == assets.ModelType.MLFLOW:
-        mlflow_utils = MLFlowModelUtils(
-            name=asset.name,
-            task_name=model_config.task_name.value,
-            flavor=model_config.flavor.value,
-            mlflow_model_dir=model_dir,
+        # TODO: udpate this once we start consuming git based model
+        model_utils = ModelUtils(
```
skasturi, line 98: >ModelUtils [](http://example.com/codeflow?start=22&length=10)

Could we use `ModelConfig` or something similar?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -266,6 +266,7 @@ class ModelDownloadType(Enum):
     FTP = 'ftp'
     GIT = 'git'
     HTTP = 'http'
+    AZURE = 'azure'
```
skasturi, line 269: Could we please add notes on implications of using these? 

For example,
`ftp` - to be implemented
`git` - supports only lfs
`http` - to be implemented
`azure` - Anything supported by azcopy

?

skasturi: Also, we could raise errors accordingly

skasturi:
>     """Update the yaml file after getting the model has been prepared."""

(nit) please update docs for the params

---
Refers to: scripts/azureml-assets/azureml/assets/release/asset_publish.py:61 in 5198b05. [](commit_id = 5198b05ad3992cfbcd0ea67d4c4ac1c1929c3065, deletion_comment = False)

skasturi:
>         model_file = yaml.safe_load(f)

Error handling? 

---
Refers to: scripts/azureml-assets/azureml/assets/release/asset_publish.py:63 in 5198b05. [](commit_id = 5198b05ad3992cfbcd0ea67d4c4ac1c1929c3065, deletion_comment = False)

skasturi:
> def test_files_location(dir: Path):

(nit) Prefix `test_` is usually used to represent a test method.

Could we please rename to `find_testfiles`? 

---
Refers to: scripts/azureml-assets/azureml/assets/release/asset_publish.py:27 in 5198b05. [](commit_id = 5198b05ad3992cfbcd0ea67d4c4ac1c1929c3065, deletion_comment = False)

skasturi:
> def test_files_preprocess(test_jobs, asset_ids: dict):

Rename to `preprocess_testfiles`? 

---
Refers to: scripts/azureml-assets/azureml/assets/release/asset_publish.py:42 in 5198b05. [](commit_id = 5198b05ad3992cfbcd0ea67d4c4ac1c1929c3065, deletion_comment = False)

skasturi (changes requested):
:clock1:

.github/workflows/assets-test.yaml
```diff
@@ -131,6 +136,10 @@ jobs:
       - name: Install dependencies
         run: pip install -e $scripts_azureml_assets_dir
 
+      - name: Setup azcopy
+        shell: bash
+        run: bash ${{ scripts_dir.scripts_dir }}/install_azcopy.sh
```
elesel, line 141:
```suggestion
        run: bash $scripts_dir/install_azcopy.sh
```

.github/workflows/assets-test.yaml
```diff
@@ -79,6 +80,10 @@ jobs:
...
+
```
elesel, line 86: Please remove. azcopy won't be used during this job. It's just for checking the execution context.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
elesel: Only if it's passed as an arg to functions in model_utils please, since only want YAML handling code in ModelConfig.

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
```diff
@@ -67,15 +67,27 @@ def _download_git_model(self) -> bool:
             if commit_exists != 0:
                 # TODO: Error handling incase of incorrect commit hash.
                 logger.log_warning("Commit hash doesn't exist. Using model from latest HEAD.")
-        git_path = os.path.join(self.model_dir, '.git')
+        git_path = os.path.join(self.model_dir, ".git")
```
elesel, line 70: We're trying to use the `Path` module for directory handling where possible. Consider `self.model_dir / ".git"` instead of `os.path.join...`

scripts/azureml-assets/azureml/assets/config.py
novaturient95: added

.github/workflows/assets-test.yaml
novaturient95: We don't test assets publish to some temp ws during PR creation?

scripts/azureml-assets/azureml/assets/release/asset_publish.py
novaturient95: we have modelConfig. Made ModelUtils class static

scripts/install_azcopy.sh
novaturient95: sure

.github/workflows/assets-test.yaml
elesel: Awesome. It is needed for the test job below, so let's leave it there.

scripts/install_azcopy.sh
novaturient95: Done

scripts/azureml-assets/azureml/assets/release/model_publish_utils/utils.py
novaturient95: addressed

novaturient95:
Closing in favor of: https://github.com/Azure/azureml-assets/pull/280

PR #277, opened 2022-11-18
make assets publishing ordered

quchuyuan:
since some asset types may depend on others, making different types of assets publish in order.

### DIFF
--- a/scripts/azureml-assets/azureml/assets/release/asset_publish.py
+++ b/scripts/azureml-assets/azureml/assets/release/asset_publish.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from string import Template
 from subprocess import PIPE, STDOUT, run
 from tempfile import TemporaryDirectory
+from collections import defaultdict
 from typing import List
 import azureml.assets as assets
 import azureml.assets.util as util
@@ -235,7 +236,7 @@ if __name__ == "__main__":
     all_assets = util.find_assets(input_dirs=assets_dir)
     assets_by_type = defaultdict(list)
     for asset in all_assets:
-        assets_by_type[asset.type].append(asset)
+        assets_by_type[asset.type.value].append(asset)
 
     for publish_asset_type in PUBLISH_ORDER:
         logger.print(f"now publishing {publish_asset_type.value}s.")
--- a/scripts/azureml-assets/azureml/assets/release/asset_publish.py
+++ b/scripts/azureml-assets/azureml/assets/release/asset_publish.py
@@ -23,6 +23,7 @@ from azureml.assets.util import logger
 ASSET_ID_TEMPLATE = Template(
     "azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
 TEST_YML = "tests.yml"
+PUBLISH_ORDER = [assets.AssetType.ENVIRONMENT, assets.AssetType.COMPONENT, assets.AssetType.MODEL]
 
 
 def test_files_location(dir: Path):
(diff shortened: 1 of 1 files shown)


scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -231,49 +232,52 @@ def _str2bool(v: str) -> bool:
...
-        final_version = asset.version + "-" + \
-            passed_version if passed_version else asset.version
-        logger.print(f"final version: {final_version}")
-        asset_ids[asset.name] = ASSET_ID_TEMPLATE.substitute(
-            registry_name=registry_name,
-            asset_type=f"{asset.type.value}s",
-            asset_name=asset.name,
-            version=final_version,
-        )
+        for asset in util.find_assets(input_dirs=assets_dir, types=[publish_asset_type]):
```
elesel, line 238: This is one of the more expensive operations. You probably want to do this first for all, without specifying `types=...` and then have inner loops for PUBLISH_ORDER, then publish_list.

quchuyuan: Made changes to this.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -231,49 +232,58 @@ def _str2bool(v: str) -> bool:
...
     failure_list = []
-    for asset in util.find_assets(input_dirs=assets_dir):
-        asset_names = publish_list.get(asset.type.value, [])
-        if not ("*" in asset_names or asset.name in asset_names):
-            logger.print(
-                f"Skipping registering asset {asset.name} because it is not in the publish list")
+    all_assets = util.find_assets(input_dirs=assets_dir)
+    assets_by_type = {}
+    for asset in all_assets:
+        assets_by_type[asset.type.value] = assets_by_type.get(asset.type.value, []).append(asset)
```
elesel, line 238: defaultdict is perfect here. You'll need to add `from collections import defaultdict` up above.
```suggestion
    assets_by_type = defaultdict(list)
    for asset in all_assets:
        assets_by_type[asset.type].append(asset)
```

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -231,49 +232,58 @@ def _str2bool(v: str) -> bool:
...
+
+    for publish_asset_type in PUBLISH_ORDER:
+        logger.print(f"now publishing {publish_asset_type.value}s.")
+        if publish_asset_type.value not in publish_list:
             continue
```
elesel, line 243: Consider removing these since the `for` below will do the same thing.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -231,49 +232,58 @@ def _str2bool(v: str) -> bool:
...
-        final_version = asset.version + "-" + \
-            passed_version if passed_version else asset.version
-        logger.print(f"final version: {final_version}")
-        asset_ids[asset.name] = ASSET_ID_TEMPLATE.substitute(
-            registry_name=registry_name,
-            asset_type=f"{asset.type.value}s",
-            asset_name=asset.name,
-            version=final_version,
-        )
+        for asset in assets_by_type.get(publish_asset_type.value, []):
```
elesel, line 244:
```suggestion
        for asset in assets_by_type.get(publish_asset_type, []):
```

wayliums (dismissed):

scripts/azureml-assets/azureml/assets/release/asset_publish.py
quchuyuan: What if a type in `PUBLISH_ORDER` is not in `publish_list`?
Will the for-loop throw errors if iterate an empty list?

elesel: Then it will hit this part:
```
logger.print(
    f"Skipping registering asset {asset.name} because it is not in the publish list")
    continue
```

elesel: Up to you... You can leave it in as a shortcut.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
quchuyuan: I think we should have `.value` since we are using lower cases in keys

elesel: If you use enum values as keys in `assets_by_type` you just need to be consistent in setting and reading them. This part below uses the string value against the YAML that was read in: `if not ("*" in asset_names or asset.name in asset_names):`

quchuyuan: I think we are always using `.value` when interacting with the `publish_list` since it is using lower cases.

quchuyuan: changed `assets_by_type`'s keys to string

elesel (approved):

wayliums (approved):

PR #289, opened 2022-12-01
Removed azcopy login and some minor fixes

novaturient95:


scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -108,11 +108,11 @@ def model_prepare(model_config: assets.ModelConfig, spec_file_path: Path, model_
     elif model_config.type == assets.ModelType.MLFLOW:
         can_publish_model = ModelDownloadUtils.download_model(model_config.path.type, model_config.path.uri, model_dir)
         if can_publish_model:
-            model.path = Path(model_dir) / model.name
+            model.path = model_dir
             if not model_config.flavors:
                 # try fetching flavors from MLModel file
                 mlmodel_file_path = (
-                    model.path / MLFlowModelUtils.MLFLOW_MODEL_PATH / MLFlowModelUtils.MLMODEL_FILE_NAME
+                    Path(model_dir) / MLFlowModelUtils.MLFLOW_MODEL_PATH / MLFlowModelUtils.MLMODEL_FILE_NAME
```
elesel, line 115: You shouldn't need to convert `model_dir` to a `Path` object if it's already being passed into the `model_prepare` function as one, as we should be able to assume from the function signature. This should instead be fixed upstream in any scripts that call the function.

novaturient95: Thanks Louie. Done

elesel (dismissed):

elesel (approved):

PR #294, opened 2022-12-07
generate failed assets report

quchuyuan:
generate failed assets report to make it clear to us which assets are not published successfully.


scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -297,14 +297,27 @@ def _str2bool(v: str) -> bool:
...
+
+        logger.print(f"failed assets: {failed_assets}")
+        for asset_type, asset_names in failed_assets.items():
+            logger.log_warning(f"Failed to register {asset_type}s: {asset_names}")
+        # the following dump process will generate a yaml file for the report
+        # process in the end of the publishing script
+        result_path = Path().resolve() / "failed_assets.yml"
+        dump_data = {asset_type: asset_names for asset_type, asset_names in failed_assets.items()}
+        with open(result_path, "w") as file:
+            yaml.dump(dump_data, file, default_flow_style=False, sort_keys=False)
```
elesel, line 320: I think we need to cut down on verbosity and this should be simpler:
```suggestion
        for asset in failure_list:
            failed_assets[asset.type.value].append(asset.name)

        for asset_type, asset_names in failed_assets.items():
            logger.log_warning(f"Failed to register {asset_type}s: {asset_names}")
        # the following dump process will generate a yaml file for the report
        # process in the end of the publishing script
        with open("failed_assets.yml", "w") as file:
            yaml.dump(failed_assets, file, default_flow_style=False, sort_keys=False)
```

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -297,14 +297,27 @@ def _str2bool(v: str) -> bool:
...
+        for asset in failure_list:
+            logger.print(f"processing failed asset: {asset.name}")
+            failed_assets[asset.type.value].append(asset.name)
+
+        logger.print(f"failed assets: {failed_assets}")
+        for asset_type, asset_names in failed_assets.items():
+            logger.log_warning(f"Failed to register {asset_type}s: {asset_names}")
+        # the following dump process will generate a yaml file for the report
+        # process in the end of the publishing script
+        result_path = Path().resolve() / "failed_assets.yml"
```
elesel, line 317: Ideally this filename should come from an arg passed to the script.

scripts/azureml-assets/azureml/assets/release/asset_publish.py
```diff
@@ -176,8 +176,8 @@ def run_command(
 
     # Check command result
     if results.returncode != 0:
-        logger.log_warning(f"Error creating {asset}")
-        failure_list.append(asset.name)
+        logger.log_warning(f"Error creating {asset.type.value} : {asset.name}")
```
novaturient95, line 179: adding on to @elesel comment, let us print failure logs at one place

elesel (dismissed):

scripts/azureml-assets/azureml/assets/release/asset_publish.py
quchuyuan: The reason why I have `dump_data` is that `yaml.dump` a `defaultdict(list)` will give us a yaml file beginning with 
!!python/object/apply:collections.defaultdict
args:
- !!python/name:builtins.list ''
dictitems:

And `dump_data` could eliminate these unwanted lines. What do you think?

scripts/azureml-assets/azureml/assets/release/asset_publish.py
quchuyuan: done

elesel (approved):

scripts/azureml-assets/azureml/assets/release/asset_publish.py
elesel: You could also just do `yaml.dump(dict(failed_assets), file, default_flow_style=False, sort_keys=False)` ([reference](https://stackoverflow.com/questions/20428636/how-to-convert-defaultdict-to-dict))

wayliums (approved):

PR #317, opened 2023-01-06
Remove legacy prefixes, clean up asset names

elesel:
- Remove microsoft_azureml_ prefix
- Change underscores to dashes in asset names
- Clean up directory names
- Remove "component" asset name suffixes


training/automl/components/text_ner/spec.yaml
```diff
@@ -1,5 +1,5 @@
 $schema: http://azureml/sdk-2-0/AutoMLComponent.json
-name: microsoft_azureml_automl_text_ner_component
+name: automl-text-ner
```
novaturient95, line 2: @shouryaj can you please help to confirm naming of AutoML components. "-" will not work for registering component using CLI/SDK. But in general if we should get rid of **microsoft_azureml** prefix

elesel: There's an ongoing discussion regarding dashes in component names in the AzureML Python SDK channel.

skasturi:
Do we happen to have workflows that are executing these as BVTs?

skasturi:
If yes, do we need to change those as well as i see assets & spec being changed? 

May be related to the infra work that @quchuyuan is doing? 

---
In reply to: [1378301920](https://github.com/Azure/azureml-assets/pull/317#issuecomment-1378301920) [](http://example.com/codeflow?ancestors=1378301920)

PR #321, opened 2023-01-06
Make AssetConfig objects sortable

elesel:


test/resources/config/env1-2/environment.yaml
```diff
@@ -0,0 +1,11 @@
+image:
+  name: azureml/curated/tensorflow-2.8-ubuntu20.04-py38-cuda11-gpu
```
vizhur, line 2: should we have dummy image names for both base and result image? to not overwrite by accident and to not keep static references to images that eventually will be deprecated?

elesel: Sure!

vizhur (approved):

PR #322, opened 2023-01-06
update mltable dependency to 1.0.0 in RAI images

imatiach-msft:
update mltable dependency to 1.0.0 in RAI images

One of our customers ran into an issue where the decimal columns were treated as string in a loaded mltable in the dpv2 components.  This bug is fixed in mltable 1.0.0 dependency.  This PR fixes the customer issue by upgrading our images to use the latest mltable==1.0.0.

Also updated the packages: 'azure-storage-blob<=12.13.0' 'numpy<1.24.0'
to fix image build failures.  The numpy pin is due to newest numpy release not being compatible with latest numba, and azure-storage-blob pin is to remain compatible with both latest azure-ai-ml and latest azureml-mlflow packages and their dependencies.

### DIFF
--- a/responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/context/Dockerfile
+++ b/responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/context/Dockerfile
@@ -12,7 +12,7 @@ RUN conda env create -p $AZUREML_CONDA_ENVIRONMENT_PATH -f conda_dependencies.ya
     conda run -p $AZUREML_CONDA_ENVIRONMENT_PATH pip cache purge && \
     conda clean -a -y
 
-RUN pip install --pre 'azure-ai-ml'
+RUN pip install --pre 'azure-ai-ml' 'azure-storage-blob<=12.13.0' 'numpy<1.24.0'
 
 # Workaround for dependency conflict around tqdm with nlp-feature-extractors
 RUN pip install 'datasets==1.18.4'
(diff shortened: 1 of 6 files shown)


responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/context/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN conda env create -p $AZUREML_CONDA_ENVIRONMENT_PATH -f conda_dependencies.ya
     conda run -p $AZUREML_CONDA_ENVIRONMENT_PATH pip cache purge && \
     conda clean -a -y
 
-RUN pip install --pre 'azure-ai-ml'
+RUN pip install --pre 'azure-ai-ml' 'azure-storage-blob<=12.13.0' 'numpy<1.24.0'
```
gaugup, line 15: Can we add the numpy and azure-storage-blob to conda dependencies? Seems like lots of dependencies in one line. 

Also does the pre flag apply to the last two dependencies?

gaugup (commented):
Can we add the numpy and azure-storage-blob to conda dependencies? Seems like lots of dependencies in one line.

Also does the pre flag apply to the last two dependencies?

imatiach-msft:
"Can we add the numpy and azure-storage-blob to conda dependencies? Seems like lots of dependencies in one line."
they need to be on the same line due to the azure-ai-ml dependency upgrading to latest.  The azure-ai-ml can't be with other packages due to some dependency conflict I believe.

tongyu-microsoft (approved):

PR #362, opened 2023-01-30
Unpin pip to get latest base python packages for fixing vulnerability.

kicha0:


responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/context/conda_dependencies.yaml
```diff
@@ -6,7 +6,6 @@ channels:
   - pytorch
 dependencies:
   - python=3.8
-  - pip=22.1.2
```
imatiach-msft, line 9: nit: maybe still pin but just to the latest version?  or do you think it's better to not pin at all?

imatiach-msft (dismissed):

responsibleai/environments/responsibleai-text-ubuntu20.04-py38-cpu/context/conda_dependencies.yaml
kicha0: I suspect by pinning pip version we will end up having to update it every once in a while for python package vulnerabilities.

tongyu-microsoft (approved):

PR #391, opened 2023-02-13
Ayushmishra/allow short env ref in component

novaturient95:


scripts/release/asset_publish.py
```diff
@@ -277,33 +279,60 @@ def _str2bool(v: str) -> bool:
...
-                    registry_name, final_version, resource_group, workspace, debug_mode)
-                # Run command
-                run_command(cmd, failure_list, debug_mode)
+            if asset.type == assets.AssetType.COMPONENT:
+                # load component and check if environment exists
+                logger.print(f"spec's path: {asset.spec_with_path}")
+                component = load_component(asset.spec_with_path)
+                env = component.environment
+                ws_env_pattern = re.compile("^(.+):(.+)$")
+                registry_env_pattern = re.compile("^azureml://registries/.+/environments/(.+)/versions/(.+)")
```
skasturi, line 288: `ws_env_pattern` is a superset of `registry_env_pattern`. 

For example: `azureml://registries/r1/environments/e1/versions/1` matches both the regexes.

skasturi: I assume you are trying to distinguish b/w workspace vs registry based assets?

scripts/release/asset_publish.py
```diff
@@ -277,33 +279,60 @@ def _str2bool(v: str) -> bool:
...
+                        continue
+
+                # both ws and registry env follows the same grouping
+                env_name = match.group(1)
+                env_version = match.group(2)
+                logger.print(f"Env name:version => {env_name}:{env_version}")
+                logger.print(f"Using final_version {final_version} to check if env registered")
+                try:
+                    mlclient = MLClient(DefaultAzureCredential(), subscription_id=subscription_id, registry_name=registry_name) 
+                    env = mlclient.environments.get(name=env_name, version=final_version)
```
skasturi, line 304: Is it expected for the publishing script to have access to workspace assets?

skasturi:
> def model_prepare(model_config: assets.ModelConfig, spec_file_path: Path, model_dir: Path) -> bool:

(nit) Rename to "prepare_model" since we are following "verb_noun" convention across the file. 

---
Refers to: scripts/release/asset_publish.py:79 in e0405db. [](commit_id = e0405dbf22f6502eeb3c47ae0e49aa519dc196af, deletion_comment = False)

scripts/release/asset_publish.py
```diff
@@ -63,14 +65,14 @@ def preprocess_test_files(test_jobs, asset_ids: dict):
                           sort_keys=False)
 
 
-def dump_model_spec(model, spec_file) -> bool:
-    """Update the yaml file after getting the model has been prepared."""
+def update_spec(asset, spec_file) -> bool:
```
skasturi, line 68: (nit) Docs & type hints.

skasturi (dismissed):
:clock1:

scripts/release/asset_publish.py
novaturient95: We are accessing registry env asset here. `mlclient` is initalised with `registry_name`, which would make request to registry instead of ws

scripts/release/asset_publish.py
novaturient95: Thank you. I have reversed the flow now. Checking registry env pattern first followed by ws env. Its just to fetch env name from string and then checking if env is already registered in the registry.

scripts/release/asset_publish.py
novaturient95: Done

vizhur (dismissed):

scripts/release/asset_publish.py
```diff
@@ -63,22 +65,30 @@ def preprocess_test_files(test_jobs, asset_ids: dict):
...
+    """
     try:
-        model_dict = json.loads(json.dumps(model._to_dict()))
-        util.dump_yaml(model_dict, spec_file)
+        asset_dict = json.loads(json.dumps(asset._to_dict()))
+        util.dump_yaml(asset_dict, spec_file)
         return True
     except Exception as e:
-        logger.log_error(f"Failed to update model spec => {str(e)}")
+        logger.log_error(f"Failed to update spec => {str(e)}")
```
elesel, line 83: For f-strings, `str()` is implicit.
```suggestion
        logger.log_error(f"Failed to update spec: {e}")
```

scripts/release/asset_publish.py
```diff
@@ -63,22 +65,30 @@ def preprocess_test_files(test_jobs, asset_ids: dict):
                           sort_keys=False)
 
 
-def dump_model_spec(model, spec_file) -> bool:
-    """Update the yaml file after getting the model has been prepared."""
+def update_spec(asset: Any, spec_file: Path) -> bool:
```
elesel, line 68: Would `dict` be more appropriate than `Any`?

scripts/release/asset_publish.py
```diff
@@ -277,33 +286,66 @@ def _str2bool(v: str) -> bool:
...
-                    asset.type.value, str(asset.spec_with_path),
-                    registry_name, final_version, resource_group, workspace, debug_mode)
-                # Run command
-                run_command(cmd, failure_list, debug_mode)
+            if asset.type == assets.AssetType.COMPONENT:
+                # load component and check if environment exists
+                logger.print(f"spec's path: {asset.spec_with_path}")
+                component = load_component(asset.spec_with_path)
+                env = component.environment
+                ws_env_pattern = re.compile("^(.+):(.+)$")
```
elesel, line 294: Shouldn't this be:
```suggestion
                ws_env_pattern = re.compile("^azureml:(.+):(.+)$")
```

scripts/release/asset_publish.py
```diff
@@ -277,33 +286,66 @@ def _str2bool(v: str) -> bool:
...
+                registry_env_pattern = re.compile("^azureml://registries/.+/environments/(.+)/versions/(.+)")
+                match = registry_env_pattern.search(env)
+                if not match:
+                    match = ws_env_pattern.search(env)
```
elesel, line 298: Nit: You probably want `match()` instead.
```suggestion
                match = registry_env_pattern.match(env)
                if not match:
                    match = ws_env_pattern.match(env)
```

scripts/release/asset_publish.py
```diff
@@ -277,33 +286,66 @@ def _str2bool(v: str) -> bool:
...
+                logger.print(f"Using final_version {final_version} to check if env registered")
+                try:
+                    mlclient = MLClient(
+                        DefaultAzureCredential(),
+                        subscription_id=subscription_id,
+                        registry_name=registry_name,
+                    )
+                    env = mlclient.environments.get(name=env_name, version=final_version)
+                except Exception as e:
+                    logger.print(f"Fetching component env from registry failed => str{e}")
```
elesel, line 320: `str` would be a literal in the f-string.
```suggestion
                    logger.print(f"Fetching component env from registry failed: {e}")
```

scripts/release/asset_publish.py
```diff
@@ -277,33 +286,66 @@ def _str2bool(v: str) -> bool:
...
-                    registry_name, final_version, resource_group, workspace, debug_mode)
-                # Run command
-                run_command(cmd, failure_list, debug_mode)
+            if asset.type == assets.AssetType.COMPONENT:
+                # load component and check if environment exists
+                logger.print(f"spec's path: {asset.spec_with_path}")
+                component = load_component(asset.spec_with_path)
+                env = component.environment
+                ws_env_pattern = re.compile("^(.+):(.+)$")
+                registry_env_pattern = re.compile("^azureml://registries/.+/environments/(.+)/versions/(.+)")
```
elesel, line 295: Consider making these globals at the top of the script.

scripts/release/asset_publish.py
```diff
@@ -277,33 +286,66 @@ def _str2bool(v: str) -> bool:
...
+                match = registry_env_pattern.search(env)
+                if not match:
+                    match = ws_env_pattern.search(env)
+                    if not match:
+                        logger.print(
+                            "Env ID doesnot match with expected workspace or registry pattern"
```
elesel, line 301:
```suggestion
                            "Env ID doesn't match workspace or registry pattern"
```

scripts/release/asset_publish.py
```diff
@@ -277,33 +286,66 @@ def _str2bool(v: str) -> bool:
...
-                            cmd = assemble_command(
-                                asset.type.value, str(asset.spec_with_path),
-                                registry_name, asset.version, resource_group, workspace, debug_mode)
-                            # Run command
-                            run_command(cmd, failure_list, debug_mode)
+                        if not prepare_model(model_config, asset.spec_with_path, Path(tempdir)):
+                            raise Exception(f"Could not prepare model at {asset.spec_with_path}")
                 except Exception as e:
-                    logger.log_error(f"Exception in loading model config: {str(e)}")
+                    logger.log_error(f"Model prepare exception. Error => {str(e)}")
```
elesel, line 336:
```suggestion
                    logger.log_error(f"Model prepare exception. Error => {e}")
```

scripts/release/asset_publish.py
elesel, line 294: Consider adding support for `azureml:<env_name>@latest`?

scripts/release/asset_publish.py
novaturient95: Done

scripts/release/asset_publish.py
novaturient95: Thanks for suggestion. Done

scripts/release/asset_publish.py
novaturient95: This would be type: azure.ai.ml.entities._asset.Asset.
Updated to below: 

`def update_spec(asset: Union[Component, Environment, Model], spec_file: Path) -> bool:`

scripts/release/asset_publish.py
novaturient95: Done, thank you.

scripts/release/asset_publish.py
novaturient95: Done

scripts/release/asset_publish.py
novaturient95: To fetch env or model in registry using latest tag is still not supported. I get following error when tried: 

image

novaturient95: `azureml:` prefix is omitted when I call load_component. So for a component with `environment: azureml:AzureML-sklearn-1.0-ubuntu20.04-py38-cpu@latest` it captures env as below: 

image

elesel: This could be in a later PR, though. I'm just saying that this feature could be fleshed out even more by providing support for `@latest`.

elesel (approved):

scripts/release/asset_publish.py
novaturient95: Yeah make sense. I can add support label in next PR

scripts/release/asset_publish.py
```diff
@@ -277,33 +289,65 @@ def _str2bool(v: str) -> bool:
...
+                env_version = match.group(2)
+                logger.print(f"Env name:version => {env_name}:{env_version}")
+                logger.print(f"Using final_version {final_version} to check if env registered")
+                try:
+                    mlclient = MLClient(
+                        DefaultAzureCredential(),
+                        subscription_id=subscription_id,
+                        registry_name=registry_name,
+                    )
+                    env = mlclient.environments.get(name=env_name, version=final_version)
```
elesel, line 320: This approach will work for environments in azureml-preview and azureml (prod), where there aren't any timestamp/release-based suffixes added to versions. Note that for now environments in azureml need to be released via a separate pipeline first, then the component that references it can be released.

novaturient95: thank you Louie. I have updated code path to check for env version mentioned in the spec file and final version calculated as part of asset publish process. It should cover both the use cases. Can you please check once more and help to review?Thanks

scripts/release/asset_publish.py
```diff
@@ -311,17 +311,29 @@ def _str2bool(v: str) -> bool:
                 env_version = match.group(2)
                 logger.print(f"Env name:version => {env_name}:{env_version}")
                 logger.print(f"Using final_version {final_version} to check if env registered")
+
+                mlclient = MLClient(
+                    DefaultAzureCredential(),
+                    subscription_id=subscription_id,
+                    registry_name=registry_name,
+                )
```
elesel, line 319: Consider moving this outside of the loop.

scripts/release/asset_publish.py
```diff
@@ -311,17 +311,29 @@ def _str2bool(v: str) -> bool:
...
+
+                if not env:
+                    try:
+                        env = mlclient.environments.get(name=env_name, version=final_version)
+                    except Exception as e:
+                        logger.print(
+                            f"Fetching component env {env_name}:{final_version} from registry failed. Error\n\n{e}"
+                        )
+                        failure_list.append(asset)
+                        continue
```
elesel, line 336: Consider creating a `for` loop over `[env_version, final_version]` to simplify the logic here.

scripts/release/asset_publish.py
novaturient95: Thank you done

scripts/release/asset_publish.py
novaturient95: Done. Thank you

scripts/release/asset_publish.py
skasturi: Is this taken care of or are we targeting in the next PR?

skasturi:
Could you please add detailed description?

scripts/release/asset_publish.py
```diff
@@ -14,20 +14,25 @@
...
+from azure.ai.ml import MLClient, load_component, load_model
+from azure.ai.ml.entities import Component, Environment, Model
+from azure.identity import DefaultAzureCredential
+
 
 ASSET_ID_TEMPLATE = Template(
     "azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
 TEST_YML = "tests.yml"
 PUBLISH_ORDER = [assets.AssetType.ENVIRONMENT, assets.AssetType.COMPONENT, assets.AssetType.MODEL]
+WORKSPACE_ENV_PATTERN = re.compile("^(.+):(.+)$")
```
skasturi, line 34: > re.compile("^(.+):(.+)$" [](http://example.com/codeflow?start=23&length=25)

Isn't this too broad?

scripts/release/asset_publish.py
```diff
@@ -63,22 +68,30 @@ def preprocess_test_files(test_jobs, asset_ids: dict):
                           sort_keys=False)
 
 
-def dump_model_spec(model, spec_file) -> bool:
-    """Update the yaml file after getting the model has been prepared."""
+def update_spec(asset: Union[Component, Environment, Model], spec_file: Path) -> bool:
+    """Update the yaml spec file with updated properties in asset.
```
skasturi, line 72: >properties [](http://example.com/codeflow?start=46&length=10)

Are properties immutable? Or are tags?

scripts/release/asset_publish.py
```diff
@@ -63,22 +68,30 @@ def preprocess_test_files(test_jobs, asset_ids: dict):
...
+
+    :param asset: Asset loaded using load_*(component, environemnt, model) method.
+    :type asset: Union[Component, Environment, Model]
+    :param spec_file: path to model spec file
```
skasturi, line 76: >model [](http://example.com/codeflow?start=30&length=5)

(nit) asset?

scripts/release/asset_publish.py
novaturient95: Updated the regex for both ws and registry to consider labels as well. But to be able to fetch env asset using latest label we will need to wait for support in SDK. Added a bug here for tacking: #415

scripts/release/asset_publish.py
novaturient95: Updated

scripts/release/asset_publish.py
```diff
@@ -14,20 +14,26 @@
...
 
-ASSET_ID_TEMPLATE = Template(
-    "azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
+
+ASSET_ID_TEMPLATE = Template("azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
 TEST_YML = "tests.yml"
 PUBLISH_ORDER = [assets.AssetType.ENVIRONMENT, assets.AssetType.COMPONENT, assets.AssetType.MODEL]
+WORKSPACE_ENV_PATTERN = re.compile(r"^([\w_-]+)(?:\:([0-9.]+)|\@([a-z]+))$")
+REGISTRY_ENV_PATTERN = re.compile(
+    r"^azureml://registries/[\w_-]+/environments/([\w_-]+)/(?:versions/([0-9.]+)|labels/([a-z]+))"
```
elesel, line 35: Simplifying a bit. See https://regex101.com/r/fvDNmy/1
```suggestion
    r"^azureml://registries/.+/environments/(.+)/(?:versions/(.+)|labels/(.+))"
```

scripts/release/asset_publish.py
```diff
@@ -277,33 +293,75 @@ def _str2bool(v: str) -> bool:
...
+                # both ws and registry env follows the same grouping
+                env_name = match.group(1)
+                env_version = match.group(2)
+                env_label = match.group(3)
+                logger.print(f"Env name: {env_name}, version: {env_version}, label: {env_label}")
+
+                if env_label:
+                    # TODO: Add fetching env from label
+                    # https://github.com/Azure/azureml-assets/issues/415
+                    print("Unexpected !!! Registering a component with env label is still to be supported.")
```
elesel, line 320:
```suggestion
                    logger.print("Unexpected !!! Registering a component with env label is not yet supported.")
```

elesel (changes requested):
WORKSPACE_ENV_PATTERN needs the azureml: prefix but the rest of the changes look good.

scripts/release/asset_publish.py
```diff
@@ -14,20 +14,26 @@
...
+from azure.ai.ml.entities import Component, Environment, Model
+from azure.identity import DefaultAzureCredential
 
-ASSET_ID_TEMPLATE = Template(
-    "azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
+
+ASSET_ID_TEMPLATE = Template("azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
 TEST_YML = "tests.yml"
 PUBLISH_ORDER = [assets.AssetType.ENVIRONMENT, assets.AssetType.COMPONENT, assets.AssetType.MODEL]
+WORKSPACE_ENV_PATTERN = re.compile(r"^([\w_-]+)(?:\:([0-9.]+)|\@([a-z]+))$")
```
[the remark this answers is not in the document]
novaturient95: For workspace, when we load_component, it omits `azureml:`.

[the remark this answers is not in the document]
elesel: Ahh, gotcha. However, I think it would be great to include it since it matches the short form for environments.

scripts/release/asset_publish.py
novaturient95: image

@elesel I got this comment at top by @skasturi to not have it this broad. Since we are using internally and mostly to get group values (name, version, label) and not using in validation gates/workflows to make sure naming is as per standard, going broader can be okay. Do we plan to make use of regex to validate spec details in future?

scripts/release/asset_publish.py
```diff
@@ -295,45 +299,43 @@ def _str2bool(v: str) -> bool:
                 component = load_component(asset.spec_with_path)
                 env = component.environment
 
-                match = REGISTRY_ENV_PATTERN.match(env)
+                for pattern in [REGISTRY_ENV_PATTERN, WORKSPACE_ENV_PATTERN]:
+                    if match := pattern.match(env):
```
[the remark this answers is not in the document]
novaturient95: Done

scripts/release/asset_publish.py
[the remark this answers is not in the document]
novaturient95: Something like this: https://regex101.com/r/x0d39L/1 ?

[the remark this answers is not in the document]
elesel: This is a little simpler: https://regex101.com/r/D12buY/1

scripts/release/asset_publish.py
elesel: Both versions and labels can be alphanumeric. I don't see the point of making the regex more strict, especially when we're already checking for a few static elements.

scripts/release/asset_publish.py
```diff
@@ -317,15 +317,15 @@ def _str2bool(v: str) -> bool:
...
                     # TODO: Add fetching env from label
                     # https://github.com/Azure/azureml-assets/issues/415
-                    print("Unexpected !!! Registering a component with env label is still to be supported.")
+                    logger.print("Unexpected !!! Registering a component with env label is not yet supported.")
                     failure_list.append(asset)
                     continue
 
                 env = None
-                for version in set(env_version, final_version):
+                for version in set((env_version, final_version)):
```
elesel, line 325: You probably want these to be ordered:
```suggestion
                for version in [env_version, final_version]:
```

scripts/release/asset_publish.py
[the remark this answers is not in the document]
novaturient95: updated

scripts/release/asset_publish.py
novaturient95: updated

scripts/release/asset_publish.py
novaturient95: Done

elesel (approved):

PR #400, opened 2023-02-17
Fix vulnerability for mldesigner and mldesigner-minimal

zhengfeiwang:
Pin `wheel` to 0.38.1 to resolve vulnerability following this link: <https://github.com/advisories/GHSA-qwmp-2cf2-g9g6>


vizhur (approved):
having conda-forge (or the only channel) should resolve the issue without adding the explicit dependency

vizhur (approved):

PR #409, opened 2023-02-23
Update RAI images to fix dependency vulnerability

tongyu-microsoft:
We have 5 vulnerabilities:
https://github.com/advisories/GHSA-ffxj-547x-5j7c
https://github.com/advisories/GHSA-r9hx-vwmv-q579
https://github.com/advisories/GHSA-6hrg-qmvc-2xh8
https://github.com/advisories/GHSA-hwqr-f3v9-hwxr
https://github.com/advisories/GHSA-j8fq-86c5-5v2r

This PR pins those package versions to resolve dependency vulnerability issue.

### DIFF
--- a/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
+++ b/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
@@ -30,7 +30,12 @@ RUN pip install 'azureml-dataset-runtime==1.48.0' \
 RUN pip install 'ml-wrappers~=0.4.2' \
                 'shap==0.40.0' \
                 'interpret-community==0.28.0' \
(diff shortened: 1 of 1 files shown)


romanlutz (approved):

imatiach-msft:
I think these may conflict with AutoML.  These are all coming from automl vision package.  I would hold off on these changes and 1.) first make sure they won't break AutoML DNN vision package, which I think they might 2.) if possible ask them to make these changes instead of us adding all of these pins

imatiach-msft (changes requested):
I think these may conflict with AutoML. These are all coming from automl vision package. I would hold off on these changes and 1.) first make sure they won't break AutoML DNN vision package, which I think they might 2.) if possible ask them to make these changes instead of us adding all of these pins

tongyu-microsoft:
this is fixed in automl-vision package

PR #410, opened 2023-02-24
Fix build failure for responsibleai-ubuntu20.04-py38-cpu

gaugup:


responsibleai/environments/responsibleai-ubuntu20.04-py38-cpu/context/Dockerfile
```diff
@@ -26,6 +26,12 @@ RUN pip install --no-deps 'charset-normalizer==2.0.12' \
                           'webencodings==0.5.1' \
                           'domonic==0.9.10'
 
+# Install azureml packages
```
tongyu-microsoft, line 29: why moving these out of yaml file fix the build failure? Is this because of dependency conflict that Ilya mentioned previously?

imatiach-msft (approved):

responsibleai/environments/responsibleai-ubuntu20.04-py38-cpu/context/Dockerfile
gaugup: I think breaking up the dependencies in multiple pip commands helps with resolution though not 100% why it does.

PR #425, opened 2023-03-08
Validate asset names, environment build/publishing

elesel:


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -4,13 +4,17 @@
...
+import warnings
 from enum import Enum
 from functools import total_ordering
 from pathlib import Path
 from setuptools._vendor.packaging import version
 from typing import Dict, List, Tuple
 from yaml import safe_load
 
+# Ignore setuptools warning about replacing distutils
+warnings.filterwarnings("ignore", message="Setuptools is replacing distutils.", category=UserWarning)
```
vizhur, line 16: where do we see warning about distutils from :o

vizhur (dismissed):

scripts/azureml-assets/azureml/assets/config.py
elesel: It's showing up because of this import `from setuptools._vendor.packaging import version`, which allows us to do smart version ordering without needing an extra dependency. But if setuptools isn't installed this warning is thrown.

PR #426, opened 2023-03-09
Refactor asset publish

novaturient95:
**Problem**
- Using MLClient is consistently causing failures in fetching token from environment.
- MLClient was used to fetch if env version exists before component publish. 

**Solution**
- Used az ml cli to perform the same operations
- Created a thread internally to discuss mlclient issue
- Tested locally and triggering a test release

### DIFF
--- a/scripts/release/asset_publish.py
+++ b/scripts/release/asset_publish.py
@@ -187,7 +187,8 @@ def validate_update_command_component(
 
     # TODO: Bug in env list, which does not give complete detail
     # https://github.com/Azure/azure-sdk-for-python/issues/29248
-    component.environment = f"azureml://registries/{registry_name}/environments/{env['name']}/versions/{env['version']}"
+    env_id = f"azureml://registries/{registry_name}/environments/{env['name']}/versions/{env['version']}"
+    component.environment = env_id
     if not update_spec(component, spec_path):
         logger.print(f"Component update failed for asset spec path: {asset.spec_path}")
         return False
--- a/scripts/release/asset_publish.py
+++ b/scripts/release/asset_publish.py
@@ -178,14 +178,16 @@ def validate_update_command_component(
 
     env = None
     # Check if component's env is registered
-    registered_envs = get_registered_asset_versions(assets.AssetType.ENVIRONMENT, env_name, registry_name)
+    registered_envs = get_registered_asset_versions(assets.AssetType.ENVIRONMENT.value, env_name, registry_name)
     env = next((x for x in registered_envs if x['version'] in [env_version, final_version]), None)
 
     if not env:
         logger.print(f"Could not find a registered env for {component.name}. Please retry again!!!")
         return False
 
-    component.environment = env["id"]
+    # TODO: Bug in env list, which does not give complete detail
+    # https://github.com/Azure/azure-sdk-for-python/issues/29248
+    component.environment = f"azureml://registries/{registry_name}/environments/{env['name']}/versions/{env['version']}"
     if not update_spec(component, spec_path):
         logger.print(f"Component update failed for asset spec path: {asset.spec_path}")
         return False
@@ -248,18 +250,13 @@ def publish_asset(
     debug_mode: bool = None
 ):
     """Publish asset to registry."""
-    registered_assets = get_registered_asset_versions(asset.type.value, asset.name, registry_name, return_dict=True)
-    if version in registered_assets:
-        print(f"Version already registered. Skipping publish for asset: {asset.name}")
-        return
-
     cmd = asset_publish_command(
         asset.type.value, str(asset.spec_with_path),
         registry_name, version, resource_group, workspace_name, debug_mode
     )
 
     # Run command
-    result = run_command(cmd, failure_list, debug_mode)
+    result = run_command(cmd)
     if debug_mode:
         # Capture and redact output
         redacted_output = re.sub(r"Bearer.*", "", result.stdout)
@@ -399,6 +396,13 @@ if __name__ == "__main__":
                         failure_list.append(asset)
                         continue
             elif asset.type == assets.AssetType.MODEL:
+                final_version = asset.version
+                registered_assets = get_registered_asset_versions(
+                    asset.type.value, asset.name, registry_name, return_dict=True)
+                if final_version in registered_assets:
+                    print(f"Version already registered. Skipping publish for asset: {asset.name}")
+                    continue
+
                 try:
                     model_config = asset.extra_config_as_object()
                     with TemporaryDirectory() as tempdir:
--- a/scripts/release/asset_publish.py
+++ b/scripts/release/asset_publish.py
@@ -28,7 +28,7 @@ from azure.ai.ml.entities import Component, Environment, Model
 
 ASSET_ID_TEMPLATE = Template("azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
 TEST_YML = "tests.yml"
-PUBLISH_ORDER = [assets.AssetType.MODEL, assets.AssetType.ENVIRONMENT, assets.AssetType.COMPONENT]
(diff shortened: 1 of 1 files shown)


scripts/release/asset_publish.py
```diff
@@ -179,26 +178,45 @@ def validate_update_command_component(
 
     env = None
     # Check if component's env is registered
-    for version in [env_version, final_version]:
-        try:
-            if (env := mlclient.environments.get(name=env_name, version=version)) is not None:
-                break
-        except Exception as e:
-            logger.print(f"Fetching component env {env_name}:{version} from registry failed. Error: {e}")
+    registered_envs = get_registered_asset_versions(assets.AssetType.ENVIRONMENT.value, env_name, registry_name)
```
vizhur, line 181: container returns NextVersion, it is probably more robust to use that.

if you retrieve or attempt to register, you can also retrieve by label to resolve to the version to address your label support question

vizhur (dismissed):

scripts/release/asset_publish.py
novaturient95: Thanks @vizhur, as discussed offline I will move to using labels to check latest version, when support for the same is provided in CLI. Going with this approach for the time being

scripts/release/asset_publish.py
```diff
@@ -179,26 +178,47 @@ def validate_update_command_component(
...
+    asset_name: str,
+    registry_name: str,
+) -> List[str]:
+    """Command to list of registered asset versions."""
+    cmd = [
+        "az", "ml", asset_type, "list",
+        "--name", asset_name,
+        "--registry-name", registry_name,
+    ]
+    print(cmd)
```
elesel, line 216: Please remove after testing.

scripts/release/asset_publish.py
```diff
@@ -179,26 +178,47 @@ def validate_update_command_component(
...
+    return True
+
+
+def run_command(cmd: List[str]):
+    """Run the command for and return result."""
+    result = run(cmd, stdout=PIPE, stderr=PIPE, encoding=sys.stdout.encoding, errors="ignore")
+    return result
+
+
+def asset_list_command(
```
elesel, line 205: This is only called once. Consider just creating the `cmd` list in `get_registered_asset_versions`.

scripts/release/asset_publish.py
```diff
@@ -221,34 +240,59 @@ def assemble_command(
...
+    return_dict=False
+) -> Union[Dict, List]:
+    """Return list/dict of registered asset versions."""
+    result = run_command(asset_list_command(
+        asset_type=asset_type,
+        asset_name=asset_name,
+        registry_name=registry_name,
+    ))
+    if result.returncode != 0:
+        print(f"Error in listing asset version. stdout:\n{result.stdout}")
```
elesel, line 288: Consider just using `check=True` in the call to `run` in `run_command`.

scripts/release/asset_publish.py
```diff
@@ -221,34 +240,59 @@ def assemble_command(
...
+    if result.returncode != 0:
+        print(f"Error creating {asset.type.value} : {asset.name}")
         failure_list.append(asset)
 
 
+def get_registered_asset_versions(
+    asset_type: str,
+    asset_name: str,
+    registry_name: str,
+    return_dict=False
```
elesel, line 279: It would be cleaner to always return a list since the CLI only returns versions right now. If you want to pick up asset IDs later, you could switch to a dict or class return type.

scripts/release/asset_publish.py
```diff
@@ -179,26 +178,47 @@ def validate_update_command_component(
...
+    component.environment = env_id
     if not update_spec(component, spec_path):
         logger.print(f"Component update failed for asset spec path: {asset.spec_path}")
         return False
+    return True
+
+
+def run_command(cmd: List[str]):
+    """Run the command for and return result."""
+    result = run(cmd, stdout=PIPE, stderr=PIPE, encoding=sys.stdout.encoding, errors="ignore")
```
elesel, line 201: Simpler:
```suggestion
    result = run(cmd, capture_output=True, encoding=sys.stdout.encoding, errors="ignore")
```

scripts/release/asset_publish.py
```diff
@@ -11,26 +11,25 @@
 import sys
 from pathlib import Path
 from string import Template
-from subprocess import PIPE, STDOUT, run
+from subprocess import PIPE, run
```
elesel, line 14: If you accept my suggestion about `capture_output=True` below:
```suggestion
from subprocess import run
```

scripts/release/asset_publish.py
```diff
@@ -221,34 +240,59 @@ def assemble_command(
...
-            errors="ignore",
-        )
-        redacted_output = re.sub(r"Bearer.*", "", results.stdout)
-        logger.print(redacted_output)
-    else:
-        results = run(cmd)
+        redacted_output = re.sub(r"Bearer.*", "", result.stdout)
+        redacted_err = re.sub(r"Bearer.*", "", result.stderr)
+        print(f"stdout: {redacted_output}")
+        print(f"stderr: {redacted_err}")
```
elesel, line 268: Consider creating a regex at the top of the script and also creating a function that sanitizes output.

scripts/release/asset_publish.py
```diff
@@ -221,34 +240,59 @@ def assemble_command(
...
-            stdout=PIPE,
-            stderr=STDOUT,
-            encoding=sys.stdout.encoding,
-            errors="ignore",
-        )
-        redacted_output = re.sub(r"Bearer.*", "", results.stdout)
-        logger.print(redacted_output)
-    else:
-        results = run(cmd)
+        redacted_output = re.sub(r"Bearer.*", "", result.stdout)
```
elesel, line 265: Print the command above this line, to make the logs more readable.

scripts/release/asset_publish.py
elesel, line 268: A little more readable:
```suggestion
        if redacted_output:
            print(f"STDOUT:\n{redacted_output}")
        if redacted_err:
            print(f"STDERR:\n{redacted_err}")
```

scripts/release/asset_publish.py
```diff
@@ -221,34 +240,59 @@ def assemble_command(
...
+        redacted_err = re.sub(r"Bearer.*", "", result.stderr)
+        print(f"stdout: {redacted_output}")
+        print(f"stderr: {redacted_err}")
 
-    # Check command result
-    if results.returncode != 0:
-        logger.log_warning(f"Error creating {asset.type.value} : {asset.name}")
+    if result.returncode != 0:
+        print(f"Error creating {asset.type.value} : {asset.name}")
```
elesel, line 271: Would it help to include any of the command's output in case of error?

elesel (changes requested):

scripts/release/asset_publish.py
```diff
@@ -221,34 +240,59 @@ def assemble_command(
     if debug_mode:
         cmd.append("--debug")
     print(cmd)
```
elesel, line 242: Remove after testing.

scripts/release/asset_publish.py
elesel, line 271: Why not use `logger.log_warning()`?

scripts/release/asset_publish.py
novaturient95: Done

scripts/release/asset_publish.py
novaturient95: Doing only in case of debug. Logging stderr in the logs now

scripts/release/asset_publish.py
novaturient95: I am using it at couple of places. For model where I get a comprehensive output. And Env where I get {name and version}

vizhur (approved):

scripts/release/asset_publish.py
novaturient95: Updated

elesel (approved):

PR #437, opened 2023-03-14
Component asset - Model Deployment

HrishikeshGeedMS:
Component asset created to support model deployment.


elesel (approved):

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,161 @@
...
+        subscription_id=ws._subscription_id,
+        resource_group_name=ws._resource_group,
+        workspace_name=ws._workspace_name,
+    )
+    return ml_client
+
+
+def main(args):
+    """Run main function."""
+    ml_client = get_ml_client()
```
novaturient95, line 115: batch deployment in subsequent PR?

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,161 @@
...
+        help="Maximum queue wait time of a request in ms",
+    )
+
+    # parse args
+    args = parser.parse_args()
+    print("args received ", args)
+    # return args
+
+    # Validating passed input values
+    if args.max_concurrent_requests_per_instance < 1:
```
novaturient95, line 80: Validation for max limits?

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,161 @@
...
+    parser.add_argument(
+        "--endpoint_name",
+        type=str,
+        help="Name of the endpoint",
+    )
+    parser.add_argument(
+        "--deployment_name", type=str, help="Name of the the deployment"
+    )
+    parser.add_argument(
+        "--instance_type",
```
novaturient95, line 38: instance_type validation?

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,161 @@
...
+def main(args):
+    """Run main function."""
+    ml_client = get_ml_client()
+    model_id = (Path(args.registration_details)).read_text()
+    endpoint = get_endpoint(args)
+
+    ml_client.begin_create_or_update(endpoint).wait()
+
+    # deployment
+    deployment = ManagedOnlineDeployment(
```
novaturient95, line 122: There are many deployment properties missing. For instance, request and probe settings in here: https://learn.microsoft.com/en-us/azure/machine-learning/reference-yaml-deployment-managed-online 
How are we planning to add them?

common/components/deploy_model/spec.yaml
```diff
@@ -0,0 +1,77 @@
+$schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
+name: deploy_model
+version: 0.0.1
+type: command
+
+is_deterministic: True
+
+display_name: Deploy model
+description:
+  Deploy a model to a workspace. The component works on compute with [MSI](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-create-manage-compute-instance?tabs=python) attached.
```
novaturient95, line 10: Should we add support for custom models as well? It would require scoring script and env at runtime. How will we capture those details from user?

common/components/deploy_model/spec.yaml
```diff
@@ -0,0 +1,77 @@
...
+  $[[--request_timeout_ms ${{inputs.request_timeout_ms}}]]
+  $[[--max_queue_wait_ms ${{inputs.max_queue_wait_ms}}]]
+  --model_deployment_details ${{outputs.model_deployment_details}}
+
+inputs:
+  # Output of registering component
+  registration_details:
+    type: uri_file
+    optional: false
+    description: Text file that contains the ID of registered model to be deployed
```
novaturient95, line 32: nit: Update to: `File that contains model registration details`?

HrishikeshGeedMS: Any specific reason for this ? Last time louie specifically mentioned to add file type (like text,json) to have clarity

common/components/deploy_model/spec.yaml
HrishikeshGeedMS: Yes ! Initial plan was for ml flow only, so not added support for that.
Should we add it in another pr as next features?

novaturient95: Yes, next PR sounds okay to me if its a lot of effort

common/components/deploy_model/spec.yaml
novaturient95: Yes, specifying a file type is needed.
What I was thinking is we probably need to dump more details like:
```
model_name: <>
model_version: <>
model_type: <>
model_flavors: <>
id: <> [Asset URI]
path: <> [datastore/storage asset URI]
registry or workspace registration? If registry have that as well [This can be figured out from id as well, but having an explicit property will not do any harm]
```

The reason having these details would be useful is, though there is a way to figure out some details using asset ID itself, but for some other we would need to do models.get operation which would require downstream components to do n/w call as well as run on MSI compute. Please let me know what you think and whether having these details make sense in registration component which can be consumed in deploy component

common/src/model_deployment/deploy.py
HrishikeshGeedMS: Yes @novaturient95

common/components/deploy_model/spec.yaml
HrishikeshGeedMS: Make sense @novaturient95 .We will need model_type parameter, when we will add support deployment for custom model  but in current case we do not require such parameter.
Shall we add this parameters in reg component in next release,which will include support for custom model?

novaturient95: yeah, as part of this PR or next PR is okay

common/src/model_deployment/deploy.py
HrishikeshGeedMS: Thanks for the suggestion. Changes it to enum type

common/src/model_deployment/deploy.py
HrishikeshGeedMS: Need to figure out the better way to deal with maximum limit validation. Other than having static mapping,not able to figure out any better idea. As of now do not want to add that just for validation of one parameter.
May be we can discuss on this further and add this in next pr.
For other two parameters, I have added the max limit validation

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+        instance_type=args.instance_type,
+        instance_count=args.instance_count,
+        request_settings=OnlineRequestSettings(
+            max_concurrent_requests_per_instance=args.max_concurrent_requests_per_instance,
+            request_timeout_ms=args.request_timeout_ms,
+            max_queue_wait_ms=args.max_queue_wait_ms,
+        ),
+        liveness_probe=ProbeSettings(
+            failure_threshold=args.failure_threshold_liveness_probe,
+            timeout=args.timeout_liveness_probe,
```
novaturient95, line 203: `success_threshold` missing for liveness_probe

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+        readiness_probe=ProbeSettings(
+            failure_threshold=args.failure_threshold_readiness_probe,
+            success_threshold=args.success_threshold_readiness_probe,
+            timeout=args.timeout_readiness_probe,
+            period=args.period_readiness_probe,
+            initial_delay=args.initial_delay_readiness_probe,
+        ),
+        egress_public_network_access = args.egress_public_network_access,
+    )
+    ml_client.online_deployments.begin_create_or_update(deployment).wait()
```
novaturient95, line 216: Should we catch exception in case endpoint or deployment creation fails?

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+        "--initial_delay_liveness_probe",
+        type=int,
+        default=10,
+        help="The number of seconds after the container has started before the liveness probe is initiated",
+        choices = range(1,500)
+    )
+    parser.add_argument(
+        "--egress_public_network_access",
+        type=int,
+        default="enabled",
```
novaturient95, line 135: type is `int` here. But default provided is a string. Please change type to string

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+        "--timeout_liveness_probe",
+        type=int,
+        default=10,
+        help="The number of seconds after which the liveness probe times out",
+        choices = range(1,500)
+    )
+    parser.add_argument(
+        "--period_liveness_probe",
+        type=int,
+        default=10,
```
novaturient95, line 121: Is probing every 10s recommended? What will be the effect of probing on inference response time?

common/components/deploy_model/spec.yaml
```diff
@@ -0,0 +1,187 @@
...
+  deployment_name:
+    type: string
+    optional: false
+    description: Name of the deployment
+  
+  instance_type:
+    type: string
+    optional: true
+    enum:
+      - Standard_DS1_v2
```
novaturient95, line 58: @skasturi is this okay to have? Suggested @HrishikeshGeedMS to add this, as component user would need to jump and search places and make sure input is added correctly.

novaturient95 (changes requested):

common/src/model_deployment/deploy.py
HrishikeshGeedMS: For them, validation is in parser, if you see there is one parameter choices = range(1,n).. So that will make sure that correct values has been passed or not

common/src/model_deployment/deploy.py
HrishikeshGeedMS: success_threshold for liveness_probe is fixed i.e 1. So I have not added it .
Ref - https://learn.microsoft.com/en-us/azure/machine-learning/reference-yaml-deployment-managed-online#probesettings:~:text=2-,success_threshold,-integer

common/src/model_deployment/deploy.py
HrishikeshGeedMS: Thanks Ayush

common/src/model_deployment/deploy.py
```diff
@@ -0,0 +1,243 @@
...
+    ml_client.begin_create_or_update(endpoint).wait()
+
+    # write deployment details to file
+    endpoint_type = "aml_online_inference"
+    endpoint_response = ml_client.online_endpoints.get(endpoint.name)
+    deployment_details = {
+        "endpoint_name": endpoint.name,
+        "deployment_name": deployment.name,
+        "endpoint_uri": endpoint_response.__dict__["_scoring_uri"],
+        "endpoint_type": endpoint_type,
```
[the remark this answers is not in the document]
HrishikeshGeedMS: Sure

common/components/deploy_model/spec.yaml
```diff
@@ -0,0 +1,187 @@
...
+      - Standard_NC12s_v3
+      - Standard_NC16as_T4_v3
+      - Standard_NC24s_v2
+      - Standard_NC24s_v3
+      - Standard_NC64as_T4_v3
+      - Standard_ND40rs_v2
+      - Standard_ND96asr_v4
+      - Standard_ND96amsr_A100_v4
+    default: Standard_F8s_v2
+    description: Compute instance type to deploy model
```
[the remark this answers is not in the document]
HrishikeshGeedMS: Sure, I will add it in the description. What do you mean by training script here??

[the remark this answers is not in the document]
novaturient95: Deployment python code that gets executed

common/src/model_deployment/deploy.py
HrishikeshGeedMS: Thanks . Added

common/src/model_deployment/deploy.py
HrishikeshGeedMS: Cool. Lets go with the default one as of now, if we will get any better idea we can update it later

common/components/deploy_model/spec.yaml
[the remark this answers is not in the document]
HrishikeshGeedMS: For now I have added in description for instance type, instance count. 
Would it better if some how we can show/or reflect in UI -instance type and it's quota . So that user can choose wisely at that point of time.

novaturient95:
LGTM, please not use model_info for dumping registration details

PR #439, opened 2023-03-15
Add feature attribution drift component

hawestra:
Add feature attribution drift component which computes ndcg metrics using explanations of baseline and production dataset; this is a first draft of the component which includes hard coded values that need to be changed once the model monitoring contract is  updated

sample component: https://ml.azure.com/module/feature_attribution_drift_v1/name/feature_attribution_drift?wsid=/subscriptions/e0fd569c-e34a-4249-8c24-e8d723c7f054/resourceGroups/hawestra-rg/providers/Microsoft.MachineLearningServices/workspaces/hawestra-ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47

### DIFF
--- a/responsibleai/model_monitor/components/src/calculate_attribution.py
+++ b/responsibleai/model_monitor/components/src/calculate_attribution.py
@@ -37,13 +37,14 @@ _get_logger()
 def create_lightgbm_model(X, y, task_type):
     """Create model on which to calculate feature importances
 
-    Args:
-      X: pandas dataframe of x values (data excluding target columns)
-      y: nparray of y values (target column data)
-      task_type: str, the task type (regression or classification) of the resulting model
-
-    Returns:
-       An appropriate model.
+      :param X x values (data excluding target columns)
+      :type X pandas.Dataframe
+      :param y nparray of y values (target column data)
+      :type y nparray
+      :param task_type the task type (regression or classification) of the resulting model
+      :type task_type string
+      :return an appropriate model wrapper
+      :rtype: LightGBMClassifier or LightGBMRegressor
     """
     if (task_type == "classification"):
         lgbm = LGBMClassifier(boosting_type='gbdt', learning_rate=0.1,
@@ -60,16 +61,18 @@ def create_lightgbm_model(X, y, task_type):
 def get_model_wrapper(task_type, target_column, baseline_dataframe, production_dataframe):
     """Create model wrapper using ml-wrappers on which to calculate feature importances
 
-    Args:
-      task_type: str, The task type (regression or classification) of the resulting model
-      target_column: str, the column to predict
-      baseline_dataframe: The baseline dataframe meaning the dataframe used to create the
+      :param task_type The task type (regression or classification) of the resulting model
+      :type task_type string
+      :param target_column the column to predict
+      :type target_column string
+      :param baseline_dataframe The baseline data meaning the data used to create the
       model monitor
-      production_dataframe: The production dataframe meaning the most recent set of data
+      :type baseline_dataframe pandas.DataFrame
+      :param production_dataframe: The production data meaning the most recent set of data
       sent to the model monitor, the current set of data
-
-    Returns:
-       An appropriate model wrapper
+      :type production_dataframe pandas.DataFrame
+      :return an appropriate model wrapper
+      :rtype: PredictionsModelWrapperRegression or PredictionsModelWrapperClassification
     """
     y_train = baseline_dataframe[target_column]
     x_train = baseline_dataframe.drop([target_column], axis=1)
@@ -94,13 +97,13 @@ def get_model_wrapper(task_type, target_column, baseline_dataframe, production_d
 def compute_categorical_features(baseline_dataframe, target_column):
     """Compute which features are categorical based on data type of the columns.
 
-    Args:
-      baseline_dataframe: The baseline dataframe meaning the dataframe used to create the
+      :param baseline_dataframe The baseline data meaning the data used to create the
       model monitor
-      target_column: str, the column to predict
-
-    Returns:
-      list: A list of categorical features.
+      :type baseline_dataframe pandas.DataFrame
+      :param target_column the column to predict
+      :type target_column string
+      :return categorical features
+      :rtype list[string]
     """
     categorical_features = []
     for column in baseline_dataframe.columns:
@@ -114,15 +117,16 @@ def compute_categorical_features(baseline_dataframe, target_column):
 def compute_explanations(model_wrapper, dataframe, categorical_features, target_column, task_type):
     """Compute explanations (feature importances) for a given dataset
 
-    Args:
-      model_wrapper: wrapper around a model that can be used to calculate explanations
-      dataframe: The dataframe
-      categorical_features: list of categorical features not including the target column
-      target_column: str, the column to predict
-      task_type: str, the task type (regression or classification) of the resulting model
-
-    Returns:
-      list: A list of explanation scores
+      :param model_wrapper wrapper around a model that can be used to calculate explanations
+      :param  dataframe: The data used to calculate the explanations
+      :type dataframe pandas.DataFrame
+      :param categorical_features: list of categorical features not including the target column
+      :param target_column the column to predict
+      :type target_column string
+      :param task_type the task type (regression or classification) of the resulting model
+      :type task_type string
+      :return explanation scores for the input data
+      :rtype list[float]
     """
     # Create the RAI Insights object, use baseline as train and test data
     rai_i: RAIInsights = RAIInsights(
@@ -139,12 +143,12 @@ def compute_explanations(model_wrapper, dataframe, categorical_features, target_
 def calculate_attribution_drift(baseline_explanations, production_explanations):
     """Compute feature attribution drift given two sets of explanations
 
-    Args:
-      baseline_explanations: list of explanations calculated using the baseline dataframe
-      production_explanations: list of explanations calculated using the production dataframe
-
-    Returns:
-      float: the ndcg metric between the baseline and production data
+      :param baseline_explanations list of explanations calculated using the baseline dataframe
+      :type baseline_explanations
+      :param production_explanations list of explanations calculated using the production dataframe
+      :type production_explanations
+      :return the ndcg metric between the baseline and production data
+      :rtype float
     """
     true_relevance = np.asarray([baseline_explanations])
     relevance_score = np.asarray([production_explanations])
@@ -156,16 +160,18 @@ def compute_attribution_drift(task_type, target_column, baseline_dataframe, prod
     """Compute feature attribution drift by calculating feature importances on each
     dataframe input and using these to calculate the ndcg metric
 
-    Args:
-      task_type: str, the task type (regression or classification) of the resulting model
-      target_column: str, the column to predict
-      baseline_dataframe: The baseline dataframe meaning the dataframe used to create the
+      :param task_type The task type (regression or classification) of the resulting model
+      :type task_type string
+      :param target_column the column to predict
+      :type target_column string
+      :param baseline_dataframe The baseline data meaning the data used to create the
       model monitor
-      production_dataframe: The production dataframe meaning the most recent set of data
+      :type baseline_dataframe pandas.DataFrame
+      :param production_dataframe: The production data meaning the most recent set of data
       sent to the model monitor, the current set of data
-
-    Returns:
-      float: the ndcg metric between the baseline and production data
+      :type production_dataframe pandas.DataFrame
+      :return the ndcg metric between the baseline and production data
+      :rtype float
     """
 
     if len(baseline_dataframe.columns.difference(production_dataframe.columns)) > 0:
--- a/responsibleai/model_monitor/components/src/calculate_attribution.py
+++ b/responsibleai/model_monitor/components/src/calculate_attribution.py
@@ -1,0 +1,176 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""This file contains the core logic for feature attribution drift component."""
+import numpy as np
+import mltable
+import pandas as pd
+
+from responsibleai import RAIInsights
+from sklearn.datasets import  load_diabetes
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import ndcg_score
+from ml_wrappers.model.predictions_wrapper import (
+    PredictionsModelWrapperClassification,
+    PredictionsModelWrapperRegression)
+
+try:
+    from lightgbm import LGBMClassifier, LGBMRegressor
+except ImportError:
+    pass
+
+from tabular.components.src._telemetry._loggerfactory import _LoggerFactory, track
+
+_logger = logging.getLogger(__file__)
+_ai_logger = None
+
+
+def _get_logger():
+    global _ai_logger
+    if _ai_logger is None:
+        _ai_logger = _LoggerFactory.get_logger(__file__)
+    return _ai_logger
+
+_get_logger()
+
+
+def create_lightgbm_model(X, y, task_type):
+    """Create model on which to calculate feature importances
+
+    Args:
+      X: list of x values (data excluding target columns)
+      y: list y values (target column data)
+      task_type: str, the task type (regression or classification) of the resulting model
+
+    Returns:
+       An appropriate model.
+    """
+    if(task_type == "classification"): 
+         lgbm = LGBMClassifier(boosting_type='gbdt', learning_rate=0.1,
+                          max_depth=5, n_estimators=200, n_jobs=1, random_state=777)
+    else: 
+        lgbm = LGBMRegressor(boosting_type='gbdt', learning_rate=0.1,
+                         max_depth=5, n_estimators=200, n_jobs=1, random_state=777)
+    model = lgbm.fit(X, y)
+    
+    _logger.info("Created lightgbm model")
+    return model
+
+
+def get_model_wrapper(task_type, target_column, baseline_df, production_df):
+    """Create model wrapper using ml-wrappers on which to calculate feature importances
+    
+    Args:
+      task_type: str, The task type (regression or classification) of the resulting model
+      target_column: str, the column to predict
+      baseline_df: The baseline dataframe
+      production_df: The production dataframe
+
+    Returns:
+       An appropriate model wrapper
+    """
+    y_train = baseline_df[target_column]
+    x_train = baseline_df.drop([target_column], axis=1)
+    x_test = production_df.drop([target_column], axis=1)
+    model = create_lightgbm_model(x_train, y_train, task_type)
+
+    if task_type == 'classification':
+      all_data = pd.concat([x_test, x_train])
+      model_predict = model.predict(all_data)
+      model_predict_proba = model.predict_proba(all_data)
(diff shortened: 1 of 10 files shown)


responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+
+from responsibleai import RAIInsights
+from sklearn.metrics import ndcg_score
+from ml_wrappers.model.predictions_wrapper import (
+    PredictionsModelWrapperClassification,
+    PredictionsModelWrapperRegression)
+
+try:
+    from lightgbm import LGBMClassifier, LGBMRegressor
+except ImportError:
```
romanlutz, line 17: What'll happen in this case? In don't see any special handling later on.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def get_model_wrapper(task_type, target_column, baseline_df, production_df):
+    """Create model wrapper using ml-wrappers on which to calculate feature importances
+
+    Args:
+      task_type: str, The task type (regression or classification) of the resulting model
+      target_column: str, the column to predict
+      baseline_df: The baseline dataframe
+      production_df: The production dataframe
```
romanlutz, line 66: I'm guessing "production" refers to data that came in while monitoring? What is "baseline"? A previous time period's data? Would be nice to have a brief comment I think.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      baseline_df: The baseline dataframe
+      target_column: str, the column to predict
+
+    Returns:
+      list: A list of categorical features.
+    """
+    categorical_features = []
+    for column in baseline_df.columns:
+        baseline_column = pd.Series(baseline_df[column])
+        if baseline_column.dtype.name == 'object' and baseline_column.name != target_column:
```
romanlutz, line 106: what about the "category" dtype?

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      baseline_explanations: list of explanations calculated using the baseline dataframe
+      production_explanations: list of explanations calculated using the production dataframe
+ 
+    Returns:
+      float: the ndcg metric between the baseline and production data
+    """
+    true_relevance = np.asarray([baseline_explanations])
+    relevance_score = np.asarray([production_explanations])
+    feature_attribution_drift = ndcg_score(true_relevance, relevance_score)
+    print(feature_attribution_drift)
```
romanlutz, line 150: Perhaps it's intended, but usually we don't have print statements. If intended please ignore my comment 🙂

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      model_wrapper: wrapper around a model that can be used to calculate explanations
+      df: The dataframe
+      categorical_features: list of categorical features not including the target column
+      target_column: str, the column to predict
+      task_type: str, the task type (regression or classification) of the resulting model
+
+    Returns:
+      list: A list of explanations.
+    """
+    # Create the RAI Insights object, use either baseline as train and test data
```
romanlutz, line 125:
```suggestion
    # Create the RAI Insights object, use baseline as train and test data
```
This seems like it doesn't belong there.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    model_wrapper = get_model_wrapper(task_type, target_column, baseline_df, production_df)
+
+    categorical_features = compute_categorical_features(baseline_df, target_column)
+
+    baseline_explanations = compute_explanations(model_wrapper, baseline_df, categorical_features, target_column, task_type)
+    _logger.info("Successfully computed explanations for baseline dataset")
+
+    production_explanations = compute_explanations(model_wrapper, production_df, categorical_features, target_column, task_type)
+    _logger.info("Successfully computed explanations for productiond dataset")
```
romanlutz, line 175:
```suggestion
    _logger.info("Successfully computed explanations for production dataset")
```

responsibleai/model_monitor/components/src/io_utils.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""This file contains utilities to read write data."""
```
romanlutz, line 4:
```suggestion
"""This file contains utilities to read or write data."""
```

romanlutz (commented):
Nice! Just a few smaller comments, but I don't see any blockers!

responsibleai/model_monitor/components/src/calculate_attribution.py
imatiach-msft: maybe you can just set a boolean var and throw in the code below if it's invoked to say the env is missing this dependency

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+    # Create the RAI Insights object, use either baseline as train and test data
+    rai_i: RAIInsights = RAIInsights(
+        model_wrapper, df, df, target_column, task_type, categorical_features=categorical_features
+    )
+
+    # Add the explanation and compute
+    rai_i.explainer.add()
+    rai_i.compute()
+
+    return rai_i.explainer._explanation.global_importance_values
```
imatiach-msft, line 134: I think this is a good start... we can probably optimize this code a bit more by using the interpret-community APIs directly or some other explanation method, perhaps in a follow-up PR

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def calculate_attribution_drift(baseline_explanations, production_explanations):
+    """Compute feature attribution drift given two sets of explanations
+
+    Args:
```
imatiach-msft, line 140: Great comments in code.  Interesting how this style differs from most of the other AzureML repos.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      baseline_explanations: list of explanations calculated using the baseline dataframe
+      production_explanations: list of explanations calculated using the production dataframe
+ 
+    Returns:
+      float: the ndcg metric between the baseline and production data
+    """
+    true_relevance = np.asarray([baseline_explanations])
+    relevance_score = np.asarray([production_explanations])
+    feature_attribution_drift = ndcg_score(true_relevance, relevance_score)
```
imatiach-msft, line 149: I feel like this should all live in some sdk, like perhaps in interpret-community eventually, as opposed to just in this component in this repo

responsibleai/model_monitor/components/src/run.py
```diff
@@ -0,0 +1,57 @@
...
+    return _ai_logger
+
+
+_get_logger()
+
+
+def parse_args():
+    # Parse arguments
+    parser = argparse.ArgumentParser()
+    # parser.add_argument("--class_names", type=str)
```
imatiach-msft, line 30: we usually remove commented code (?)

responsibleai/model_monitor/components/src/run.py
```diff
@@ -0,0 +1,57 @@
...
+
+    return args
+
+
+def run(args):
+
+    baseline_df = load_mltable_to_df(args.baseline_data)
+    production_df = load_mltable_to_df(args.production_data)
+
+    # todo: read from args when available
```
imatiach-msft, line 46: I'll assume this will be done in a future PR

imatiach-msft (dismissed):

responsibleai/model_monitor/components/src/run.py
```diff
@@ -0,0 +1,57 @@
...
+    task_type = "classification"
+    target_column = "target"
+
+    compute_attribution_drift(task_type, target_column, baseline_df, production_df)
```
imatiach-msft, line 50: do we do something with this value then?  how is it used or returned?

imatiach-msft: just trying to understand what is the plan for how this drift/ndcg score will be used since it seems like it's not passed to anything else from here

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    if (task_type == "classification"): 
+        lgbm = LGBMClassifier(boosting_type='gbdt', learning_rate=0.1,
+                              max_depth=5, n_estimators=200, n_jobs=1, random_state=777)
+    else: 
+        lgbm = LGBMRegressor(boosting_type='gbdt', learning_rate=0.1,
+                            max_depth=5, n_estimators=200, n_jobs=1, random_state=777)
+    model = lgbm.fit(X, y)
+    
+    _logger.info("Created lightgbm model")
```
gaugup, line 55: Maybe have the type of model (classifier/regressor) also be a part of the debug statement.

gaugup: Also is the intention to send the telemetry to app-inshgts here? Then you should use _ai_logger instead of _logger.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+_get_logger()
+
+
+def create_lightgbm_model(X, y, task_type):
+    """Create model on which to calculate feature importances
+
+    Args:
+      X: list of x values (data excluding target columns)
```
gaugup, line 40: I think X is pandas DataFrame and not a list.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      y: list y values (target column data)
```
gaugup, line 41: y might be a numpy array and not a list

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+    Returns:
+       An appropriate model wrapper
+    """
+    y_train = baseline_df[target_column]
+    x_train = baseline_df.drop([target_column], axis=1)
+    x_test = production_df.drop([target_column], axis=1)
+    model = create_lightgbm_model(x_train, y_train, task_type)
+
+    if task_type == 'classification':
+        all_data = pd.concat([x_test, x_train])
```
gaugup, line 77: Lines 76-77 and 85-86 are identical and could be put out of the individual if/else as common code.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+    _logger.info("Successfully categorized columns")
+    return categorical_features
+
+
+def compute_explanations(model_wrapper, df, categorical_features, target_column, task_type):
+    """Compute explanations (feature importances) for a given dataset
+
+    Args:
+      model_wrapper: wrapper around a model that can be used to calculate explanations
+      df: The dataframe
```
gaugup, line 117: could you use a dataframe or data_df instead of df?

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      categorical_features: list of categorical features not including the target column
+      target_column: str, the column to predict
+      task_type: str, the task type (regression or classification) of the resulting model
```
gaugup, line 120: We usually use `param` and `type` separately to document what the variable does and what type it can take.

https://github.com/microsoft/responsible-ai-toolbox/blob/f5496c9d5d9c3e61ef60e84e36d7678f62e9bd35/responsibleai/responsibleai/rai_insights/rai_insights.py#L137

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+      list: A list of explanations.
+    """
+    # Create the RAI Insights object, use either baseline as train and test data
+    rai_i: RAIInsights = RAIInsights(
+        model_wrapper, df, df, target_column, task_type, categorical_features=categorical_features
+    )
+
+    # Add the explanation and compute
+    rai_i.explainer.add()
+    rai_i.compute()
```
gaugup, line 132: What happens is the code from lines 126-132 fails with exception? How do we handle the situation when explanations can't be calculated.

responsibleai/model_monitor/components/src/io_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+import mltable
+import pandas as pd
+
+
+def load_mltable_to_df(mltable_path: str) -> pd.DataFrame:
```
gaugup, line 10: Do we need this function for a line of code. Why don't you call mltable.load(mltable_path).to_pandas_dataframe() directly from the run.py?

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,177 @@
...
+    y_train = baseline_df[target_column]
+    x_train = baseline_df.drop([target_column], axis=1)
+    x_test = production_df.drop([target_column], axis=1)
+    model = create_lightgbm_model(x_train, y_train, task_type)
+
+    if task_type == 'classification':
+        all_data = pd.concat([x_test, x_train])
+        model_predict = model.predict(all_data)
+        model_predict_proba = model.predict_proba(all_data)
+        model_wrapper = PredictionsModelWrapperClassification(
```
gaugup, line 80: Would we have access to model in this component? Then we can directly use model and not worry about wrapping predictions.

responsibleai/model_monitor/components/src/calculate_attribution.py
hawestra: intended for the time being, but will remove before checking this PR in!

responsibleai/model_monitor/components/src/run.py
hawestra: yep! waiting on updates from model monitoring folks!

responsibleai/model_monitor/components/src/run.py
hawestra: yes, so once model monitoring is set up to support our metric, we would return the score to them but for now, I believe they only support uri folders or mltable as output

responsibleai/model_monitor/components/src/calculate_attribution.py
hawestra: checked with model monitoring team and they don't have a well-defined scenario for when a signal fails to load a metric; this is a good callout! I'll update with a try catch and output some message to the user

responsibleai/model_monitor/components/src/io_utils.py
hawestra: honestly this is copied from the azureml-asset repo; they have a special function for writing from mltable to df. IMO it doesn't make too much of a difference either way...one way its a little more readable and the other it's one less function call. Leaving for now unless you're strongly opposed.

responsibleai/model_monitor/components/src/calculate_attribution.py
hawestra: what do you mean ?

responsibleai/model_monitor/components/src/calculate_attribution.py
hawestra: I pulled this code from https://github.com/microsoft/ml-wrappers/blame/cb7fb17a2ca58d6a8df9d64a6e8d3bdcde2a59f1/tests/common_utils.py and based on git blame it looks like Ilya added this; do you remember why you have try except here (the files from your pr won't load because it's so old : https://github.com/microsoft/ml-wrappers/pull/34/files)

responsibleai/model_monitor/components/src/calculate_attribution.py
romanlutz: These are pandas types, right? `"object"` is a common one, the way I understand it it's kind of the "kitchen sink"/catch-all type, but `"category"` is actually meant for categorical features. That's why I was wondering why `"category"` isn't mentioned here.

romanlutz (dismissed):

responsibleai/model_monitor/components/src/calculate_attribution.py
hawestra: discussed offline; seems like using dtype is generally the way to determine the type and we'd have to classify columns as category which is a bit of overkill in this instance because the categories themselves aren't necessary

responsibleai/model_monitor/components/src/calculate_attribution.py
gaugup: The _ai_logger object (if you look into the code) come from loggerFactory which is a wrapper around telemetry logger. _logger is python only logger and would send logs to python logging file.

responsibleai/model_monitor/components/src/io_utils.py
gaugup: yeah, it is weird in the sense that you have maintain an extra function with docs extra. Not strongly opposed to this.

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,192 @@
...
+def get_model_wrapper(task_type, target_column, baseline_dataframe, production_dataframe):
+    """Create model wrapper using ml-wrappers on which to calculate feature importances
+
+      :param task_type The task type (regression or classification) of the resulting model
+      :type task_type string
+      :param target_column the column to predict
+      :type target_column string
+      :param baseline_dataframe The baseline data meaning the data used to create the
+      model monitor
+      :type baseline_dataframe pandas.DataFrame
```
gaugup, line 70: Missing semicolon after baseline_dataframe. In general, after the variable name there is a ":".

responsibleai/model_monitor/components/src/calculate_attribution.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def _get_logger():
+    global _ai_logger
+    if _ai_logger is None:
+        _ai_logger = _LoggerFactory.get_logger(__file__)
+    return _ai_logger
+
+
+_get_logger()
```
gaugup, line 34: If you don't intend to use _ai_logger, then remove it's initialization.

responsibleai/model_monitor/components/src/run.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # todo: read from args when available
+    task_type = "classification"
+    target_column = "target"
+    try:
+        compute_attribution_drift(task_type, target_column, baseline_df, production_df)
+    except Exception as e:
+        _logger.info("Error encountered when executing feature attribution component: {0}", e)
+
+    _logger.info("Successfully executed the feature attribution component.")
```
gaugup, line 51: Line 51 should go in the try block otherwise even if the code in try fails it will print success message.

gaugup (approved):

gaugup (approved):

PR #440, opened 2023-03-16
Add support for PRS in build config scripts

mastloui-msft:
Parallel run steps leverage `task.code` for the reference to its code asset whilst component uses `code`.
In order to support parallel run steps, I have added support for 
- Detecting the component type in the spec configuration
- Fetching the code directory from task.code rather than code in the case of a parallel run step type

### DIFF
--- a/scripts/azureml-assets/azureml/assets/config.py
+++ b/scripts/azureml-assets/azureml/assets/config.py
@@ -156,6 +156,13 @@ class Config:
         return [path]
 
 
+class ComponentType(Enum):
+    """Enum for path types supported for model publishing."""
+
+    PIPELINE = "component"  # A pipeline component which allows multi-stage jobs.
+    PARALLEL = "parallel" # A parallel component, aka PRSv2.
+    COMMAND = "command" # A command component.
+
 class Spec(Config):
     """Load and access spec file properties.
 
@@ -220,6 +227,9 @@ class Spec(Config):
     @property
     def code_dir(self) -> str:
         """Component code directory."""
+        if(self._yaml.get('type') == ComponentType.PARALLEL):
+            task = self._yaml.get('task')
+            return None if task is None else task.get('code')
         return self._yaml.get('code')
 
     @property
--- a/scripts/azureml-assets/azureml/assets/config.py
+++ b/scripts/azureml-assets/azureml/assets/config.py
@@ -160,8 +160,8 @@ class ComponentType(Enum):
     """Enum for component types."""
 
     PIPELINE = "pipeline"  # A pipeline component which allows multi-stage jobs.
-    PARALLEL = "parallel" # A parallel component, aka PRSv2.
-    COMMAND = "command" # A command component.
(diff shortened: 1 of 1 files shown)


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -156,6 +156,13 @@ def _expand_path(path: Path) -> List[Path]:
         return [path]
 
 
+class ComponentType(Enum):
+    """Enum for component types."""
+
+    PIPELINE = "component"  # A pipeline component which allows multi-stage jobs.
```
sunjolia, line 162: shouldn't this be PIPELINE = "pipeline"?

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -220,6 +227,9 @@ def image(self) -> str:
     @property
     def code_dir(self) -> str:
         """Component code directory."""
+        if(self._yaml.get('type') == ComponentType.PARALLEL):
```
elesel, line 230:
```suggestion
        if(self._yaml.get('type') == ComponentType.PARALLEL.value):

```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -156,6 +156,13 @@ def _expand_path(path: Path) -> List[Path]:
...
 
 
+class ComponentType(Enum):
+    """Enum for component types."""
+
+    PIPELINE = "component"  # A pipeline component which allows multi-stage jobs.
+    PARALLEL = "parallel" # A parallel component, aka PRSv2.
+    COMMAND = "command" # A command component.
+
 class Spec(Config):
```
elesel, line 166:
```suggestion
    COMMAND = "command" # A command component.


class Spec(Config):
```

scripts/azureml-assets/azureml/assets/config.py
mastloui-msft: yes

sunjolia: Also I'm assuming the enum is being added just for future-proofing, since the other values aren't being used right now?

elesel (changes requested):

scripts/azureml-assets/azureml/assets/config.py
mastloui-msft: yeap!

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -220,6 +228,10 @@ def image(self) -> str:
     @property
     def code_dir(self) -> str:
         """Component code directory."""
+        if(self._yaml.get('type') == ComponentType.PARALLEL.value):
+
+            task = self._yaml.get('task')
```
elesel, line 233: Sorry, GitHub added that extra blank line. Also, reformatted `if` for consistency with the rest of the file.
```suggestion
        if self._yaml.get('type') == ComponentType.PARALLEL.value:
            task = self._yaml.get('task')
```

PR #451, opened 2023-03-21
add component to calculate feature importance

hawestra:
add feature importance component and update feature attribution to write output to mltable


responsibleai/model_monitor/components/src/compute_attribution_drift.py
```diff
@@ -0,0 +1,130 @@
...
+      model monitor
+      :type baseline_dataframe: pandas.DataFrame
+      :param production_dataframe: The production data meaning the most recent set of data
+      sent to the model monitor, the current set of data
+      :type production_dataframe: pandas.DataFrame
+      :return: the ndcg metric between the baseline and production data
+      :rtype: float
+    """
+    if len(baseline_dataframe.columns.difference(production_dataframe.columns)) > 0:
+        raise UserErrorException("Dataset columns differ in baseline and production datasets")
```
gaugup, line 80: Should this be SystemError?

gaugup (approved):

responsibleai/model_monitor/components/src/compute_attribution_drift.py
hawestra: imo, no because the user provides both the baseline and production data so if they don't match that's a user issue

PR #452, opened 2023-03-21
add bert pretrain test

cassieesvelt:
Add BERT pretrain as a test for training curated environments.


scripts/setup/create_azure_resources_daily.sh
```diff
@@ -51,6 +52,13 @@ else
         echo "Creating ${resource_name}"
         az ml compute create --name $gpu_cluster --size Standard_NC6 --min-instances 0 --max-instances 10 --type AmlCompute
     fi
+
+    resource_name="compute cluster ${gpu_v100_cluster}"
+    echo "Checking ${resource_name}"
+    if ! az ml compute show --name $gpu_v100_cluster --output none >/dev/null 2>&1; then
+        echo "Creating ${resource_name}"
+        az ml compute create --name $gpu_v100_cluster --size Standard_ND40rs_v2 --min-instances 0 --max-instances 2 --type AmlCompute
```
elesel, line 60: Is this absolutely necessary? We have a limited budget for the resources that support the GitHub workflows and the NDv2 series VMs are very expensive to run.

cassieesvelt: No, not needed, we just need v100s because of the cuda versions in the ACPT environments. Maybe we could use Standard_NC12s_v3 ?

elesel: Would you please test to see whether Standard_NC6s_v3 would work for the test? If you just need a v100 that's the lowest cost option.

elesel (approved):

savitamittal1 (approved):

rtanase:
@cassieesvelt, pls consolidate all the ACPT tests scripts into a single parameterized python file that can be reused across all environments. that way we eliminate the duplication.

rtanase (dismissed):

savitamittal1 (approved):

PR #463, opened 2023-03-29
support for license and lineage

HrishikeshGeedMS:


novaturient95 (approved):

training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -20,6 +20,8 @@ command: >
   --mlflow-flavor transformers
   --model-path ${{inputs.model_path}}
   --mlflow-model-output-dir ${{outputs.mlflow_model_folder}}
+  --model-job-path ${{outputs.model_job_path}}
```
skasturi, line 23: >model-job-path [](http://example.com/codeflow?start=4&length=14)

What does "model-job-path" correspond to?

skasturi: May be rename to model import job or something like that?

skasturi (approved):
:shipit:

PR #488, opened 2023-04-02
Fix Model evaluation environment version - Set to 1 was earlier 0.0.1.preview -> auto failing "auto" update

skasturi:


skasturi:
@vizhur @elesel I observed the asset release is failing when parsing environment version. 

This seems to be happening since evaluation environment version has been set to 0.0.1-preview in the first release. 

In subsequent PR: #483 it was set to auto. Therefore, while trying to figure out the earlier version & update it, the version is unable to be parsed as it's expected to be int. 

image

I tried this PR to see if that's fixed and it looks like it. 

image

ShipraJain01 (approved):

PR #490, opened 2023-04-03
Add auto versioning support for models

novaturient95:


novaturient95:
Using fetching asset versions in PR: #453

elesel (changes requested):
With this PR there's no good way to map asset versions in a registry back to the definition in GitHub (assuming prod registry). I'll bring this up in Teams.

novaturient95:
> With this PR there's no good way to map asset versions in a registry back to the definition in GitHub (assuming prod registry). I'll bring this up in Teams.

Thanks Louie. Why is there a need to have a mapping in github? Is there someway we can do for models as well?

PR #527, opened 2023-04-10
Updating ACFT-HF-NLP components to 0.0.6

skanakamedal:


training/finetune_acft_hf_nlp/components/finetune/question_answering/spec.yaml
```diff
@@ -1,14 +1,14 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 name: question_answering_finetune
-version: 0.0.5
+version: 0.0.6
```
vizhur, line 3: would it be possible to start with non-zero major version in public registry?

PR #534, opened 2023-04-11
[Model Evaluation] Update component versions for PuP

ankushbhatia2:


training/model_evaluation/components/model_prediction/spec.yaml
```diff
@@ -79,7 +78,7 @@ outputs:
     type: mltable
 
 code: ../../src
-environment: azureml://registries/azureml/environments/model-evaluation/versions/4
+environment: azureml://registries/azureml/environments/model-evaluation/versions/3
```
ShipraJain01, line 81: why there is difference between environment versions here?

ShipraJain01: in remaining 2 components we are mapping version 4 but in Model Prediction we are mapping to version 3. why?

ShipraJain01 (approved):

novaturient95 (dismissed):

skasturi (dismissed):

ShipraJain01 (approved):

PR #553, opened 2023-04-14
deprecate cuda 11.5 environment

savitamittal1:


training/general/src/environments/acpt-pytorch-1.11-cuda11.5/spec.yaml
```diff
@@ -14,6 +13,8 @@ build:
 os_type: linux
 
 tags:
+  Deprecated: ""
+  DeleteOn: "2023-04-14"
```
vizhur, line 17: We are not deleting it. We do not support metadata update yet (i.e. archive)

savitamittal1: ok, so just keep the deprecate tag?

savitamittal1: Also how do we make sure, its removed from vulnerabilities?

elesel (approved):

training/general/src/environments/acpt-pytorch-1.11-cuda11.5/spec.yaml
savitamittal1: thanks @elesel . I have removed the tag 'DeleteOn'

cassieesvelt (approved):

PR #569, opened 2023-04-18
read me generated using script

HrishikeshGeedMS:
This PR contains the readme docs generated using python script - https://github.com/ManojBableshwar/azureml-assets/blob/main/basic/tools/component_readme.py


common/components/deploy_model/spec.yaml
```diff
@@ -1,6 +1,6 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 name: deploy_model
-version: 0.0.1
+version: 0.0.2
```
novaturient95, line 3: Could you please share why is version update required? Are we going to update spec's description as well to contain `See [docs](https://aka.ms/azureml/components/register_model) to learn more.`?

HrishikeshGeedMS: Yes, Since we are updating spec's descrciption , we need to update comp version as well

skasturi: Aren't we now able to update description without having to update the version?

HrishikeshGeedMS: This version update is not for model, this is because since we are changing the spec.yaml, we need to update the component version.

PR #570, opened 2023-04-18
Update model card versions

susridhar:


models/system/bert-base-cased/spec.yaml
```diff
@@ -11,4 +11,4 @@ tags:
   license: apache-2.0
   min_inference_sku: Standard_DS2_v2
   task: fill-mask
-version: 2
+version: 3
```
novaturient95, line 14: @susridhar this will not be required as prod still pointing to version 1.  @jpmann is updating model card from finetuning end as well. We will wait till that PR is merged and then do a release to update models to version 2

cc: @skasturi @praveenmathamsetty

PR #581, opened 2023-04-24
Update container path for whisper model

novaturient95:


skasturi:
Could you please share details of what has changed in the description? 

Are we also changing the environment to support batch inference?

skasturi (dismissed):

HrishikeshGeedMS (approved):

skasturi (approved):

PR #588, opened 2023-04-25
Oai_pipeline_correction

vikasagrawal-ms:
Pipeline Component name correction from oaie-> oai


openai/components/oai_finetune_pipeline_component.yml
```diff
@@ -1,5 +1,7 @@
 $schema: https://azuremlschemas.azureedge.net/latest/pipelineComponent.schema.json
-name: microsoft_com_azureml_oaie_finetune_pipeline_v2
+
+name: microsoft_com_azureml_oai_finetune_pipeline_v2
```
novaturient95, line 3: @elesel can you please check component naming as well? Should *microsoft_com_azureml* be omitted?

novaturient95: image

https://github.com/Azure/azureml-assets/wiki/Asset-naming-convention

novaturient95: @vikasagrawal-ms please check if component name can follow above standards. Thanks

vikasagrawal-ms: @novaturient95 , The name is ok, as reviewed by Naveen.

novaturient95: @elesel can you please confirm

elesel: @vikasagrawal-ms, please remove the `microsoft_com_azureml_` prefix. What @novaturient95 shared is correct.

elesel: Also, versioning (v2) should be in the version field, not the environment name, unless there's a compelling reason otherwise.

elesel (changes requested):
Name prefixes need to be removed.

openai/components/oai_finetune_pipeline_component.yml
elesel, line 1: This component should also have an asset.yaml file accompanying it. See https://github.com/Azure/azureml-assets/wiki/Environments#assetyaml

openai/components/oai_finetune_pipeline_component.yml
vikasagrawal-ms: @elesel v2 is to suggest its DPv2 supported component

elesel: All registry components are DPv2. Probably not worth including v2 in the name.

openai/components/oai_finetune_pipeline_component.yml
vikasagrawal-ms: Ack

openai/components/oai_finetune_pipeline_component.yml
vikasagrawal-ms: Updated the name with openai_completions_pipeline after deliberation with PMs

elesel (approved):

openai/components/asset.yaml
elesel, line 1: Even though there's only one component so far, please move these two files into their own `oai_finetune_pipeline_component` subdirectory under openai/components.

openai/components/asset.yaml
```diff
@@ -0,0 +1,2 @@
+type: component
+spec: oai_finetune_pipeline_component.yaml
```
elesel, line 2: Please rename oai_finetune_pipeline_component.yaml to spec.yaml to keep it generic after the files are moved under the subdir.

elesel (changes requested):

openai/components/asset.yaml
vikasagrawal-ms: ok

openai/components/asset.yaml
vikasagrawal-ms: ok

novaturient95 (approved):

PR #590, opened 2023-04-26
version updated due to model management environment updation

HrishikeshGeedMS:
This includes the change in the version of model management components, since environment has been updated

### DIFF
--- a/training/model_management/components/convert_model_to_mlflow/spec.yaml
+++ b/training/model_management/components/convert_model_to_mlflow/spec.yaml
@@ -9,7 +9,7 @@ is_deterministic: True
 display_name: Convert Hugging face models to MLFlow
 description: Component converts Huggingface models to MLFlow model packaging format
 
-environment: azureml://registries/azureml/environments/model-management/versions/4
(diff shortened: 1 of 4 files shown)


training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -9,7 +9,7 @@ is_deterministic: True
 display_name: Convert Hugging face models to MLFlow
 description: Component converts Huggingface models to MLFlow model packaging format
 
-environment: azureml://registries/azureml/environments/model-management/versions/4
+environment: azureml://registries/azureml/environments/model-management/versions/5
```
novaturient95, line 12: Why not update version here?

novaturient95: conver_model_to_mlflow is version is updated but not published to prod

novaturient95, line 12: Change env version to 6

skasturi (approved):

PR #591, opened 2023-04-26
Update model output signature

novaturient95:


novaturient95:
we will wait for this change as this requires changes in model card as well for all models.

@praveenmathamsetty

novaturient95:
Hi @ManojBableshwar @swatig007 can you please help to review this PR? Bug was raised in specific to whisper model to have output be in following format: 
`{"result": "this is the transcribed output"}` or `{"text": "this is the transcribed output"}`

instead of `{0: "this is the transcribed output"}`

PR #604, opened 2023-05-02
Fix [Live site bug] <class 'TypeError'> "Could not find a version that satisfies the requirement" when load mlflow model is called

tongyu-microsoft:
The root cause is, the output of subprocess.CalledProcessError is bytes-like object, not a str. So if we want to check if certain string is in the output, we need to convert str to bytes-like object.

### DIFF
--- a/responsibleai/tabular/components/src/rai_component_utilities.py
+++ b/responsibleai/tabular/components/src/rai_component_utilities.py
@@ -129,10 +129,10 @@ def load_mlflow_model(
 
 def _classify_and_log_pip_install_error(elog):
     if elog is not None:
-        if "Could not find a version that satisfies the requirement" in elog:
+        if b"Could not find a version that satisfies the requirement" in elog:
             _logger.warning("Detected unsatisfiable version requirment.")
 
-        if "package versions have conflicting dependencies" in elog:
+        if b"package versions have conflicting dependencies" in elog:
             _logger.warning("Detected dependency conflict error.")
 
 


gaugup (commented):
Why do we need this change? Can we fix the issue in RAI-vNext and then sync the code from there?

romanlutz:
> Why do we need this change? Can we fix the issue in RAI-vNext and then sync the code from there?

I'm not sure I understand the comment. We do duplicate all the code in both places, don't we? So copying the change to this repo makes perfect sense, doesn't it?

FWIW, I've pointed out the problem of having two identical code bases (in the best case) on several occasions. I will do so again in our next team meeting (Friday next week is reserved I think). Of course, if you check change by change you will find that the code bases have already diverged...

romanlutz (approved):
Amazing! Would love to hear how you figured that one out!

imatiach-msft (approved):

tongyu-microsoft:
> Amazing! Would love to hear how you figured that one out!

Yeah initially we were thinking the `e.output` might be None and causing the TypeError, so we had this fix: https://github.com/Azure/RAI-vNext-Preview/pull/177/files
image

But then after we release new component 0.7.0, yesterday we still have one customer hit this issue. So I tried to hit the exception for `subprocess.CalledProcessError` and print out the e.output. Then I hit the exact error as the customer, and found the output is actually type `bytes`, not a string or None

PR #606, opened 2023-05-02
Add new tensorflow environments with the latest version

cassieesvelt:
Add two new tensorflow environments that use the two latest versions of tensorflow. Other tensorflow envs will be deprecated.


training/general/src/environments/tensorflow-2.11-cuda11/context/conda_dependencies.yaml
```diff
@@ -0,0 +1,27 @@
+name: tensorflow-2.11-cuda11
+channels:
+- anaconda
```
vizhur, line 3: I'd prioritize conda-forge channel

cassieesvelt: changed the order

savitamittal1 (approved):

PR #609, opened 2023-05-03
Skip publishing of existing assets

novaturient95:


elesel (approved):

skasturi (approved):

skasturi:
As discussed offline, metadata only updates will need model version update after this change. 

However, that is being worked upon by @elesel 

Cc: @vizhur

PR #617, opened 2023-05-05
Small doc script updates

vizhur:


scripts/docs/generate_asset_documentation.py
```diff
@@ -72,23 +158,31 @@ def add_intro(doc, asset):
         if "languages" in asset["properties"]:
             doc.add_paragraph("**Languages**: " + asset["properties"]["languages"])
 
+    if "tags" in asset:
+        doc.add_heading("Tags ", level=3)
+        tags = []
+        for tag, value in asset["tags"].items():
+            tags.append("`{}` ".format(tag if len(str(value)) == 0 else (tag + " : " + str(value))))
+        tags.sort()
```
lykelly19, line 166: `tags.sort(key=lambda x: x[0].lower())` (to avoid case-insensitivty although I'm not sure if tags can be lower/upper-case, this will also need to be done for references later too

scripts/docs/generate_asset_documentation.py
```diff
@@ -36,27 +39,110 @@ def create_asset_doc(spec_file_name, asset_type):
...
+    doc.add_paragraph("**Docker image**: " + img)
+    return doc
+
+
+def add_docker_context(doc, asset, asset_config):
+    """Add docker context content."""
+    doc.add_heading("Docker build context", level=2)
+    context = {}
+    dockerfile = ""
+    for context_file in search(str(asset_config.extra_config_as_object().context_dir_with_path) + "/*"):
```
elesel, line 106: You should be able to use pathlib here too: https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob
```suggestion
    for context_file in asset_config.extra_config_as_object().context_dir_with_path.glob("**"):
```
and then you might need to clean up the filenames afterward.

scripts/docs/generate_asset_documentation.py
```diff
@@ -36,27 +39,110 @@ def create_asset_doc(spec_file_name, asset_type):
...
+        content = ""
+        with open(context_file, "r") as f:
+            # do not expect nested structure for now
+            content = f.read()
+        filename = os.path.basename(context_file)
```
elesel, line 111: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to should help here.

vizhur: will have that as a follow up!

scripts/docs/generate_asset_documentation.py
```diff
@@ -36,27 +39,110 @@ def create_asset_doc(spec_file_name, asset_type):
...
+        if filename.lower() == "dockerfile":
```
elesel, line 112: Dockerfile name can be specified in the environment.yaml file (readable via asset_config.extra_config_as_object().dockerfile_with_path), so you could compare against it instead of doing this check.

elesel (approved):

scripts/docs/generate_asset_documentation.py
vizhur: it didn't work, good point, I'll follow up on that

scripts/docs/generate_asset_documentation.py
vizhur: Saw that, I plan to follow up to support nested structure. Thanks for the suggestion about relative_to from pathlib, that will be handy

elesel (approved):

PR #620, opened 2023-05-05
Categorization for each asset type

lykelly19:


models/system/openai-whisper-large/model.yaml
```diff
@@ -1,6 +1,6 @@
 path:
   container_name: models
-  container_path: huggingface/openai-whisper-large/27b5bb3/mlflow_model_folder
+  container_path: huggingface/openai-whisper-large/6fdb54a/mlflow_model_folder
```
novaturient95, line 3: Looks like a lot of previous changes are present. You may need to take a latest pull from main

PR #623, opened 2023-05-08
Add flag to skip assets validation while publishing

novaturient95:


scripts/release/asset_publish.py
```diff
@@ -490,6 +504,10 @@ def _str2bool(v: str) -> bool:
         "-d", "--debug", type=_str2bool, nargs="?",
         const=True, default=False, help="debug mode",
     )
+    parser.add_argument(
+        "-va", "--validate-assets", type=_str2bool, nargs="?",
```
elesel, line 508: This arg appears to skip validation and update of spec files. Consider renaming the flag to something like `--skip-update-spec`.

scripts/release/asset_publish.py
```diff
@@ -563,13 +584,13 @@ def _str2bool(v: str) -> bool:
                     component_type = asset.spec_as_object().type
                     if component_type == assets.ComponentType.PIPELINE.value:
                         if not validate_and_prepare_pipeline_component(
-                            asset.spec_with_path, passed_version, registry_name
+                            asset.spec_with_path, passed_version, registry_name, validate_assets
```
elesel, line 587: Why pass it into the function only to have it exit immediately? Just use it here to skip the call to `validate_and_prepare_pipeline_component` instead.

elesel (changes requested):

scripts/release/asset_publish.py
novaturient95: Crossed my mind as well but thought too many if conditions in that part of code. Updated now. Let me know if you have any suggestions.

scripts/release/asset_publish.py
```diff
@@ -490,6 +490,10 @@ def _str2bool(v: str) -> bool:
         "-d", "--debug", type=_str2bool, nargs="?",
         const=True, default=False, help="debug mode",
     )
+    parser.add_argument(
+        "-sk", "--skip-update-spec", type=_str2bool, nargs="?",
+        const=False, default=True, help="Skip asset's spec update",
```
elesel, line 495: Should probably be:
```suggestion
        const=True, default=False, help="Skip asset's spec update",
```

scripts/release/asset_publish.py
```diff
@@ -490,6 +490,10 @@ def _str2bool(v: str) -> bool:
         "-d", "--debug", type=_str2bool, nargs="?",
         const=True, default=False, help="debug mode",
     )
+    parser.add_argument(
+        "-sk", "--skip-update-spec", type=_str2bool, nargs="?",
```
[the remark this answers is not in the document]
novaturient95: Done. thank you

elesel (approved):

PR #626, opened 2023-05-09
Docs: add initial classes for categories

vizhur:


scripts/docs/generate_asset_documentation.py
```diff
@@ -333,6 +333,69 @@ def doc(self) -> snakemd.document.Document:
...
+    _categories = {}
+
+    def __init__(self):
+        self._categories[AssetType.ENVIRONMENT.value] = {}
+        self._categories[AssetType.COMPONENT.value] = {}
+        self._categories[AssetType.MODEL.value] = {}
+
+    def classify_asset(self, asset: AssetInfo):
+        for category in asset.categories:
+            cats = category.split("/")
```
elesel, line 347: Can you give an example of a category value that includes a / ?

scripts/docs/generate_asset_documentation.py
```diff
@@ -333,6 +333,69 @@ def doc(self) -> snakemd.document.Document:
...
+    _assets = []
+    _categories = {}
+    _parent = None
+
+    def __init__(self, name: str, parent=None):
+        self._name = name
+        self._parent = parent
+
+    def add_asset(self, asset: AssetInfo, sub_categories:List[str]):
+        if len(sub_categories)>0:
```
elesel, line 387: Or:
```suggestion
        if sub_categories:
```

scripts/docs/generate_asset_documentation.py
lykelly19: I think this might be for the case where there are subcategories

elesel (approved):

elesel (approved):

elesel (approved):

PR #640, opened 2023-05-11
update model import components to include local validation + deployment + inference validation

novaturient95:


training/model_management/components/import_model/spec.yaml
```diff
@@ -75,6 +88,154 @@ inputs:
...
+      - Standard_NC12s_v2
+      - Standard_NC12s_v3
+      - Standard_NC16as_T4_v3
+      - Standard_NC24s_v2
+      - Standard_NC24s_v3
+      - Standard_NC64as_T4_v3
+      - Standard_ND40rs_v2
+      - Standard_ND96asr_v4
+      - Standard_ND96amsr_A100_v4
+    default: Standard_NC24s_v3
```
skasturi, line 148: Thanks for setting the default to NC24/V100 GPU.

common/src/model_deployment/deploy.py
```diff
@@ -234,31 +221,85 @@ def main(args):
...
+    model_info = {}
+    with open(args.registration_details) as f:
+        model_info = json.load(f)
+    model_id = model_info["id"]
+    model_name = model_info["name"]
+
+    # make sure underscores and slashes are replaced by hyphens and convert them to lower case
+    endpoint_name = re.sub('[/_ ]', '-', model_name)
+    endpoint_name = f"{endpoint_name.lower()}-{int(time.time())}"
+    endpoint_name = endpoint_name[:32].lower()
```
elesel, line 269: A long model name could cause this to be non-unique. Consider adding the time suffix after truncating the endpoint name. Also, no need to lowercase the endpoint name twice.

common/src/model_deployment/deploy.py
```diff
@@ -234,31 +221,85 @@ def main(args):
...
+
+def main(args):
+    """Run main function."""
+    ml_client = get_ml_client()
+    # get registered model id
+    model_info = {}
+    with open(args.registration_details) as f:
+        model_info = json.load(f)
+    model_id = model_info["id"]
+    model_name = model_info["name"]
```
elesel, line 264: Nit: We're generally using single quotes around dict keys in this repo.

common/src/model_deployment/deploy.py
```diff
@@ -149,19 +157,12 @@ def parse_args():
         parser.error("Arg max_concurrent_requests_per_instance cannot be less than 1")
     if args.request_timeout_ms < 1 or args.request_timeout_ms > 90000:
         parser.error("Arg request_timeout_ms should lie between 1 and 90000")
-    if args.max_queue_wait_ms < 1 or args.max_queue_wait_ms > 500:
-        parser.error("Arg max_queue_wait_ms should lie between 1 and 500")
+    if args.max_queue_wait_ms < 1 or args.max_queue_wait_ms > 90000:
+        parser.error("Arg max_queue_wait_ms should lie between 1 and 90000")
```
elesel, line 161: Is 60000 or 90000 the max? Consider creating a constant to ensure these values are consistent.

common/src/model_deployment/deploy.py
```diff
@@ -234,31 +221,85 @@ def main(args):
...
+    print("Saving deployment details ...")
+
     # write deployment details to file
     endpoint_type = "aml_online_inference"
-    endpoint_response = ml_client.online_endpoints.get(endpoint.name)
     deployment_details = {
         "endpoint_name": endpoint.name,
         "deployment_name": deployment.name,
-        "endpoint_uri": endpoint_response.__dict__["_scoring_uri"],
+        "endpoint_uri": endpoint.__dict__["_scoring_uri"],
```
elesel, line 302: What's going on here?

training/model_management/components/import_model/spec.yaml
```diff
@@ -75,6 +88,154 @@ inputs:
...
+    type: integer
+    default: 10
+    optional: true
+    description: The number of seconds after the container has started before the readiness probe is initiated
+
+  deployment_failure_threshold_liveness_probe:
+    type: integer
+    default: 30
+    optional: true
+    description: No of times system will try after failing the liveness probe
```
elesel, line 209:
```suggestion
    description: Number of times system will try after failing the liveness probe
```

common/src/model_deployment/deploy.py
novaturient95: Done

common/src/model_deployment/deploy.py
novaturient95: Removed fetching the endpoint details again, and sending response which we receive as part of endpoint traffic update.

common/src/model_deployment/deploy.py
novaturient95: Its 90K. Updated at places where 60K was mentioned as max value.

skasturi: Could we dynamically read this variable from somewhere? 

Also +1 to @elesel's suggestion on using a constant.

common/components/deploy_model/spec.yaml
```diff
@@ -121,15 +121,15 @@ inputs:
 
   max_queue_wait_ms:
     type: integer
-    default: 500
+    default: 60000
```
elesel, line 124: Should probably be 90000 to match the script.
```suggestion
    default: 90000
```

elesel: Never mind. 60K is the default.

common/src/model_deployment/deploy.py
novaturient95: > Could we dynamically read this variable from somewhere?

Not sure, let me figure out and add it in another PR. Whatever public docs we have just show default values and not max value that can be accepted

@skasturi

skasturi (approved):

PR #648, opened 2023-05-12
Alphabetize assets and categories & add links

lykelly19:


scripts/docs/generate_asset_documentation.py
```diff
@@ -416,6 +423,7 @@ def save(self):
             doc.add_heading(self._name.capitalize(), level=1)
 
         if self._sub_categories:
+            self._sub_categories = dict(sorted(self._sub_categories.items(), key=lambda i: i[0].lower()))
```
vizhur, line 426: very minor: maybe not overwrite but create a new dict

scripts/docs/generate_asset_documentation.py
```diff
@@ -304,6 +304,9 @@ def doc(self) -> snakemd.document.Document:
         self._add_doc_asset_version(_doc)
         self._add_doc_tags(_doc)
 
+        link = "https://ml.azure.com/registries/azureml/components/{}/version/{}".format(self.name, self.version)
+        _doc.add_paragraph("**View in Studio**:  [{}]({})".format(link, link))
```
elesel, line 308: You could generalize this with a helper function that also takes the AssetType.

scripts/docs/generate_asset_documentation.py
```diff
@@ -133,6 +133,12 @@ def _add_doc_tags(self, doc):
             tags.sort(key=lambda x: x[0].lower())
             doc.add_raw(" ".join(tags))
         return doc
+    
+    def _add_doc_link(self, doc):
+        link = "https://ml.azure.com/registries/azureml/{}/{}/version/{}".format(
+            self.type + "s", self.name, self.version)
```
elesel, line 139: Consider putting the 's' in the string
```suggestion
        link = "https://ml.azure.com/registries/azureml/{}s/{}/version/{}".format(
            self.type, self.name, self.version)
```

scripts/docs/generate_asset_documentation.py
```diff
@@ -420,12 +418,12 @@ def save(self):
         if self._is_root:
             doc.add_heading(self._type.capitalize() + "s", level=1)
         else:
-            doc.add_heading(self._name.capitalize(), level=1)
+            doc.add_heading(self._name, level=1)
 
         if self._sub_categories:
-            self._sub_categories = dict(sorted(self._sub_categories.items(), key=lambda i: i[0].lower()))
+            sorted_sub_categories = dict(sorted(self._sub_categories.items(), key=lambda i: i[0].lower()))
```
elesel, line 424: You'll probably need an OrderedDict instead of a regular one, to maintain the order in which the items were inserted: https://docs.python.org/3/library/collections.html#collections.OrderedDict

lykelly19: Would it be ordered if we're using Python 3.8? https://stackoverflow.com/questions/34305003/difference-between-dictionary-and-ordereddict

elesel: Nice find! Looks like we're all set, then!

elesel (approved):

PR #650, opened 2023-05-12
Enable category checking in validate_assets.py

lykelly19:


scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -227,11 +227,27 @@ def validate_name(asset_config: assets.AssetConfig) -> int:
...
+    Args:
+        asset_config (AssetConfig): Asset config.
+
+    Returns:
+        int: Number of errors.
+    """
+    if len(asset_config.categories) == 0:
+        _log_error(asset_config, "Categories not found")
+
+    return len(asset_config.categories) == 0
```
elesel, line 242: Maybe a little simpler:
```suggestion
    if not asset_config.categories:
        _log_error(asset_config, "Categories not found")
        return 1
    return 0
```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -227,11 +227,27 @@ def validate_name(asset_config: assets.AssetConfig) -> int:
...
+    """Validate asset categories.
+
+    Args:
+        asset_config (AssetConfig): Asset config.
+
+    Returns:
+        int: Number of errors.
+    """
+    if len(asset_config.categories) == 0:
+        _log_error(asset_config, "Categories not found")
```
elesel, line 240: We should also point to some documentation so that the engineer will know how to add categories. I'm updating the documentation right now and will get back to you shortly.

elesel: I think this would be a good link: https://github.com/Azure/azureml-assets/wiki/Assets#assetyaml

elesel:
@lykelly19 would you please also add the `-C` flag here as part of this PR? https://github.com/Azure/azureml-assets/blob/main/.github/workflows/assets-validation.yaml#L58

scripts/azureml-assets/azureml/assets/validate_assets.py
elesel: I was thinking you could do something like this: https://github.com/Azure/azureml-assets/blob/main/scripts/azureml-assets/azureml/assets/validate_assets.py#L337-L339

lykelly19: ah that makes sense, will make that edit!

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -24,6 +24,9 @@
 INVALID_STRINGS = ["microsoft", ["azureml", "azure"], "aml"]
 COMMON_NAME_PATTERN = re.compile(r"^[a-z][a-z0-9_.-]{0,254}$")
 
+# Asset config convention
+ASSET_CONFIG_CONVENTION_URL = "https://github.com/Azure/azureml-assets/wiki/Assets#assetyaml"
```
elesel, line 28:
```suggestion
ASSET_CONFIG_URL = "https://github.com/Azure/azureml-assets/wiki/Assets#assetyaml"
```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -338,11 +363,16 @@ def validate_assets(input_dirs: List[Path],
     if args.check_names:
         logger.print(f"Asset naming convention: {NAMING_CONVENTION_URL}")
 
+    # Share asset config convention URL
+    if args.check_categories:
+        logger.print(f"Asset config convention: {ASSET_CONFIG_CONVENTION_URL}")
```
elesel, line 368:
```suggestion
    # Share asset config reference
    if args.check_categories:
        logger.print(f"Asset config reference: {ASSET_CONFIG_URL}")
```

elesel (approved):

PR #654, opened 2023-05-15
update ft recommended sku as Standard_NC24rs_v3

esrao:
Fix for [Bug 2375196](https://msdata.visualstudio.com/Vienna/_workitems/edit/2375196): Update model tags and samples with the GPU sku recommended by the compute team

Per Pavan - Standard_NC24rs_v3 is the compute we have been using so far for testing.

### DIFF
--- a/common/components/deploy_model/asset.yaml
+++ b/common/components/deploy_model/asset.yaml
@@ -1,2 +1,3 @@
 type: component
 spec: spec.yaml
+categories: ["Models"]
--- a/common/components/deploy_model/spec.yaml
+++ b/common/components/deploy_model/spec.yaml
@@ -94,6 +94,7 @@ inputs:
       - Standard_NC16as_T4_v3
       - Standard_NC24s_v2
       - Standard_NC24s_v3
+      - Standard_NC24rs_v3
       - Standard_NC64as_T4_v3
       - Standard_ND40rs_v2
       - Standard_ND96asr_v4
--- a/models/system/bert-base-uncased/asset.yaml
+++ b/models/system/bert-base-uncased/asset.yaml
@@ -1,3 +1,4 @@
 extra_config: model.yaml
 spec: spec.yaml
 type: model
+categories: ["Models"]
--- a/models/system/bert-base-cased/asset.yaml
+++ b/models/system/bert-base-cased/asset.yaml
@@ -1,3 +1,4 @@
 extra_config: model.yaml
 spec: spec.yaml
 type: model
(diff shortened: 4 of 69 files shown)


common/components/deploy_model/asset.yaml
```diff
@@ -1,2 +1,3 @@
 type: component
-spec: spec.yaml
\ No newline at end of file
+spec: spec.yaml
+categories: ["Models"]
```
esrao, line 3: @skasturi categories section for the modified assets is now required. Please confirm "Models" is the right category to use for these changes.

common/components/deploy_model/spec.yaml
```diff
@@ -87,6 +87,7 @@ inputs:
       - Standard_NC16as_T4_v3
       - Standard_NC24s_v2
       - Standard_NC24s_v3
+      - Standard_NC24rs_v3
```
esrao, line 90: @novaturient95 - is there a reason why you did not add this your change.

novaturient95: It would have been missed. Thanks for adding @esrao

common/components/deploy_model/asset.yaml
novaturient95: @esrao I see categories for environment only. Do we need to add for component and model spec files as well?
cc: @elesel

novaturient95: Ohh I see new changes requires all assets to have categories

elesel: I think "Models" makes sense for this component because it's a component that works with models.

models/system/bert-base-uncased/asset.yaml
```diff
@@ -1,3 +1,4 @@
 extra_config: model.yaml
 spec: spec.yaml
 type: model
+categories: ["Models"]
```
elesel, line 4: I don't think applying the "Models" category to all of the models in this PR makes sense. Maybe "Hugging Face" would be worth considering, as well as "Language" or other things that could be used to group bert-base-cased and similar models.

esrao: will do this next pr

skasturi: Would need help from @ManojBableshwar @swatig007 here. 

We will push "Models" to start with and change as necessary.

jpmann (approved):

skasturi (approved):

PR #666, opened 2023-05-17
added basic workflow to clone azureml-examplkes repo

HrishikeshGeedMS:


elesel (changes requested):
Why is this needed?

HrishikeshGeedMS:
> 
Hi @elesel , We are trying to create a kind of status table(something like this - https://github.com/Azure/azureml-hf/tree/main/status)
, which will show whether we are able to import all the major Hugging face models or not. This process will include
1)Cloning notebook from azureml-examples and use that to import the model(Since this is customer facing notebook)

This requires to create a workflows, so I have just created this pr with small dev to start with

PR #682, opened 2023-05-19
[draft][prompt flow runtime] directly point to a mcr image

zhengfeiwang:


vizhur (changes requested):
won't work

pipelines/environments/promptflow-runtime/spec.yaml
```diff
@@ -6,9 +6,10 @@ description: >-
 name: "{{asset.name}}"
 version: "{{asset.version}}"
 
-build:
-  path: "{{image.context.path}}"
-  dockerfile_path: "{{image.dockerfile.path}}"
+# build:
+#   path: "{{image.context.path}}"
+#   dockerfile_path: "{{image.dockerfile.path}}"
+image: mcr.microsoft.com/azureml/promptflow/promptflow-runtime:latest
```
elesel, line 12: As deployed to the azureml registry, this environment points to the mcr.microsoft.com/azureml/curated/promptflow-runtime:2 image, which was built using this [Dockerfile](https://github.com/Azure/azureml-assets/blob/efa7780257fbc173165aed22614e191d596bf2d1/latest/environment/promptflow-runtime/context/Dockerfile). For reproducability we always pin images to specific immutable (non-"latest") labels or digests. If any updates are needed for the environment those would be included in a newer version of it, rather than using a "latest" label.

elesel: This would also result in the system registry building an image in its ACR using this command: `FROM mcr.microsoft.com/azureml/promptflow/promptflow-runtime:latest`, which is effectively the same thing the current process is doing, except we build the image and promote it to MCR instead. No images are stored in the azureml system registry's ACR.

PR #695, opened 2023-05-24
Remove deployment job from import model pipeline

novaturient95:
**Problem**
Deployment generally has compute quota issues and it's difficult to use it for scale testing.

**Solution**
1. Remove deployment job
2. Use local validation component and have a specific compute provided to it to make sure model can be loaded. This is still an optional step in pipeline, and can be skipped by not passing test data needed for inference

### DIFF
--- a/training/model_management/components/import_model/asset.yaml
+++ b/training/model_management/components/import_model/asset.yaml
@@ -1,2 +1,3 @@
 type: component
 spec: spec.yaml
+categories: ["Import Model"]
--- a/training/model_management/components/import_model/asset.yaml
+++ b/training/model_management/components/import_model/asset.yaml
@@ -1,3 +1,3 @@
 type: component
 spec: spec.yaml
-categories: ["Import Model"]
+categories: ["Models"]
--- a/training/model_management/components/import_model/spec.yaml
+++ b/training/model_management/components/import_model/spec.yaml
@@ -4,14 +4,18 @@ type: pipeline
 name: import_model
 display_name: Import model
 description: Import a model into a workspace or a registry
-version: 0.0.5
+version: 0.0.6
 
 # Pipeline inputs
 inputs:
   # pipeline specific compute
-  compute:
+  compute_common:
     type: string
-    description: Compute to run pipeline job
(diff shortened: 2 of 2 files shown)


HrishikeshGeedMS (approved):

training/model_management/components/import_model/asset.yaml
```diff
@@ -1,2 +1,3 @@
 type: component
 spec: spec.yaml
+categories: ["Import Model"]
```
skasturi, line 3: Are we listing broad level categories somewhere?

training/model_management/components/import_model/spec.yaml
```diff
@@ -4,14 +4,18 @@ type: pipeline
...
 display_name: Import model
 description: Import a model into a workspace or a registry
-version: 0.0.5
+version: 0.0.6
 
 # Pipeline inputs
 inputs:
   # pipeline specific compute
-  compute:
+  compute_common:
```
skasturi, line 12: What is the reason to rename this input?

training/model_management/components/import_model/spec.yaml
```diff
@@ -4,14 +4,18 @@ type: pipeline
...
     type: string
-    description: Compute to run pipeline job
+    description: Common compute for model download, mlflow conversion and registration
+
+  compute_local_validation:
```
skasturi, line 16: Let us get the interface reviewed by @ManojBableshwar

skasturi (changes requested):
Need interface review

training/model_management/components/import_model/asset.yaml
novaturient95: Hi @elesel could you please help to respond to above query?

novaturient95: Thanks @elesel. Updated categories to `Models`

training/model_management/components/import_model/spec.yaml
novaturient95: Reverted

skasturi (approved):

vikasagrawal-ms (approved):
Approved

PR #704, opened 2023-05-28
[Model Monitor] - Add model monitoring component CI which runs unit test.

mastloui-msft:
- Adding a new workflow file which runs the model monitoring unit tests.
- Workflow runs only for changes under the model_monitoring/ folder and changes to the file itself.

### DIFF
--- a/.github/workflows/model-monitoring-ci.yml
+++ b/.github/workflows/model-monitoring-ci.yml
@@ -1,15 +1,16 @@
 name: model-monitoring-ci
 
 on:
-  push:
+  pull_request:
     branches:
-      - main
-    paths: 
+      - main 
+    paths:
       - model_monitoring/**
-  workflow_dispatch:
+      
 
-concurrency: ${{ github.workflow }}
-
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/model-monitoring-ci.yml
```diff
@@ -0,0 +1,88 @@
...
+  id-token: write
+
+defaults:
+  run:
+    shell: cmd
+
+jobs:
+  run-e2e-tests:
+    name: Run Monitoring Component Tests
+    runs-on: windows-latest
```
elesel, line 35: Does it specifically need Windows? This will be the first non-Linux test in the repo, I believe.

elesel (dismissed):

.github/workflows/model-monitoring-ci.yml
mastloui-msft: let me see if ubuntu works! It used to need windows due to the azure cli task in ADO :)

mastloui-msft: Looks like a lot of the commands are still windows-specific - I'll do a follow-up to convert to ubuntu once the e2e is in :)

elesel: Sounds good. Thanks for testing.

elesel (approved):

PR #707, opened 2023-05-30
Update Dolly Model

ankushbhatia2:


models/system/databricks-dolly-v2-12b/spec.yaml
```diff
@@ -17,4 +17,4 @@ tags:
   Preview: ''
   license: mit
```
novaturient95, line 18: @ankushbhatia2 as discussed lets change this to apache?

novaturient95: anything we need to update in properties specific to finetuning and evaluation min sku's?

jpmann (approved):

models/system/databricks-dolly-v2-12b/spec.yaml
```diff
@@ -15,6 +15,6 @@ properties:
 tags:
   Featured: ''
   Preview: ''
-  license: mit
+  license: apache
```
novaturient95, line 18: Please use `apache-2.0` 
https://github.com/Azure/azureml-assets/blob/main/models/system/bert-base-cased/spec.yaml

ShipraJain01 (approved):

PR #721, opened 2023-06-02
Revert dolly update

skasturi:


models/system/databricks-dolly-v2-12b/spec.yaml
```diff
@@ -4,9 +4,9 @@ path: ./
 properties:
   SHA: d0aa7ea43dc3548a8b499cc88605e150c4a2704d
   datasets: databricks/databricks-dolly-15k
-  evaluation-min-sku-spec: 6|0|56|112
-  evaluation-recommended-sku: Standard_DS5_v2
-  inference-min-sku-spec: 6|0|56|112
+  evaluation-min-sku-spec: 8|0|28|56
```
haritha-bendapudi, line 7: I have reduced the number of cores, so that during the UI sku filter, skus with >56 GB ram but lower number of cores can be used.

models/system/databricks-dolly-v2-12b/spec.yaml
```diff
@@ -4,9 +4,9 @@ path: ./
...
+  evaluation-recommended-sku: Standard_DS4_v2
+  inference-min-sku-spec: 4|0|56|112
```
haritha-bendapudi, line 9: I have reduced the number of cores, so that in the UI vm filter, SKUs with lower cores but sufficient memory are not filtered.

skasturi: I see that you pushed a commit to fix it. Thank you!

haritha-bendapudi (approved):

vizhur (approved):

haritha-bendapudi (approved):

jpmann (approved):

PR #723, opened 2023-06-03
Model card inference updates

susridhar:


skasturi:
You might have to recreate the PR without using a fork so that the validations run.

PR #742, opened 2023-06-12
Update Dolly Model spec

haritha-bendapudi:


skasturi:
@haritha-bendapudi Could you please target the "main" branch?

PR #747, opened 2023-06-12
Skip unreleased environment versions

lykelly19:


scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -30,10 +30,15 @@ def copy_asset(asset_config: assets.AssetConfig,
...
         return None
+    
+    if not assets.check_new_or_preview_release(asset_config) or not assets.previous_release_tag_exists(asset_config, release_directory_root):
+        # Skip if previous version was not released
+        return None
 
     # Copy asset to output directory
-    util.copy_asset_to_output_dir(asset_config=asset_config, output_directory=output_directory_root, add_subdir=True,
-                                  use_version_dir=use_version_dir)
+    # util.copy_asset_to_output_dir(asset_config=asset_config, output_directory=output_directory_root, add_subdir=True,
```
lykelly19, line 39: Commented out for testing purposes, will uncomment line later

scripts/azureml-assets/azureml/assets/update_assets.py
```diff
@@ -63,6 +65,40 @@ def release_tag_exists(asset_config: assets.AssetConfig, release_directory_root:
...
+
+    # If no tags exist for this asset, return True
+    if len(ordered_commits) == 0:
+        return True
+
+    # Get the latest tag
+    latest_tag = commits_tags[ordered_commits[-1]].name
+    latest_version = int(latest_tag.split("/")[2])
+
+    return latest_version + 1 == int(asset_config.version)
```
elesel, line 99: What do you think about this type of approach? I just figured out that we're grabbing commit timestamps this way, but not tag timestamps, which will follow by some amount of time.

```suggestion
    tags = [t for t in repo.tags if tag.name.startswith(f"{asset_config.partial_name}/")]
    if not tags:
       # No releases
        return True

    # Get the latest tag
    ordered_tags = sorted(tags, key=lambda t: t.tag.tagged_date)
    latest_tag = ordered_tags[-1].name
    _, _, latest_version = assets.AssetConfig.parse_full_name(latest_tag)

    return int(latest_version) + 1 == int(asset_config.version)
```

Note that the last line just compares the asset version against the latest version. Perhaps this function should just return the last version (that should be released) that you can then check against MCR?

scripts/azureml-assets/azureml/assets/update_assets.py
```diff
@@ -63,6 +65,34 @@ def release_tag_exists(asset_config: assets.AssetConfig, release_directory_root:
...
+        return True
+
+    repo = Repo(release_directory_root)
+    tags = [t for t in repo.tags if tag.name.startswith(f"{asset_config.partial_name}/")]
+    if not tags:
+       # No releases
+        return True
+
+    # Get the latest tag
+    ordered_tags = sorted(tags, key=lambda t: t.tag.tagged_date)
```
lykelly19, line 89: For just the TagReference (t) environment/promptflow-runtime/1, the TagObject (t.tag) is None, which causes an exception to be thrown when trying to access t.tag.tagged_date

scripts/azureml-assets/azureml/assets/update_assets.py
lykelly19: Does checking against MCR work similarly to _get_latest_image_suffix?

scripts/azureml-assets/azureml/assets/update_assets.py
elesel: Thanks for testing this! I think that tag must be lightweight instead of annotated ([reference](https://git-scm.com/book/en/v2/Git-Basics-Tagging#:~:text=list%20is%20mandatory.-,Creating%20Tags,-Git%20supports%20two)). Rather than making this super complicated, maybe we can go with the lowest common denominator (commit date) after all, assuming that nobody is going to tag something out of sequence?

```suggestion
    ordered_tags = sorted(tags, key=lambda t: t.commit.authored_datetime)
```

If we were to try using the tag timestamp first and then falling back to the commit date as needed, we'd have to do a lot of date conversions that wouldn't be fun, since the former is stored as seconds since the epoch and the latter as a datetime object. AND let's say we were to change that tag to an annotated one...it would have a newer timestamp than all of the other tags in the repo which would throw the sort off.

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -30,10 +30,15 @@ def copy_asset(asset_config: assets.AssetConfig,
     if release_directory_root is not None and assets.release_tag_exists(asset_config, release_directory_root):
         # Skip a released version
         return None
+    
+    if not assets.check_new_or_preview_release(asset_config) or not assets.previous_release_tag_exists(asset_config, release_directory_root):
+        # Skip if previous version was not released
+        return None
```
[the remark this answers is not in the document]
elesel: I think we'll want to log an error and ensure the script eventually exits with `exit(1)`. We'll also want to add the flag (argparse) that enables this "checking for previous release" behavior.

scripts/azureml-assets/azureml/assets/update_assets.py
elesel: I think it will be more like [_get_latest_tag_or_digest](https://github.com/Azure/azureml-assets/blob/702a9a7d96c559bc8854179b7aadff7f49a57e6c/scripts/azureml-assets/azureml/assets/environment/pin_image_versions.py#L57-L62), in that we want to make a manifest call using a specific tag, to see whether the tag exists.

scripts/azureml-assets/azureml/assets/update_assets.py
lykelly19: Got it, thanks for the explanation!

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +35,31 @@ def copy_asset(asset_config: assets.AssetConfig,
         # Skip a released version
         return None
 
+    if check_previous_release:
+        # Skip if previous version was not released
+        tag = assets.validate_new_release(asset_config, release_directory_root)
+
+        if tag is not None:
```
lykelly19, line 42: Should this code be left here or should it be a function inside of pin_image_versions? Secondly, should a retry also be done?

scripts/azureml-assets/azureml/assets/update_assets.py
```diff
@@ -63,6 +65,35 @@ def release_tag_exists(asset_config: assets.AssetConfig, release_directory_root:
     return asset_config.full_name in repo.tags
 
 
+
+def validate_new_release(asset_config: assets.AssetConfig, release_directory_root: Path) -> str:
```
elesel, line 69: Please rename to something that indicates what the function is doing now. Something like `get_latest_release_tag_version`?

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +35,31 @@ def copy_asset(asset_config: assets.AssetConfig,
         # Skip a released version
         return None
 
+    if check_previous_release:
+        # Skip if previous version was not released
+        tag = assets.validate_new_release(asset_config, release_directory_root)
```
elesel, line 40: Maybe `previous_release_version` instead of `tag`?

scripts/azureml-assets/azureml/assets/copy_assets.py
elesel: I think it makes sense to reuse the manifest retrieval code in `pin_image_versions`, if you don't think it's a bad idea to take a dependency on it. Then you'd get the retry logic, etc. And then here you'd just decide what to do the response (404 or otherwise).

scripts/azureml-assets/azureml/assets/copy_assets.py
elesel: I don't see what's actually commented out...?

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +35,31 @@ def copy_asset(asset_config: assets.AssetConfig,
...
+            encoded_tag = urllib.parse.quote(previous_release_version, safe="")
+
+            request = Request(f"https://{hostname}/v2/{repo}/manifests/{encoded_tag}",
+                                method="HEAD",
+                                headers={'Accept': "application/vnd.docker.distribution.manifest.v2+json"})
+
+            try:
+                response = assets._urlopen_with_retries(request)
+            except Exception as e:
+                logger.log_error(f"Image {image} does not exist in MCR")
```
lykelly19, line 56: Should this logging only be for when we get a 404?

elesel: Yes. If there's some other exception we'd probably want to log that too, but not say that any error means the image doesn't exist.

scripts/azureml-assets/azureml/assets/environment/pin_image_versions.py
```diff
@@ -36,6 +36,22 @@ def _urlopen_with_retries(request: Union[Request, str]) -> HTTPResponse:
...
+        tag (str): Tag for the image.
+        hostname (str): Hostname of the container registry.
+        repo (str): Repository of the image.
+    """
+    encoded_tag = urllib.parse.quote(tag, safe="")
+    request = Request(f"https://{hostname}/v2/{repo}/manifests/{encoded_tag}",
+                      method="HEAD",
+                      headers={'Accept': "application/vnd.docker.distribution.manifest.v2+json"})
+
+    return request
```
elesel, line 52: How about just calling _urlopen_with_retries from here? I don't think there's a need to ever create a request without also executing it.

scripts/azureml-assets/azureml/assets/__init__.py
```diff
@@ -24,8 +24,13 @@
 from .update_assets import (
     pin_env_files,
     release_tag_exists,
+    get_latest_release_tag_version,
     update_asset,
     update_assets,
 )
+from .environment.pin_image_versions import (
+    _urlopen_with_retries,
+    _get_manifest
```
elesel, line 33: Remove leading underscore from this function since it's no longer private.

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +34,33 @@ def copy_asset(asset_config: assets.AssetConfig,
...
+            request = assets._get_manifest(previous_release_version, hostname, repo)
+
+            try:
+                response = assets._urlopen_with_retries(request)
+            except HTTPError as e:
+                if e.code == 404:
+                    logger.log_error(f"Image {image} does not exist in MCR")
+                else:
+                    logger.log_error(f"Unexpected error when looking for image {image} in MCR")
+                raise Exception(f"Failed to retrieve manifest for {repo}:{previous_release_version}: {e}")
```
elesel, line 56: Indent

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +34,33 @@ def copy_asset(asset_config: assets.AssetConfig,
...
+            # Check against MCR by making a manifest call to see if tag exists
+            (hostname, repo) = image.split("/", 1)
+
+            request = assets._get_manifest(previous_release_version, hostname, repo)
+
+            try:
+                response = assets._urlopen_with_retries(request)
+            except HTTPError as e:
+                if e.code == 404:
+                    logger.log_error(f"Image {image} does not exist in MCR")
```
elesel, line 53: Maybe we should increment a counter here and then fail after all of the unreleased environments have been listed?
```suggestion
                    logger.log_error(f"Image {image} not found in MCR. Please release {asset_config.name} version {previous_release_version} before continuing.")
```

scripts/azureml-assets/azureml/assets/copy_assets.py
lykelly19: Should we indent this Exception and then add an exit(1) after the logging error for a 404?

elesel: Yes, I think that's a great approach.

scripts/azureml-assets/azureml/assets/update_assets.py
```diff
@@ -63,6 +65,34 @@ def release_tag_exists(asset_config: assets.AssetConfig, release_directory_root:
...
+
+    Args:
+        asset_config (assets.AssetConfig): Asset config
+        release_directory_root (Path): Release branch location
+
+    Returns:
+        str: Latest version found, or None if asset is not an environment or no tags not found
+    """
+    if asset_config.type != AssetType.ENVIRONMENT:
+        return None
```
elesel, line 79: What would you think of moving this check to copy_assets.py? Something like `if check_previous_release and asset_config.type == AssetType.ENVIRONMENT:`

We'll have other asset types that will need similar handling (although not checking via MCR), so keeping this function generic gives us some more options.

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +35,33 @@ def copy_asset(asset_config: assets.AssetConfig,
...
+
+            # Check against MCR by making a manifest call to see if tag exists
+            (hostname, repo) = image.split("/", 1)
+
+            try:
+                response = assets.get_manifest(previous_release_version, hostname, repo)
+            except HTTPError as e:
+                if e.code == 404:
+                    logger.log_error(f"Image {image} not found in MCR. Please release {asset_config.name} version \
+                                      {previous_release_version} before continuing.")
```
elesel, line 53: This approach avoids adding a bunch of white space in the middle of the string.
```suggestion
                    logger.log_error(f"Image {image} not found in MCR. Please release {asset_config.name} version "
                                     f"{previous_release_version} before continuing.")

```

scripts/azureml-assets/azureml/assets/copy_assets.py
```diff
@@ -31,9 +35,33 @@ def copy_asset(asset_config: assets.AssetConfig,
...
+        previous_release_version = assets.get_latest_release_tag_version(asset_config, release_directory_root)
+
+        if previous_release_version is not None:
+            image = asset_config.extra_config_as_object().get_full_image_name()
+
+            # Check against MCR by making a manifest call to see if tag exists
+            (hostname, repo) = image.split("/", 1)
+
+            try:
+                response = assets.get_manifest(previous_release_version, hostname, repo)
```
elesel, line 49: We don't care about the result, just that we didn't get an error.
```suggestion
                _ = assets.get_manifest(previous_release_version, hostname, repo)
```

elesel (approved):

PR #752, opened 2023-06-14
[Model Monitoring] - Fix DQ & Data Drift thresholds

mastloui-msft:
- Add a fix for data quality

### DIFF
--- a/model_monitoring/components/src/data_drift_compute_metrics/io_utils.py
+++ b/model_monitoring/components/src/data_drift_compute_metrics/io_utils.py
@@ -20,7 +20,6 @@ def _get_output_spark_df_schema():
             StructField("metric_value", FloatType(), True),
             StructField("data_type", StringType(), True),
             StructField("metric_name", StringType(), True),
-            StructField("threshold_value", FloatType(), True),
         ]
     )
     return schema
--- a/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
+++ b/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
@@ -278,12 +278,17 @@ def compute_dtype_violation_count_modify_dataset(
 
     # Loop through each column in the original DataFrame
     for column in df.columns:
-        dtype_baseline = (
-            data_stats_table_mod.select(["featureName", "dataType"])
-            .filter(data_stats_table_mod.featureName == column)
-            .select("dataType")
-            .collect()[0][0]
-        )
+
+        dtype_baseline = data_stats_table_mod.select(["featureName", "dataType"]) \
+            .filter(data_stats_table_mod.featureName == column) \
+            .select("dataType")\
+            .collect()
+
+        if(len(dtype_baseline) == 0):
+            print(f"Feature '{column}' is not present in data statistics. Skipping data type violation count for '{column}'.")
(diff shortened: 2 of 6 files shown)


model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
```diff
@@ -278,12 +278,17 @@ def compute_dtype_violation_count_modify_dataset(
...
-            .select("dataType")
-            .collect()[0][0]
-        )
+
+        dtype_baseline = data_stats_table_mod.select(["featureName", "dataType"]) \
+            .filter(data_stats_table_mod.featureName == column) \
+            .select("dataType")\
+            .collect()
+
+        if(len(dtype_baseline) == 0):
```
elesel, line 287: Or:
```suggestion
        if not dtype_baseline:
```

janhavi13 (dismissed):

elesel (approved):

janhavi13 (dismissed):

elesel (approved):

model_monitoring/components/src/data_drift_compute_metrics/io_utils.py
```diff
@@ -20,7 +20,6 @@ def _get_output_spark_df_schema():
             StructField("metric_value", FloatType(), True),
             StructField("data_type", StringType(), True),
             StructField("metric_name", StringType(), True),
-            StructField("threshold_value", FloatType(), True),
```
sunjolia, line 23: Is this removing the restriction for threshold_value to be float since in the component code you handle casting nan to float? Or why don't we need this anymore?

model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
```diff
@@ -278,12 +278,18 @@ def compute_dtype_violation_count_modify_dataset(
...
+            .filter(data_stats_table_mod.featureName == column) \
+            .select("dataType")\
+            .collect()
+
+        if not dtype_baseline:
+            print(f"Feature '{column}' is not present in data statistics. " +
+                  f"Skipping data type violation count for '{column}'.")
+            continue
+
+        dtype_baseline = dtype_baseline[0][0]
```
sunjolia, line 292: NIT: Can you add a comment explaining what this is accessing?

sunjolia (approved):

model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
mastloui-msft: I'll add as a follow up!

model_monitoring/components/src/data_drift_compute_metrics/io_utils.py
mastloui-msft: We're adding the threshold value as a post step after we calculate the metrics. In this case, it assumed it was being given *whilst* calculating the metric which broke some of the algos.

mastloui-msft: We're changing it to be strictly as a post-metric computation step now :)

janhavi13 (approved):

PR #764, opened 2023-06-20
add promptflow data for faiss index sample

augaria:


promptflow/data/faiss-index-lookup/faiss_index_sample/index.faiss
elesel, line 1: Are these binary files available anywhere else? I'm asking largely because we want to avoid adding large (subjective, I know) binary files to this repo, if possible.

augaria: hi, thanks for the feedback. however, we haven't found other suitable place to store these files and make them accessible to all Azure ML workspaces. These files are about a small sample case (< 1mb) used for a sample flow from PromptFlow, like the other cases of ask-Wikipedia, web-classification, etc..

elesel (approved):

PR #765, opened 2023-06-20
Disable automatic registration of model while importing

skasturi:
User already has access to the imported model as an anonymous asset & can register through the UI or CLI. 

The issue with register_model is due to the user having to do additional steps for registering the model by using MSI, which is not straightforward for customers.


HrishikeshGeedMS (approved):

skasturi:
Closing as we are thinking the approach.

PR #777, opened 2023-06-23
Make AssetConfig hashable, add DeployConfig.should_create 

elesel:


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -798,6 +798,10 @@ def __lt__(self, other) -> bool:
         # Compare versions using packaging's version object
         return version.parse(self.version) < version.parse(other.version)
 
+    def __hash__(self) -> int:
+        """Hash an AssetConfig object."""
+        return hash((self.type.value, self.name, self.version))
```
jitupad, line 803: why two set of parentheses ther?

jitupad (commented):
parentheses comment

MattSable (approved):

scripts/azureml-assets/azureml/assets/config.py
elesel: To create a tuple, because `hash()` accepts an object.

jitupad (approved):

PR #780, opened 2023-06-26
Update Components with latest Environment

vikasagrawal-ms:
Update Components with latest Environment

### DIFF
--- a/training/model_management/components/import_model/spec.yaml
+++ b/training/model_management/components/import_model/spec.yaml
@@ -160,25 +160,25 @@ jobs:
     outputs:
       mlflow_model_folder: ${{parent.outputs.mlflow_model_folder}}
 
-  # register_model:
-  #   component: azureml:register_model:0.0.3
-  #   compute: ${{parent.inputs.compute}}
-  #   resources:
-  #     instance_type: '${{parent.inputs.instance_type}}'
-  #   identity:
-  #     type: user_identity
-  #   inputs:
-  #     model_name: ${{parent.inputs.custom_model_name}}
-  #     model_version: ${{parent.inputs.model_version}}
-  #     model_description: ${{parent.inputs.model_description}}
-  #     registry_name: ${{parent.inputs.registry_name}}
-  #     model_metadata: ${{parent.inputs.model_metadata}}
-  #     model_type: mlflow_model
-  #     model_download_metadata: ${{parent.jobs.download_model.outputs.model_download_metadata}}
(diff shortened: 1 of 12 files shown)


training/model_management/components/import_model/spec.yaml
```diff
@@ -160,25 +160,25 @@ jobs:
...
-      model_description: ${{parent.inputs.model_description}}
-      registry_name: ${{parent.inputs.registry_name}}
-      model_metadata: ${{parent.inputs.model_metadata}}
-      model_type: mlflow_model
-      model_download_metadata: ${{parent.jobs.download_model.outputs.model_download_metadata}}
-      model_path: ${{parent.jobs.mlflow_model_local_validation.outputs.mlflow_model_folder}}
-      model_import_job_path: ${{parent.jobs.convert_model_to_mlflow.outputs.model_import_job_path}}
-    outputs:
-      registration_details: ${{parent.outputs.model_registration_details}}
+  # register_model:
```
novaturient95, line 163: Could we please share context on removing register component? Is it because customer is not aware of how to set MSI on compute specifically when submitting using designer?

novaturient95: nit: can remove the commented lines

training/model_management/src/run_model_download.py
```diff
@@ -29,18 +29,17 @@ def _get_parser():
...
 
-    for arg, value in args.__dict__.items():
-        tc_log(f"{arg} => {value}")
-
     if not ModelSource.has_value(model_source):
         tc_log(f"Unsupported model source {model_source}")
         raise Exception(f"Unsupported model source {model_source}")
 
+    tc_log(f"Model source: {model_source}")
+    tc_log(f"Model id: {model_id}")
```
skasturi, line 37: Could we please verify this before logging?

skasturi: ... using HFAPI. Hrishikesh has code to verify this.

training/model_management/components/import_model/spec.yaml
novaturient95: there is no MSI issue per se. Its an `UserError`, and we need better logging to help custom prevent it in subsequent trials, as shared over chat.

vikasagrawal-ms: Reverted as discussed

novaturient95 (approved):

novaturient95 (approved):

PR #783, opened 2023-06-26
update cryptography version to 41.0.0

bobogogo1990:
update cryptography version to 41.0.0

### DIFF
--- a/pipelines/environments/mldesigner-minimal/context/conda_dependencies.yaml
+++ b/pipelines/environments/mldesigner-minimal/context/conda_dependencies.yaml
@@ -6,5 +6,5 @@ dependencies:
 - pip=21.3.1
 - pip:
   - mldesigner=={{latest-pypi-version:pre}}
-  - cryptography==39.0.1
+  - cryptography==41.0.0
   - requests==2.31.0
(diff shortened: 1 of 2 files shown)


zhengfeiwang (approved):

pipelines/environments/mldesigner-minimal/context/conda_dependencies.yaml
```diff
@@ -6,5 +6,5 @@ dependencies:
 - pip=21.3.1
 - pip:
   - mldesigner=={{latest-pypi-version:pre}}
-  - cryptography==39.0.1
+  - cryptography==41.0.0
```
vizhur, line 9: you shouldn't specify it here, have you checked mldesigner dependencies if it brings older version of cryptography? 
not a blocking comment, you can follow up separately

vizhur (approved):

pipelines/environments/mldesigner-minimal/context/conda_dependencies.yaml
zhengfeiwang: thanks for pointing this out! in fact, `mldesigner` does not depend on `cryptography` at all, it should be mis-introduced by the `requests`

@bobogogo1990 let's remove the pin lines for `cryptography` and `requests` in these two files, the channel `conda-forge` can help us always get the latest version - which is also pointed out by Vitaliy before

zhengfeiwang: btw, @vizhur could you please help add label to this PR, so that the failing checks can run normally?

zhengfeiwang (changes requested):
remove the pin lines for `cryptography` and `requests`.

zhengfeiwang (approved):

PR #785, opened 2023-06-26
Fix mlflow local validation

novaturient95:


common/src/mlflow_model_local_validation/run_model_sanity_validations.py
```diff
@@ -29,6 +30,75 @@
...
+        except Exception as e:
+            print(f"Exception while initializing app insights: {e}")
+
+
+def tc_log(message):
+    """Log message to app insights."""
+    global tc
+    try:
+        print(message)
+        tc.track_event(name="FM_import_pipeline_debug_logs", properties={"message": message})
```
skasturi, line 51: What's the reason to not use traces?

skasturi (dismissed):

common/src/mlflow_model_local_validation/run_model_sanity_validations.py
novaturient95: Yes @skasturi. Discussed with Vikas, to remodel logging structure for all the components.

common/components/mlflow_model_local_validation/spec.yaml
```diff
@@ -1,17 +1,17 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 
 name: mlflow_model_local_validation
-version: 0.0.3
+version: 0.0.1
```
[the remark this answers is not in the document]
vikasagrawal-ms: got it this is first version

vikasagrawal-ms (approved):

common/components/mlflow_model_local_validation/spec.yaml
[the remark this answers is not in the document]
novaturient95: yes @vikasagrawal-ms. This would be first release of import model. Updated version in `import_model`

common/src/mlflow_model_local_validation/run_model_sanity_validations.py
```diff
@@ -29,6 +30,72 @@
...
+        tc.flush()
```
skasturi, line 52: Please remove calls to `flush`.

skasturi (approved):

training/model_management/src/azureml/model/mgmt/processors/transformers/whisper/conda.yaml
```diff
@@ -3,17 +3,32 @@ channels:
...
 - conda-forge
 dependencies:
-- pip<=21.2.4
-- python=3.8.13
+- python=3.8.16
+- pip<=23.0.1
 - ffmpeg=4.2.2
 - pip:
-  - mlflow<3,>=2.2
+  - mlflow==2.3.1
```
skasturi, line 10: 2.3.2?

common/src/mlflow_model_local_validation/run_model_sanity_validations.py
novaturient95: @skasturi without flush, logs are not pushed. Especially when using track event or track exception 
cc: @vikasagrawal-ms

training/model_management/src/azureml/model/mgmt/processors/transformers/whisper/conda.yaml
novaturient95: Model emitted right now contains 2.3.1 which has successful validations. Below is for t5:

image

common/components/mlflow_model_local_validation/spec.yaml
```diff
@@ -20,6 +20,9 @@ command: |
   cp ${{inputs.model_path}}/conda.yaml ./
   conda env create -p $ENV_PREFIX -f conda.yaml -q
   echo "conda env successfully created at $ENV_PREFIX"
+  echo "Install appinsights additionally, for logging"
```
vikasagrawal-ms, line 23: python-sdk-v2 has appinsights . Is it still needed


name: python-sdk-v2
channels:
  - conda-forge
dependencies:
  - python=3.8
  - pip=21.2.4
  - pip:
    - azure-core=={{latest-pypi-version}}
    - azure-ai-ml=={{latest-pypi-version}}
    - azure-identity=={{latest-pypi-version}}
    - azureml-core=={{latest-pypi-version}}
    - applicationinsights=={{latest-pypi-version}}

vikasagrawal-ms (approved):
Check on the availability of app insights from python-sdk-v2 env

common/src/mlflow_model_local_validation/run_model_sanity_validations.py
skasturi: When the process exits, IIRC the queue is automatically flushed. Doing a flush on every event adds overhead IMHO.

skasturi (approved):

common/components/mlflow_model_local_validation/spec.yaml
```diff
@@ -20,6 +20,9 @@ command: |
   cp ${{inputs.model_path}}/conda.yaml ./
   conda env create -p $ENV_PREFIX -f conda.yaml -q
```
elesel, line 21: This is likely not compatible with air-gapped regions (no access to conda channels). Please include any dependencies into the environment.

novaturient95: @elesel, for this component, conda env has to created and validated. Could you please share how we can do that? Could we skip publishing import components to air-gap region if it needs to create an env beforehand?

novaturient95: Also looks like this problem will be with evaluation component too, as they also create conda env on fly for the model and then do evaluation

training/model_management/src/azureml/model/mgmt/processors/transformers/whisper/conda.yaml
elesel, line 1: Similar question here: Can these dependencies be added to an environment and then that environment be used at runtime?

novaturient95: Whisper conda.yaml is used for dumping the model as mlflow. We load this conda.yaml as dict and pass that to mlflow.hftransformers.save_model to make sure, conda dependency created for whisper model is correct. This is not required most other models as that is captured from the env in which mlflow conversion component is running

elesel: Gotcha. Thanks for the explanation.

elesel (approved):

PR #787, opened 2023-06-27
bug fixes

vikasagrawal-ms:
Added flush() for each logging
Handled complex model id with single quotes

### DIFF
--- a/training/model_management/src/azureml/model/mgmt/utils/common_utils.py
+++ b/training/model_management/src/azureml/model/mgmt/utils/common_utils.py
@@ -140,7 +140,6 @@ def tc_log(message):
         tc.flush()
     except Exception as e:
         print(f"Exception while logging to app insights: {e}")
-    
 
 
 def tc_exception(e, message):
--- a/training/model_management/src/azureml/model/mgmt/utils/common_utils.py
+++ b/training/model_management/src/azureml/model/mgmt/utils/common_utils.py
@@ -136,6 +136,7 @@ def tc_log(message):
         tc.track_event(name="FM_import_pipeline_debug_logs", properties={"message": message})
     except Exception as e:
         print(f"Exception while logging to app insights: {e}")
+    tc.flush()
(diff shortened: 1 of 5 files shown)


training/model_management/src/azureml/model/mgmt/utils/common_utils.py
```diff
@@ -145,3 +146,4 @@ def tc_exception(e, message):
         tc.track_exception(value=e.__class__, properties={"exception": message})
     except Exception as e:
         print(f"Exception while logging exception to app insights: {e}")
+    tc.flush()
```
novaturient95, line 149: have it within `try`? If tc is not properly initialized it would throw error

vikasagrawal-ms: Ack

novaturient95: @skasturi without flush its not working for `custom_events`. In next iteration we would resolve this by using traces and letting appinsight handle flushing

novaturient95 (dismissed):

novaturient95 (approved):

training/model_management/src/azureml/model/mgmt/utils/common_utils.py
```diff
@@ -137,6 +137,7 @@ def tc_log(message):
     try:
         print(message)
         tc.track_event(name="FM_import_pipeline_debug_logs", properties={"message": message})
+        tc.flush()
```
skasturi, line 140: Could you please remove call to `flush`? Process exit should take care of it.

skasturi (approved):

training/model_management/src/azureml/model/mgmt/utils/common_utils.py
vikasagrawal-ms: it is not working, that is why added again

PR #797, opened 2023-06-27
Fix parallel component processing

elesel:


vizhur (approved):

scripts/release/asset_publish.py
```diff
@@ -305,21 +294,70 @@ def validate_update_component(
...
+
+    # Handle command and parallel components
+    if 'environment' in component_dict:
+        # Command component
+        obj_with_env = component_dict
+    elif 'task' in component_dict and 'environment' in component_dict['task']:
+        # Parallel component
+        obj_with_env = component_dict['task']
+    else:
+        logger.log_error(f"Environment reference not found in {component_name}")
```
vizhur, line 345: I think we have components with no environment, that will fail on line # 349?

elesel: Hmm. The current code would fail on those too. Lemme dig a bit.

elesel: References:
- [Command component](https://learn.microsoft.com/en-us/azure/machine-learning/reference-yaml-component-command?view=azureml-api-2#yaml-syntax)
- [Parallel component](https://learn.microsoft.com/en-us/azure/machine-learning/reference-yaml-job-parallel?view=azureml-api-2#attributes-of-the-task-key)

Environment is required for both, although it can be either an asset ID or an inline specification. I don't see any inline environments specified in our repos.

PR #804, opened 2023-06-29
Insert LLM Code RAG code

gjwoods:


assets/large_language_models/autoprompt/environments/autoprompt/environment.yaml
```diff
@@ -0,0 +1,9 @@
+image:
+  name: llm-auto-prompt
```
vizhur, line 2: azureml/curated/ namespace is required. Before that please issue a PR to MCR repo to declare the image

prakharg-msft: @vizhur , could you share instructions for MCR repo

elesel: I think looking at a previous PR might help: https://github.com/microsoft/mcr/pull/2533

tot0: It's not just adding the repos and tags to MCR repo so they're syndicated it's also setting the images images up so they're being published as Curated Environments.

My understanding of Curated Environments is they're meant to Users to start with/build on for their own purposes, and yes we some AzureML features publish envs they use there.
Enforcing that every Component we publish for Users to use must have its env on MCR is a lot of overhead to migrate from private to public preview for that are quite dynamic assets (python envs relying on fast changing python packages).

elesel: Correct. Environments in azureml are backed by images in MCR. Not building environments in the prod system registry makes supporting air-gapped regions possible. Once the image is onboarded with MCR it's pretty simple to release updated versions of it.

elesel: Images for registry environments aren't being handled as "true" curated environments, though. It's simply a backing mechanism. The process to onboard is:

1. Create a PR for the mcr repo that adds the image to their manifest, using the existing [YAML file](https://github.com/microsoft/mcr/blob/main/teams/azureml/base/curated-public.yml). Don't get thrown off by the name though.
2. Create the PR for the azureml-assets repo that adds the [environment referencing the MCR image](https://github.com/Azure/azureml-assets/wiki/Environments#environmentyaml)
3. After both MCR and azureml-assets PRs merge, the environment can be released via [this process](https://msdata.visualstudio.com/Vienna/_wiki/wikis/vienna-wiki/23811/release-dev-guide?anchor=registry-process). We're also working to unify all asset types into the same build and release pipelines.
4. Follow the environment release above with a release of the component.

assets/large_language_models/rag/environments/rag/asset.yaml
```diff
@@ -0,0 +1,6 @@
+name: llm-rag
+version: 0.10.0
```
elesel, line 2: Would you please set version to "auto" for environments? This will use one-up version numbers for environment versions and image tags.
```suggestion
version: auto
```

gjwoods: Awesome - thank you for the suggestion.

elesel (approved):

PR #816, opened 2023-06-30
filter sample data which are not safe to crawl

ishinzhang:


elesel (approved):

elesel:
If you have already deployed the web-classification data asset, you'll need to increment its version in GitHub and the release the updated version.

PR #818, opened 2023-06-30
fix parameter issue with tc_exception

novaturient95:


assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -171,7 +171,8 @@ def _save(
             config = str(tmp_config_dir)
             tokenizer = str(tmp_tokenizer_dir)
 
-        self._signatures = self._signatures or self.get_model_signature()
+        # set experimental flag
+        self._hf_conf["exp"] = True
```
skasturi, line 175: Does the default need to be True? Would the current version of the converter fail if we pass this new key?

novaturient95: @skasturi removed. 

Please let me know if this needs to be added in below manner: 
1. modify spec to have `experimentation` parameter which by default would be set to False
2. if set, add 'exp' in hf_conf and mark it True

vizhur (dismissed):

skasturi (approved):

vikasagrawal-ms (approved):

elesel (approved):

PR #827, opened 2023-07-04
Error to mitigate credentials issue

HrishikeshGeedMS:
This PR contains the raising of appropriate error msg, which will help customer to resolve the credentials issue faced by them while using registration/ deployment component through UI/studio

### DIFF
--- a/assets/common/src/model_deployment/deploy.py
+++ b/assets/common/src/model_deployment/deploy.py
@@ -161,21 +161,38 @@ def parse_args():
     return args
 
 
-def get_ml_client():
-    """Return ML client."""
-    credential = AzureMLOnBehalfOfCredential()
+def tc_exception(e, message):
+    """Log exception to app insights."""
+    global tc
     try:
+        tc.track_exception(value=e.__class__, properties={"exception": message})
+        tc.flush()
+    except Exception as e:
+        print(f"Exception while logging exception to app insights: {e}")
+
+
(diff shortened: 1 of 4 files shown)


assets/common/src/model_deployment/deploy.py
```diff
@@ -162,20 +162,33 @@ def parse_args():
...
-        # Fall back to ManagedIdentityCredential in case AzureMLOnBehalfOfCredential not work
-        print(f"Failed to get obo credentials - {ex}")
-        msi_client_id = os.environ.get("DEFAULT_IDENTITY_CLIENT_ID")
-        credential = ManagedIdentityCredential(client_id=msi_client_id)
-    try:
-        credential.get_token("https://management.azure.com/.default")
-    except Exception as ex:
-        raise (f"Failed to get credentials : {ex}")
+        # Fall back to ManagedIdentityCredential in case AzureMLOnBehalfOfCredential does not work
+        print(f"Failed to get OBO credentials - {ex}")
```
novaturient95, line 174: Could we also add details on how to run component using OBO?

novaturient95: Also if it make sense, could we create a notebook if not already existing to create MSI and attach to compute with sufficient enough permissions to perform registration and deployment in ws? Same for OBO

and then link them in the Exception we throw. 

cc: @skasturi

PR #830, opened 2023-07-05
Added oai_completions_finetune component

ganik:
This is for Public Preview of Azure OpenAI for 3p customers.

### DIFF
--- a/assets/training/oai_completions_finetune/spec.yaml
+++ b/assets/training/oai_completions_finetune/spec.yaml
@@ -1,0 +1,105 @@
+$schema: http://azureml/sdk-2-0/PipelineComponent.json
+display_name: OpenAI Completions Finetune Pipeline
+name: openai_completions_finetune
+version: 0.1.0
+type: pipeline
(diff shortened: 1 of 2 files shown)


assets/training/oai_completions_finetune/spec.yaml
```diff
@@ -0,0 +1,105 @@
...
+outputs:
+  output_model:
+    type: uri_folder
+    mode: mount
+    description: Dataset with the output model weights (LoRA weights)
+
+jobs:
+  fine_tune:
+    type: command
+    component: azureml://registries/azure-openai-preview/components/openai_completions_finetune/versions/0.3.0
```
vizhur, line 86: should this be in azureml registry?

ganik: It cant. This component requires IPP (Intellectual Property Protection) registry, azureml is not the one. Currently registry can only be either one IPP or non-IPP. There is a work going to make registry support both but its not there yet.

vizhur: do you want to publish the asset in IPP registry then?

elesel (approved):

PR #855, opened 2023-07-07
updating vision components environment

rjaincc:
updating vision components environment

### DIFF
--- a/assets/training/vision/components/image_classification/spec.yaml
+++ b/assets/training/vision/components/image_classification/spec.yaml
@@ -188,7 +188,7 @@ outputs:
 
 code: ../src
 
-environment: azureml://registries/azureml/environments/AzureML-AutoML-DNN-Vision-GPU/versions/117
+environment: azureml:AzureML-AutoML-DNN-Vision-GPU:118
 
 command: >-
   python -m image_classification.run
--- a/assets/training/vision/components/image_classification/spec.yaml
+++ b/assets/training/vision/components/image_classification/spec.yaml
@@ -188,7 +188,7 @@ outputs:
 
 code: ../src
 
-environment: azureml:AzureML-AutoML-DNN-Vision-GPU:115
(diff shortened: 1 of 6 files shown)


novaturient95 (dismissed):

assets/training/vision/components/image_classification/spec.yaml
```diff
@@ -188,7 +188,7 @@ outputs:
 
 code: ../src
 
-environment: azureml:AzureML-AutoML-DNN-Vision-GPU:115
+environment: azureml://registries/azureml/environments/AzureML-AutoML-DNN-Vision-GPU/versions/117
```
PhaniShekhar, line 191: Why do we need to point to registry environment ? 
Wouldn't the CE from workspace env be picked up with the previous approach ?
If not, how did the runtime components already published by Gaurav/Daniel/Shubham work ?

PhaniShekhar (changes requested):

assets/training/vision/components/image_classification/spec.yaml
```diff
@@ -188,7 +188,7 @@ outputs:
 
 code: ../src
 
-environment: azureml:AzureML-AutoML-DNN-Vision-GPU:115
```
PhaniShekhar, line 191: Version 118 is being rolled out now, that has 1.52.0.post1. Could you please use that instead

vizhur (changes requested):
these environments are NOT in the registry. Please consider to migrate them

assets/training/vision/components/image_classification/spec.yaml
rjaincc: There are some recent changes in publishing pipelines, which is why these changes. 
Earlier, it used to work.

assets/training/vision/components/image_classification/spec.yaml
rjaincc: Done

rjaincc:
Closing, as changed have went in from another PR.

PR #871, opened 2023-07-11
fetch component assetID from run details

novaturient95:
Pros; Fetch component asset ID from run. replace current hardcoding of import_model version. 
Cons: We fetch current component details and not parent component details. In simple terms, download component assetId will be logged in place of import_model parent compnent


novaturient95:
Add UT for run_details and custom_dimensions

skasturi (approved):

skasturi (approved):

PR #887, opened 2023-07-14
Add Open AI Support to Validate Deployments, QA Gen, Embeddings Components

gdyre:
PR to add support for OpenAI models in the Validate Deployments Component. Checks the validity of models by calling the get model call on OpenAI. Supports combinations of AOAI and OpenAI embeddings and completion models.


nanditamohan (dismissed):

vizhur (approved):

assets/large_language_models/rag/components/src/validate_deployments.py
```diff
@@ -404,7 +403,96 @@ def main(parser_args, activity_logger: Logger):
         json.dump({"deployment_validation_success": "true"}, f)
 
     activity_logger.info(
-        "[Validate Deployments]: Success! Deployments have been validated.")
+        "[Validate Deployments]: Success! AOAI deployments have been validated.")
+
+
+def get_openai_model(model, key, activity_logger: Logger):
+    """Get model info from OpenAI"""
+    endpoint = f"https://api.openai.com/v1/models/{model}"
```
shivamMg, line 411: Calling openai APIs directly instead of using openai package because the package gets tied to AOAI credentials?

shivamMg (approved):

assets/large_language_models/rag/components/src/validate_deployments.py
gdyre: Yes, and it uses auth to see if you have access to the model this way as well

nanditamohan (approved):

PR #898, opened 2023-07-18
resolve acpt vulns

vizhur:


assets/training/finetune_acft_hf_nlp/environments/acpt/context/Dockerfile
```diff
@@ -1,8 +1,5 @@
 #PTCA image
-FROM mcr.microsoft.com/azureml/aifx/stable-ubuntu2004-cu117-py38-torch1131:{{latest-image-tag}}
-
-# ls root dir
-RUN ls -lh
+FROM mcr.microsoft.com/azureml/aifx/stable-ubuntu2004-cu117-py38-torch1131:{{latest-image-tag:biweekly\.\d{6}\.\d{1}}}
```
elesel, line 2: In case >= 10 builds occur in one day:
```suggestion
FROM mcr.microsoft.com/azureml/aifx/stable-ubuntu2004-cu117-py38-torch1131:{{latest-image-tag:biweekly\.\d{6}\.\d+}}
```

elesel (approved):

assets/training/finetune_acft_hf_nlp/environments/acpt/context/Dockerfile
vizhur: then it should be `biweekly\.\d{6}\.\d+$`  :). They have bi-weekly releases and do not expect more than 2 versions per months. (well, there was 4 for June)

elesel: However you want to roll. I already approved it. :) Note that there's an implicit ^ and $ at the beginning and end of the string anyway due to re.fullmatch.

PR #909, opened 2023-07-19
Adding Unit Tests for Numerical and Categorical Drift Metrics

perezbecker:
This pull request adds unit tests for all numerical and categorical data drift metrics currently implemented. Unit tests assert whether the drift metric values implemented in this codebase match produced by the SciPy Python package with a 10% tolerance. Currently the match is much better that that tolerance. Metric values can depend on, for example, the data binning and Laplace smoothing strategies used. Test cases cover common drift scenarios, as well as edge cases (e.g., identical distributions, distributions with no values in common), but more test cases can be added easily.

Unit tests run using `pytest` but also show human-readable output describing the test scenario and expected and measured outputs for better understanding:      

```
#######################################################
TESTING DISTANCE MEASURE: NormalizedWassersteinDistance
#######################################################
-------------------------
test: 1a
test scenario: Distance between identical distributions should be exactly 0
expected value: 0
measured value: 0
-------------------------
test: 1b
test scenario: Distance between two samples of the different sizes of the same distributions should be close to 0
expected value: 0.02335
measured value: 0.02335
-------------------------
test: 1c
test scenario: Non-zero distance between distributions with the same mean and sample size, but different std_dev
expected value: 0.2656
measured value: 0.2656
-------------------------
test: 2a
test scenario: Confirming distance of moderately drifted distribution against SciPy
expected value: 0.2005
measured value: 0.2005
-------------------------
test: 2b
test scenario: Weak distance value dependence on sample size. Distance should be close to test 2a.
expected value: 0.1805
measured value: 0.1805
-------------------------
```
When re-implementing the Spark code from SynapseML, these tests should provide all the necessary checks we need to ensure their accuracy before deployment.

This PR also contains a small bug fix for the categorical PSI metric definition uncovered when running these tests. All current metric definitions seem accurate now. 

**IMPORTANT** - I am not sure which environment you use to run these tests. I ran them on a AzureML Notebook with Serverless Spark compute attached. You might need to modify them slightly for them to run in your environment.

### DIFF
--- a/assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
+++ b/assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
@@ -3,7 +3,7 @@
 
 """
 This file contains unit tests for the Numerical Data Drift Metrics.
-It compares the drift measures of metrics implemented in Spark against their SciPy implementation 
+It compares the drift measures of metrics implemented in Spark against their SciPy implementation.
 """
 
 from data_drift_compute_metrics.numerical_data_drift_metrics import compute_numerical_data_drift_measures_tests
@@ -26,8 +26,9 @@ distance_measures = [
 
 
 test_cases = [
-    {"name": "1a",
-     "scenario": "Distance between identical distributions should be exactly 0",
+    {
+     "name": "1a",
+     "scenario": "Distance between identical distributions should be exactly 0.",
      "x_mean": 50,
      "y_mean": 50,
      "x_std_dev": 15,
@@ -35,8 +36,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 100_000,
     },
-    {"name": "1b",
-     "scenario": "Distance between two samples of the different sizes of the same distributions should be close to 0",
+    {
+     "name": "1b",
+     "scenario": "Distance between two samples of the different sizes of the same distributions should be close to 0.",
      "x_mean": 50,
      "y_mean": 50,
      "x_std_dev": 15,
@@ -44,8 +46,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 10_000,
     },
-    {"name": "1c",
-     "scenario": "Non-zero distance between distributions with the same mean and sample size, but different std_dev",
+    {
+     "name": "1c",
+     "scenario": "Non-zero distance between distributions with the same mean and sample size, but different std_dev.",
      "x_mean": 50,
      "y_mean": 50,
      "x_std_dev": 15,
@@ -53,8 +56,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 100_000,
     },
-    {"name": "2a",
-    "scenario": "Confirming distance of moderately drifted distribution against SciPy",
+    {
+     "name": "2a",
+     "scenario": "Confirming distance of moderately drifted distribution against SciPy.",
      "x_mean": 50,
      "y_mean": 53,
      "x_std_dev": 15,
@@ -62,8 +66,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 100_000,
     },
-    {"name": "2b",
-    "scenario": "Weak distance value dependence on sample size. Distance should be close to test 2a.",
+    {
+     "name": "2b",
+     "scenario": "Weak distance value dependence on sample size. Distance should be close to test 2a.",
      "x_mean": 50,
      "y_mean": 53,
      "x_std_dev": 15,
@@ -71,8 +76,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 10_000,
     },
-    {"name": "2c",
-    "scenario": "Decreasing std_dev in prodcution dataset. Increase in distance expected when compared to test 2a.",
+    {
+     "name": "2c",
+     "scenario": "Decreasing std_dev in prodcution dataset. Increase in distance expected when compared to test 2a.",
      "x_mean": 50,
      "y_mean": 53,
      "x_std_dev": 15,
@@ -80,8 +86,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 100_000,
     },
-    {"name": "2d",
-    "scenario": "Decreasing both sample size and std_dev in prodcution dataset. Similar distance as test 2c expected.",
+    {
+     "name": "2d",
+     "scenario": "Decreasing both sample size and std_dev in prodcution dataset. Similar values as test 2c expected.",
      "x_mean": 50,
      "y_mean": 53,
      "x_std_dev": 15,
@@ -89,8 +96,9 @@ test_cases = [
      "x_obs": 100_000,
      "y_obs": 10_000,
     },
-    {"name": "3a",
-    "scenario": "Large drift. Bound distnace measures (Jensen-Shannon) approaching maximum theoretical values.",
+    {
+     "name": "3a",
+     "scenario": "Large drift. Bound distnace measures (Jensen-Shannon) approaching maximum theoretical values.",
      "x_mean": 50,
      "y_mean": 100,
      "x_std_dev": 15,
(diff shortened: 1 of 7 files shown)


mastloui-msft (approved):

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
```diff
@@ -1,87 +1,342 @@
...
+
+        x_count = np.histogram(x_array, bins=bin_edges)[0]
+        y_count = np.histogram(y_array, bins=bin_edges)[0] 
+        
+        # Laplace smoothing (incrementing the count of each bin by 1)
+        # to avoid zero values in bins and have the SPI value be finit
+        x_count = x_count + 1
+        y_count = y_count + 1 
+
+        x_percent = x_count / len(x_array)
```
yetamsft: Do we need to increment the denominator (like `len(x_array)` too so that the percentage numbers add up to 100?

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
```diff
@@ -1,87 +1,413 @@
...
+
+        x_count = np.histogram(x_array, bins=bin_edges)[0]
+        y_count = np.histogram(y_array, bins=bin_edges)[0]
+
+        # Laplace smoothing (incrementing the count of each bin by 1)
+        # to avoid zero values in bins and have the SPI value be finit
+        x_count = x_count + 1
+        y_count = y_count + 1
+
+        x_percent = x_count / len(x_array)
```
yetamsft, line 337: Do we need to increment the denominator too so that the percentage numbers still add up to 100?

yetamsft: I notice the psi metrics are calculated in the same way in the source code, so I guess this is how the Laplace smoothing works. Is that right?

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
```diff
@@ -1,87 +1,413 @@
...
+     "x_mean": 50,
+     "y_mean": 53,
+     "x_std_dev": 15,
+     "y_std_dev": 5,
+     "x_obs": 100_000,
+     "y_obs": 10_000,
+    },
+    {
+     "name": "3a",
+     "scenario": "Large drift. Bound distnace measures (Jensen-Shannon) approaching maximum theoretical values.",
```
yetamsft, line 102: nit: typo `distance`

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
```diff
@@ -1,87 +1,413 @@
...
+     "x_mean": 50,
+     "y_mean": 100,
+     "x_std_dev": 15,
+     "y_std_dev": 15,
+     "x_obs": 100_000,
+     "y_obs": 100_000,
+    },
+    {
+     "name": "3b",
+     "scenario": "Extreme drift scenario. Bound distnace measures (JSD) should reach their maximum theoretical value.",
```
yetamsft, line 112: nit: typo `distance`

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
```diff
@@ -1,87 +1,413 @@
...
+     "x_mean": 50,
+     "y_mean": 50,
+     "x_std_dev": 15,
+     "y_std_dev": 15,
+     "x_obs": 9_000,
+     "y_obs": 900,
+    },
+    {
+     "name": "5c",
+     "scenario": "Non-zero distance between distributions with the same mean and sample size, but differnt std_dev.",
```
yetamsft, line 152: nit: typo `different`

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
```diff
@@ -1,87 +1,413 @@
...
+    {
+     "name": "6d",
+     "scenario": "Decreasing both sample size and std_dev in production dataset. Similar values as test 6c expected.",
+     "x_mean": 50,
+     "y_mean": 51,
+     "x_std_dev": 15,
+     "y_std_dev": 5,
+     "x_obs": 9_000,
+     "y_obs": 900,
+    },
```
yetamsft, line 201: One question for curiosity, what is the difference between scenario group 2(2a to 2d) vs group 6(6a to 6d)? From what I am reading, they are pretty similar.

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
perezbecker: Yes! Great Catch! I did not see the issue as it was both present in the source code as well as the test, so all tests passed! 
Here is a snippet which reproduces the issue:
```python
x_array = [1, 2, 2, 3, 3, 3]
bin_edges = [0.5, 1.5, 2.5, 3.5, 4.5]

x_count = np.histogram(x_array, bins=bin_edges)[0]
print(f'x_count before: {x_count}')

x_count = x_count + 1
print(f'x_count after: {x_count}')

x_percent = x_count / len(x_array)
print(f'x_percent: {x_percent}')
print(f'Sum of percentages: {sum(x_percent)}')
```
```
x_count before: [1 2 3 0]
x_count after: [2 3 4 1]
x_percent: [0.33333333 0.5        0.66666667 0.16666667]
Sum of percentages: 1.6666666666666667
```

After applying fix: 
```python
x_array = [1, 2, 2, 3, 3, 3]
bin_edges = [0.5, 1.5, 2.5, 3.5, 4.5]

x_count = np.histogram(x_array, bins=bin_edges)[0]
print(f'x_count before: {x_count}')

x_count = x_count + 1
print(f'x_count after: {x_count}')

x_percent = x_count / (len(x_array) + len(x_count))
print(f'x_percent: {x_percent}')
print(f'Sum of percentages: {sum(x_percent)}')
```
```
x_count before: [1 2 3 0]
x_count after: [2 3 4 1]
x_percent: [0.2 0.3 0.4 0.1]
Sum of percentages: 1.0
```
I fixed the issue both in the test and source code.

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
perezbecker: This as indeed an issue. It is easy to overlook as the code for categorical data looks like this:
```python
        # Laplace smoothing (incrementing the count of each bin by 1)
        # to avoid zero values in bins and have the SPI value be finit
        x_counts = x_counts + 1
        y_counts = y_counts + 1

        x_ratios = x_counts / np.sum(x_counts)
        y_ratios = y_counts / np.sum(y_counts)
 ```
Which is very similar in structure to the erroneous expression for the numerical data, but logically different. Here we are already normalizing using the updated counts.

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
perezbecker: fixed typo

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
perezbecker: fixed typo.

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
perezbecker: Yes they are very similar. The reason why I am including both is because the current implementation of the KS test is capped at 10k samples. I wanted to include tests that covered about 100k samples, as well as some in the upper limit of what KS can handle.

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
perezbecker: fixed typo.

assets/model_monitoring/components/tests/unit/test_numerical_data_drift_metrics.py
yetamsft: Got it. Thanks for answering.

yetamsft (approved):
Looks good to me! Thank you again for safeguarding the metric calculation.

mastloui-msft (approved):

PR #917, opened 2023-07-21
Update conda_dependencies.yaml

kicha0:
Add torch and tensorflow dependencies.


romanlutz:
What's motivating this? Installing all of these is likely going to cause issues later on.

kicha0:
this enables support of tensorflow and pytorch model cpu inference. I agree adding model dependencies to rai image is not a sustainable approach, so the next steps to supporting all model types would likely be done by online endpoint

vizhur (changes requested):
remove wheels installed from a blob. also remove all pip freeze stuff and any other unnecessary instructions.

as a side note, please consider to update scipy that has a vulnerability approaching due date

imatiach-msft:
closing old PRs

PR #921, opened 2023-07-24
Add a metric output builder for generation token stats and others

yetamsft:


assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
```diff
@@ -0,0 +1,52 @@
+$schema: http://azureml/sdk-2-0/SparkComponent.json
+type: spark
+
+name: model_monitor_output_signal_metrics
+display_name: Model Monitor - Output Signal Metrics
+description: Output the computed model monitor metrics to the default datastore.
```
mastloui-msft, line 6: can you remove the default datastore here?

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+inputs:
+  signal_metrics: 
+    type: mltable
+    mode: direct
+  signal_name:
+    type: string
+  signal_type: 
+    type: string
+  version:
```
mastloui-msft, line 22: what's the version in this case?

mastloui-msft: If this is the version for the output, we don't need to expose it as an input, we can simply keep it as a constant within the source file

yetamsft: yes, so the output file will always have the same version, like `1.0.0"`. Is that right? I am unsure what this version is for.

assets/model_monitoring/components/src/model_monitor_output_signal_metrics/run.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Entry script for Model Monitor Output Metrics Component."""
+
+import argparse
```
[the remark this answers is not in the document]
yetamsft: yes!

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
yetamsft: yes, removed.

assets/model_monitoring/components/src/model_monitor_output_signal_metrics/run.py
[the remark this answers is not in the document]
yetamsft: Sorted.

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
```diff
@@ -0,0 +1,52 @@
...
+description: Output the computed model monitor metrics.
+version: 0.3.2
+is_deterministic: true
+
+code: ../../src/
+entry:
+  file: ./model_monitor_output_signal_metrics/run.py
+
+inputs:
+  signal_metrics:
```
chupadhy, line 15: nit: Since the component is named 'output signal metrics' it is a little confusing that signal metrics is the input.

How about we name the component 'Model Monitor - Metrics Output Builder'? For common components used by all signals, we have not used the term 'signal' anywhere.

Inputs can be 'computed_metrics' (since the components are called compute metrics), 'signal name' and 'signal type' and output would be 'metrics_output'.

What do you think?

chupadhy: Ah maybe keep the input as 'signal_metrics' since we use that term in the prev component - avoid breaking changes :)

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
yetamsft: Updated assuming it is the output schema version (aka, `METADATA_VERSION`).

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
yetamsft: That is a good point. I renamed the component to 'Model Monitor - Metrics Outputter' and left the name of object builder class unchanged. Also left the `signal_metrics` input name unchanged so that we don't confuse ourselves. If needed, we can rename it post-cutover.

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
mastloui-msft: That version is for a manual version bump when we want to do UX migration - we'll probably be using it when we migrate to outputv2 as a whole (aka from 1 to 2 or something)

mastloui-msft: We'll be bumping very very rarely (hopefully!!)

assets/model_monitoring/components/src/shared_utilities/constants.py
```diff
@@ -47,6 +47,19 @@
 UPPER_BOUND_PARAM = 'upperBound'
 VERSION_PARAM = 'version'
 
+# Keywords in metrics schema
+AGGREGATE = 'aggregate'
+GROUP = 'group'
+GROUP_PIVOT = 'group_pivot'
```
mastloui-msft, line 53: this changed to group_dimension!

assets/model_monitoring/components/src/shared_utilities/constants.py
```diff
@@ -47,6 +47,19 @@
...
+METRIC_NAME = 'metric_name'
```
mastloui-msft, line 54: Can we prepend/append some term to distinguish between random constants and column keys for signal metrics?
maybe SIGNAL_METRIC_GROUP_COLUMN or GROUP_COLUMN? etc..

assets/model_monitoring/components/model_monitor/model_monitor_output_signal_metrics/spec.yaml
yetamsft: Sounds good! Thanks.

assets/model_monitoring/components/src/shared_utilities/constants.py
yetamsft: Renamed.

assets/model_monitoring/components/src/shared_utilities/constants.py
yetamsft: Prepend column keys with `SIGNAL_METRICS_`.

mastloui-msft (approved):

mastloui-msft (approved):

chupadhy (approved):

PR #923, opened 2023-07-24
Fix key=None logic for completion model/deployment in PF Creation

nanditamohan:


assets/large_language_models/components_pipelines/data_ingestion_dataset_to_acs/spec.yaml
gdyre, line 1: Do you need to update version numbers? Or does staging handle the same version numbers differently?

nanditamohan: Thanks for the call out! I needed to update the top-level version - done :)

gdyre (approved):

assets/large_language_models/rag/components/src/flow_creation.py
```diff
@@ -144,6 +144,9 @@ def main(args, ws, current_run, activity_logger: Logger):
     completion_model_name = completion_config.get("model_name", "gpt-35-turbo")
     completion_deployment_name = completion_config.get(
         "deployment_name", "gpt-35-turbo")
+    # Set default if key exsits but is set to None (as it is for basic pipelines)
+    completion_model_name = "gpt-35-turbo" if completion_model_name is None else completion_model_name
```
chaoqian, line 148: when will this be none? can u share an example completion_config?

nanditamohan: Example: {"type": "azure_open_ai", "model_name": "gpt-35-invalid", "deployment_name":null, "temperature": 0, "max_tokens": 2000}

This will only send like this once ES change to stop deployment inferring for Basic pipelines is rolled out - this is blocking users from submitting pipeline today if they do not have that completion model in AOAI ws: https://msdata.visualstudio.com/Vienna/_git/vienna/pullrequest/1114473?_a=files

nanditamohan: Discussed with Prakhar that user can pick the correct model in PF itself, rather than have lack of completion model block RAG pipelines

PR #927, opened 2023-07-25
aml-benchmark folder init with dataset sampler component

iamrk04:
This PR introduces a new folder `aml-benchmark` to hold all the components along with tests that the aml-benchmarking team is developing.


assets/aml-benchmark/requirements.txt
```diff
@@ -0,0 +1,10 @@
+pytest
```
novaturient95, line 1: nit: do we need requirements.txt as well?

assets/aml-benchmark/dev_conda_env.yaml
```diff
@@ -0,0 +1,19 @@
+channels:
```
novaturient95, line 1: nit: add under tests?

iamrk04: will let it be as it contains both src and test requirements. It is not only helpful for running tests, but also for running the scripts locally.

assets/aml-benchmark/requirements.txt
iamrk04: will remove this as this is no longer required.

arun-rajora (dismissed):

novaturient95 (approved):

vizhur (changes requested):

.github/workflows/aml-benchmark-test.yaml
elesel, line 1: I don't think this workflow needs to exist because asset-related tests are already executed by the [assets-test workflow](https://github.com/Azure/azureml-assets/blob/main/.github/workflows/assets-test.yaml). See the `test` section of the [asset.yaml file](https://github.com/Azure/azureml-assets/wiki/Assets#assetyaml) and example tests [here](https://github.com/Azure/azureml-assets/tree/main/assets/training/general/environments/acpt-pytorch-1.11-cuda11.3/tests), which are wired up very similarly to yours so it shouldn't be difficult to switch.

assets/aml-benchmark/components/dataset-sampler/asset.yaml
elesel, line 1: Just add a `tests` section here, and then move the test files under this directory.

elesel (changes requested):

assets/aml-benchmark/components/dataset-sampler/asset.yaml
iamrk04: Hi @elesel, thanks for the suggestion. While I see the advantage this change can bring, we want to keep all the tests together so that we can maximize reuse of code, more components are under development. And this would require this workflow. Let me know if it's ok to leave this as it is.

assets/aml-benchmark/validation_rules.json
```diff
@@ -0,0 +1,13 @@
+{
+  "pep8": {
+    "ignore": [
+      "W503",
+      "E402"
+    ],
+    "justification": [
+      "W503: W503 and W504 are mutually exclusive. W503 is more widely used, so ignoring it.",
+      "E402: All module level imports can't be before `sys.path.append(os.getcwd())`."
```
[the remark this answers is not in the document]
iamrk04: Hi @vizhur, I will get rid of W503 as this was for older version and I don't see this being flagged now. E402 will stay because this is required to support the code structure that we have. Let me know if it's ok to live with this, will update to only ignore E402 in specific files as done in this [vallidation_rules.json](https://github.com/Azure/azureml-assets/blob/main/assets/benchmark/validation_rules.json)

vizhur (changes requested):
remove pep8 exceptions

assets/aml-benchmark/components/dataset-sampler/asset.yaml
elesel: The existing test framework supports what you're describing. Please use it instead.

assets/aml-benchmark/README.md
```diff
@@ -0,0 +1,57 @@
...
+In the root of the repo, run the following in **powershell**:
+```
+python scripts/validation/copyright_validation.py -i assets/aml-benchmark/
+```
+
+## 4. Run doc style check
+In the root of the repo, run the following in **powershell**:
+```
+python scripts/validation/doc_style.py -i assets/aml-benchmark/
+```
```
elesel, line 57: These will run automatically for PRs. They shouldn't be needed here as a manual step prior to PR creation.

assets/aml-benchmark/components/src/dataset_downloader/main.py
```diff
@@ -0,0 +1,260 @@
...
+        "--dataset_name",
+        type=str,
+        default=None,
+        help="Name of the dataset to download.",
+    )
+    parser.add_argument(
+        "--configuration",
+        type=str,
+        default=None,
+        help="Configuration of the dataset to download.",
```
elesel, line 41: Might be helpful to include some additional information about what "configuration" means.

assets/aml-benchmark/components/src/dataset_downloader/main.py
```diff
@@ -0,0 +1,260 @@
...
+    )
+    parser.add_argument(
+        "--split",
+        type=str,
+        default=None,
+        help="Split of the dataset to download.",
```
elesel, line 47: Please include information about allowed values.

assets/aml-benchmark/components/src/dataset_downloader/main.py
```diff
@@ -0,0 +1,260 @@
...
+        if len(available_configs) > 1:
+            mssg = (
+                f"Multiple configurations available for dataset '{dataset_name}'. Please specify either one of "
+                f"the following: {available_configs} or 'all'."
+            )
+            raise BenchmarkValidationException._with_error(
+                AzureMLError.create(BenchmarkValidationError, error_details=mssg)
+            )
+        else:
+            return [configuration]
```
elesel, line 83: Would `available_configs` ever be empty?

assets/aml-benchmark/components/src/dataset_downloader/main.py
```diff
@@ -0,0 +1,260 @@
...
+) -> List[str]:
+    """
+    Get the list of splits to download.
+
+    :param dataset_name: Name of the dataset to download.
+    :param configuration: Configuration of the dataset to download.
+    :param split: Split of the dataset to download.
+    :return: list of splits to download.
+    """
+    if split is None:
```
elesel, line 107: Couldn't you mark the arg as required?

assets/aml-benchmark/validation_rules.json
elesel, line 1: This file shouldn't be needed because this one will be used instead: https://github.com/Azure/azureml-assets/blob/main/validation_rules.json

assets/aml-benchmark/README.md
iamrk04: The reason they have been included is to proactively detect any such failures before creating a PR, and fix them early on. Otherwise the workflow runs, fails with some issue and then people try to fix it.

assets/aml-benchmark/components/src/dataset_downloader/main.py
iamrk04: `available_configs` won't be empty. For datasets that do not have configuration listed on the HuggingFace UI, they return a random config name. We do not want to use it, configuration should be `None` in this case to download the dataset.

assets/aml-benchmark/components/src/dataset_downloader/main.py
iamrk04: done, thanks!

assets/aml-benchmark/components/src/dataset_downloader/main.py
iamrk04: there is no restriction on the allowed values as it can be any str, have added the details about what value `all` can do.

assets/aml-benchmark/components/src/dataset_downloader/main.py
iamrk04: good catch.

assets/aml-benchmark/validation_rules.json
iamrk04: removed, thanks!

elesel (approved):

PR #929, opened 2023-07-26
changed description for bert-base-cased

KonakanchiSwathi:
changed description for bert-base-cased


KonakanchiSwathi (commented):
made changes to description

PR #938, opened 2023-07-28
Fix model name error

novaturient95:


assets/common/src/register.py
```diff
@@ -141,14 +145,16 @@ def main():
...
-            )
+        raise AzureMLException._with_error(AzureMLError.create(UnSupportedModelTypeError, model_type=model_type))
 
     if not model_name:
-        raise AzureMLException._with_error(
-            AzureMLError.create(MissingModelNameError)
-        )
+        raise AzureMLException._with_error(AzureMLError.create(MissingModelNameError))
+
+    if not re.match(VALID_MODEL_NAME_PATTERN, model_name):
```
skasturi, line 153: Do we also need to update the import notebook for this?

assets/common/src/register.py
```diff
@@ -27,6 +28,9 @@
 
 SUPPORTED_MODEL_ASSET_TYPES = [AssetTypes.CUSTOM_MODEL, AssetTypes.MLFLOW_MODEL]
```
skasturi, line 29: Could we read this directly from the v2 SDK? I believe we have "Triton" also supported now.

assets/common/src/register.py
```diff
@@ -27,6 +28,9 @@
...
 MLFLOW_MODEL_FOLDER = "mlflow_model_folder"
```
skasturi, line 30: Is this the mandatory name for the folder? If yes, how are we communicating this to the customers?

assets/common/src/register.py
```diff
@@ -178,8 +184,7 @@ def main():
                 model_version = str(int(max_version) + 1)
```
skasturi, line 184: Replace with `label="latest"` like you were suggesting in the notebook PR?

novaturient95: Done

assets/common/src/register.py
novaturient95: @skasturi pulling from v2 SDK only. Triton we are not including because that has never been tested. We can test once and include in SUPPORT_MODEL_ASSET_TYPES.

assets/common/src/register.py
novaturient95: yes, for mlflow model, it has to be `mlflow_model_folder`. We are making sure to copy model files to `mlflow_model_folder` for mlflow model. MLmodel file has to exists under `mlflow_model_folder` or it will fail. For custom model we are publishing as it is

vikasagrawal-ms (approved):

HrishikeshGeedMS (approved):

assets/common/components/register_model/spec.yaml
```diff
@@ -29,12 +29,12 @@ inputs:
   model_name:
     type: string
     optional: true
-    description: Model name to use in the registration. If name already exists, the version will be auto incremented
+    description: Model name to use in the registration. If name already exists, the version will be auto incremented. Model name should satisfy pattern `^[a-zA-Z0-9-]+$`
```
BhavanaTumma, line 33: Would a user friendly explanation make more sense here? instead of adding a regex

assets/common/src/register.py
```diff
@@ -141,14 +145,16 @@ def main():
...
-        raise AzureMLException._with_error(
-            AzureMLError.create(MissingModelNameError)
-        )
+        raise AzureMLException._with_error(AzureMLError.create(MissingModelNameError))
+
+    if not re.match(VALID_MODEL_NAME_PATTERN, model_name):
+        # update model name to one supported for registration
+        logger.info(f"Updating model name to match pattern `{VALID_MODEL_NAME_PATTERN}`")
+        model_name = re.sub(NEGATIVE_MODEL_NAME_PATTERN, "-", model_name)
+        logger.info(f"Updated model_name = {model_name}")
```
BhavanaTumma, line 158: Would the customer know if they have registered the model with an incorrect naming convention and we changed it? and to what we've changed it?

assets/common/src/register.py
```diff
@@ -165,23 +171,19 @@ def main():
...
         model_version = "1"
-        models_list = []
         try:
-            models_list = ml_client.models.list(name=model_name)
-            if models_list:
-                max_version = (max(models_list, key=lambda x: int(x.version))).version
-                model_version = str(int(max_version) + 1)
-        except Exception:
+            model_version = str(int(ml_client.models.get(name=model_name, label="latest").version) + 1)
+        except Exception as e:
```
BhavanaTumma, line 180: What if the model doesn't exist? does it return 0?

novaturient95: it will go into exception and we use `1` by default

assets/common/src/register.py
novaturient95: 1. There is a log that I have added in the PR for name change.
2. registration asset id is logged for successful registration which would contain updated model name
3. registeration_details would contain updated model name

please let me know if we need to improve logging or share details by some other means

assets/common/components/register_model/spec.yaml
novaturient95: Good point. Since this was a basic regex I updated it without much explanation. But let me update

novaturient95: updated

assets/common/src/register.py
```diff
@@ -212,7 +214,7 @@ def main():
     }
     json_object = json.dumps(model_info, indent=4)
 
-    registration_file = registration_details_folder/ComponentVariables.REGISTRATION_DETAILS_JSON_FILE
+    registration_file = registration_details_folder / ComponentVariables.REGISTRATION_DETAILS_JSON_FILE
```
skasturi, line 217: (nit) Use os.path.join?

skasturi (approved):

assets/common/src/register.py
novaturient95: Since its a `Path` object where `/` is overriden for path join operation, its recommended to use it over `os.path.join`. Both are nearly same on performance aspects

https://docs.python.org/3/library/pathlib.html

PR #945, opened 2023-07-31
fail early if SKU is not large enough for download

novaturient95:


assets/training/model_management/src/azureml/model/mgmt/utils/common_utils.py
```diff
@@ -221,3 +223,68 @@ def fetch_huggingface_model_info(model_id) -> ModelInfo:
...
+
+
+def get_vm_available_space_in_kb():
+    """Return available VM size in KB.
+
+    :return: Return available size on VM in KB
+    :rtype: int
+    """
+    try:
+        cmd = "df -k | awk '/\/$/ {print $4}'"
```
skasturi, line 260: Don't have have `os` package methods for this?

novaturient95: Let me check

novaturient95: updated

assets/training/model_management/src/azureml/model/mgmt/utils/common_utils.py
```diff
@@ -221,3 +224,67 @@ def fetch_huggingface_model_info(model_id) -> ModelInfo:
...
+    :return: total LFS blob size in KB
+    :rtype: int
+    """
+    try:
+        # 1. Execute git lfs ls-files -s command to get size of all blobs recursively
+        # 2. Filter size using awk utility and translate them to KB.
+        #    Expectation is that each blob dimension would be in [KB, MB, GB]
+        # 3. Perform sum and return total size in KB
+        cmd = (
+            f"cd {str(git_dir)} && "
```
skasturi, line 279: Where are we stopping `git lfs` from cloning large files?

novaturient95: here: 

image

novaturient95: In `GITDownloader._download`

PR #948, opened 2023-07-31
Upgrade package

raoxiang1996:


assets/large_language_models/dbcopilot/environments/dbcopilot_grounding/context/Dockerfile
```diff
@@ -25,5 +25,36 @@ RUN /bin/bash -c "source ~/.bashrc"
 
 
 # Workaround: install packages with extra index url. Will be removed after the packages are published to pypi
-RUN pip install https://ragsample.blob.core.windows.net/ragdata/wheels/dbcopilot/dbcopilot-0.3.20230613.dev4-py3-none-any.whl
-RUN pip install https://ragsample.blob.core.windows.net/ragdata/wheels/dbcopilot/db_copilot_tool-0.1.3-py3-none-any.whl
+RUN pip install https://ragsample.blob.core.windows.net/ragdata/wheels/dbcopilot/dbcopilot-0.3.20230714.1-py3-none-any.whl
+RUN pip install https://ragsample.blob.core.windows.net/ragdata/wheels/dbcopilot/db_copilot_tool-0.1.7-py3-none-any.whl
```
vizhur, line 29: no installs from blob

vizhur (changes requested):

assets/large_language_models/dbcopilot/environments/dbcopilot_grounding/context/Dockerfile
xia-xiao: unfortunately there is no solution to not install this from blob. We still need to keep this.

assets/large_language_models/dbcopilot/components/src/component_base.py
```diff
@@ -42,22 +46,30 @@ def wrapper(**kwargs):
...
                 args = parser.parse_args()
                 logging.info("\n".join(f"{k}={v}" for k, v in vars(args).items()))
-                wrapper(**vars(args))
-
+                try:
+                    wrapper(**vars(args))
+                except Exception as ex:
+                    raise ex
+                finally:
+                    logging.info("finally")
```
tot0, line 71: [nit] This isn't the most useful log :P

assets/large_language_models/dbcopilot/environments/dbcopilot_grounding/context/Dockerfile
tot0: What options are available here? There's no option to publish the code to pypi as far as I'm aware, where can we put it such that it can be built into this image?

vizhur (approved):

xia-xiao (approved):

PR #970, opened 2023-08-03
Large model publishing support added

HrishikeshGeedMS:
Support added to publish models from azureblob without downloading in ev2 agent and stream directly to registry storage

### DIFF
--- a/scripts/azureml-assets/azureml/assets/model/utils.py
+++ b/scripts/azureml-assets/azureml/assets/model/utils.py
@@ -10,7 +10,13 @@ from subprocess import PIPE, run, STDOUT
 import sys
 from azureml.assets import PathType
 from azureml.assets.util import logger
+import requests
+from azure.identity import DefaultAzureCredential
+from azure.ai.ml.entities import Model
 
+CREDENTIAL = DefaultAzureCredential()
+HEADERS = {"Authorization": "Bearer " + CREDENTIAL.get_token('https://management.azure.com/.default').token}
+ASSET_STORE_URL = "https://eastus.api.azureml.ms/assetstore/v1.0"
 
 def _onerror(func, path, exc_info):
     """Error Handler for shutil rmtree."""
@@ -41,6 +47,21 @@ class ModelDownloadUtils:
         else:
             logger.print(f"Successfully executed! Output: \n{result.stdout}")
         return result.returncode
+    
+    def _get_temporary_data_reference(registry_name: str, model_name: str, model_version: str):
+        asset_id = f"azureml://registries/{registry_name}/models/{model_name}/versions/{model_version}"
+        pending_upload_body = {
+            "assetId": asset_id, 
+            "temporaryDataReferenceId": "model_weights", 
+            "temporaryDataReferenceType": "TemporaryBlobReference"
+        }
+        response = requests.post(
+            ASSET_STORE_URL + "/temporaryDataReference/createOrGet", 
+            json=pending_upload_body, 
+            headers=HEADERS
+        )
+        pending_upload_response = response.json()
+        return pending_upload_response
 
     def _download_git_model(model_uri: str, model_dir: Path) -> bool:
         """Download model files from GIT repository.
@@ -58,7 +79,7 @@ class ModelDownloadUtils:
         shutil.rmtree(git_path, onerror=_onerror)
         return True
 
-    def _download_azure_artifacts(model_uri, model_dir) -> bool:
+    def _download_azure_artifacts(model_uri: str, registry_name: str, model_obj: Model) -> bool:
         """Download model files from blobstore.
 
         :param model_url: Publicly readable blobstore URI of model files
@@ -67,19 +88,23 @@ class ModelDownloadUtils:
         :type: Path
         """
         try:
-            download_cmd = f"azcopy cp --recursive=true {model_uri} {model_dir}"
+            temporary_data_reference = ModelDownloadUtils._get_temporary_data_reference(registry_name, model_obj.name, model_obj.version)
+            sas_uri = temporary_data_reference["blobReferenceForConsumption"]["credential"]["sasUri"]
+            blob_uri = temporary_data_reference["blobReferenceForConsumption"]["blobUri"]
+            download_cmd = f"azcopy cp --recursive=true {model_uri} {sas_uri}"
             result = ModelDownloadUtils._run(download_cmd)
             # TODO: Handle error case correctly, since azcopy exits with 0 exit code, even in case of error.
             # https://github.com/Azure/azureml-assets/issues/283
             if result:
                 logger.log_error(f"Failed to download model files with URL: {model_uri}")
                 return False
+            model_obj.path = blob_uri
             return True
         except Exception as e:
             logger.log_error(e)
             return False
 
-    def download_model(model_path_type: PathType, model_uri: str, model_dir: Path) -> bool:
+    def download_model(model_path_type: PathType, model_uri: str, model_dir: Path, model_obj: Model, registry_name: str) -> bool:
         """Prepare the Download Environment.
 
         :param model_path_type: Model path type
@@ -92,7 +117,7 @@ class ModelDownloadUtils:
         if model_path_type == PathType.GIT:
             return ModelDownloadUtils._download_git_model(model_uri, model_dir)
         if model_path_type == PathType.AZUREBLOB:
-            return ModelDownloadUtils._download_azure_artifacts(model_uri, model_dir)
+            return ModelDownloadUtils._download_azure_artifacts(model_uri, registry_name, model_obj)
         else:
             logger.print("Unsupported Model Download Method.")
         return False
--- a/scripts/release/asset_publish.py
+++ b/scripts/release/asset_publish.py
@@ -106,7 +106,7 @@ def update_spec(asset: Union[Component, Environment, Model], spec_path: Path) ->
     return False
 
 
-def prepare_model(model_config: assets.ModelConfig, spec_file_path: Path, model_dir: Path) -> bool:
+def prepare_model(model_config: assets.ModelConfig, spec_file_path: Path, model_dir: Path, registry_name: str) -> bool:
     """Prepare model.
 
     :param model_config: Model Config object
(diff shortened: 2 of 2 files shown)


scripts/azureml-assets/azureml/assets/model/utils.py
```diff
@@ -10,7 +10,13 @@
 import sys
 from azureml.assets import PathType
 from azureml.assets.util import logger
+import requests
+from azure.identity import DefaultAzureCredential
+from azure.ai.ml.entities import Model
 
+CREDENTIAL = DefaultAzureCredential()
+HEADERS = {"Authorization": "Bearer " + CREDENTIAL.get_token('https://management.azure.com/.default').token}
+ASSET_STORE_URL = "https://eastus.api.azureml.ms/assetstore/v1.0"
```
skasturi, line 19: Could we figure out at runtime primary region for the registry and then create the URL according to that? Currently, I see that it is east US. Wondering if that needs to change in case the registry has a different primary region. We might have to look at the available API and see if they programmatically allow figuring out what the primary region is.

scripts/azureml-assets/azureml/assets/model/utils.py
```diff
@@ -42,6 +52,23 @@ def _run(cmd, cwd: Path = "./") -> int:
...
+            "assetId": asset_id,
+            "temporaryDataReferenceId": "model_weights",
+            "temporaryDataReferenceType": "TemporaryBlobReference",
+        }
+        response = requests.post(
+            ASSET_STORE_URL + "/temporaryDataReference/createOrGet",
+            json=pending_upload_body,
+            headers=HEADERS,
+        )
+        pending_upload_response = response.json()
```
novaturient95, line 69: try catch? retry mechanism?

novaturient95, line 69: try catch? retry mechanism?

scripts/release/asset_publish.py
```diff
@@ -26,14 +26,24 @@
...
 
-ASSET_ID_TEMPLATE = Template("azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version")
+ASSET_ID_TEMPLATE = Template(
+    "azureml://registries/$registry_name/$asset_type/$asset_name/versions/$version"
+)
 TEST_YML = "tests.yml"
 PROD_SYSTEM_REGISTRY = "azureml"
-CREATE_ORDER = [assets.AssetType.DATA, assets.AssetType.ENVIRONMENT, assets.AssetType.COMPONENT,
-                assets.AssetType.MODEL]
+CREATE_ORDER = [
```
novaturient95, line 34: Could you run below:
`black --line-length 119 ./`

scripts/release/asset_publish.py
```diff
@@ -55,9 +65,9 @@ def find_test_files(dir: Path):
         with open(test / TEST_YML) as fp:
             data = YAML().load(fp)
             for test_group in data.values():
-                for test_job in test_group['jobs'].values():
-                    if 'job' in test_job:
-                        test_jobs.append((test / test_job['job']).as_posix())
+                for test_job in test_group["jobs"].values():
```
novaturient95, line 68: dictionary keys should continue to have `'jobs'`

vizhur (dismissed):

scripts/azureml-assets/azureml/assets/model/utils.py
novaturient95: For now keeping the hardcoding for internal and prod regions registries

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -0,0 +1,173 @@
...
+    RETRY_COUNT = 3
+
+    def get_assetstore_url(registry_name: str):
+        """Return asset URI for the registry."""
+        # USE https://eastus.api.azureml.ms/registrymanagement/v1.0/registries/{registryName}/discovery
+        if RegistryUtils.DEV_REGISTRY_NAME == registry_name:
+            return RegistryUtils.INTERNAL_REGION_ASSET_STORE_URL
+        return RegistryUtils.PROD_REGION_ASSET_STORE_URL
+
+    def get_registry_data_reference(registry_name: str, model_name: str, model_version: str):
```
elesel, line 33: This works as a proof-of-concept, but we really should leverage the SDK as much as possible, which will allow this to support sovereign and air-gapped regions after model storage accounts are created there. You'd need to create an MLClient ([example](https://msdata.visualstudio.com/Vienna/_git/azureml-asset?path=/msftkube/scripts/create_environments.py)), which then gives you most of what you'd need to call [get_sas_uri_for_registry_asset](https://github.com/Azure/azure-sdk-for-python/blob/3c9b59c72f6229bd3986c92315b388f920ccc0c1/sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_registry_utils.py#L81). This is how the SDK calls it from [model create/update](https://github.com/Azure/azure-sdk-for-python/blob/3c9b59c72f6229bd3986c92315b388f920ccc0c1/sdk/ml/azure-ai-ml/azure/ai/ml/operations/_model_operations.py#L171).

elesel: I'm a little leery about introducing azure-ai-ml dependency, but maybe this just needs to happen at some point anyway.

scripts/release/asset_publish.py
elesel, line 1: Please undo all of the changes that `black` made to this file that have no effect on its operation. It's nearly impossible to identify the substantive changes that were actually made.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -0,0 +1,156 @@
...
+            except Exception as e:
+                cnt += 1
+                logger.log_error(f"Exception in fetching data reference. Retry count {cnt}. Error: {e}")
+        if not blob_uri or not sas_uri:
+            raise Exception(f"Unable to fetch data reference for asset {asset_id}")
+        return blob_uri, sas_uri
+
+    def _get_temp_data_ref(registry_name, model_name, model_version):
+        try:
+            credential = DefaultAzureCredential()
```
elesel, line 43: Please consider using `AzureCliCredential` instead, rather than . No other credential should be available during release. See https://learn.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#authenticate-via-the-azure-cli

vizhur (approved):

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -0,0 +1,156 @@
...
+                sas_uri = response.blob_reference_for_consumption.credential.additional_properties["sasUri"]
+                if not blob_uri or not sas_uri:
+                    raise Exception("Error in fetching BLOB or SAS URI")
+                break
+            except Exception as e:
+                cnt += 1
+                logger.log_error(f"Exception in fetching data reference. Retry count {cnt}. Error: {e}")
+        if not blob_uri or not sas_uri:
+            raise Exception(f"Unable to fetch data reference for asset {asset_id}")
+        return blob_uri, sas_uri
```
elesel, line 39: Simpler:
```suggestion
        cnt = 0
        asset_id = f"azureml://registries/{registry_name}/models/{model_name}/versions/{model_version}"
        logger.print(f"getting data reference for asset {asset_id}")
        while cnt < RegistryUtils.RETRY_COUNT:
            try:
                response = RegistryUtils._get_temp_data_ref(registry_name, model_name, model_version)
                blob_uri = response.blob_reference_for_consumption.blob_uri
                sas_uri = response.blob_reference_for_consumption.credential.additional_properties["sasUri"]
                if not blob_uri or not sas_uri:
                    raise Exception("Error in fetching BLOB or SAS URI")
                return blob_uri, sas_uri
            except Exception as e:
                cnt += 1
                logger.log_error(f"Exception in fetching data reference. Retry count {cnt}. Error: {e}")
        else:
            raise Exception(f"Unable to fetch data reference for asset {asset_id}")
```

scripts/azureml-assets/setup.py
elesel, line 1: azure-ai-ml should be added as a dependency.

novaturient95: Done

scripts/azureml-assets/azureml/assets/model/model_utils.py
novaturient95: Done

novaturient95: But it says both AzureCliCredential and DefaultAzureCredential is available during CLI login. Isn't that the case during release scenario?

scripts/release/asset_publish.py
novaturient95: Ohhh this will take some time. the update is just in `prepare_model_for_registration` and its call below in `__main__`

scripts/azureml-assets/azureml/assets/model/model_utils.py
elesel: DefaultAzureCredential tries a lot of login methods and stops at the first one that succeeds. I'd rather we just go for the correct one in the first place, which AzureCliCredential provides.

novaturient95: yes true. Updated

scripts/release/asset_publish.py
elesel: Please consider grabbing the latest from the main branch and making the updates there instead of undoing `black`'s changes. We have a lot of developers working in this repo and I don't want to force everyone to onboard to using `black`. I'm fine with it being used on new files, though.

novaturient95: updated and validated for model flow

scripts/release/requirements.txt
```diff
@@ -1,5 +1,5 @@
 azureml-assets
-azure-ai-ml==1.7.0
+azure-ai-ml>=1.9.0
```
elesel, line 2: Please remove this. The azure-ai-ml dependency will come via azureml-assets now.
```suggestion
```

elesel (approved):

PR #971, opened 2023-08-03
falcon onboard

skanakamedal:


aggarwal-k (approved):

assets/training/finetune_acft_hf_nlp/components/pipeline_components/question_answering/spec.yaml
```diff
@@ -114,6 +114,18 @@ inputs:
     default: 1000
     description: Number of examples to batch before calling the tokenization function
 
+  pad_to_max_length:
+    type: boolean
```
jpmann, line 118: why is the type boolean. It should be string right? Internally, we use str2bool to convert it to boolean. Isn't that the case?

skanakamedal: yes we can do that as well. str2bool  is handeling the case sensitivity. Do you want me to change to str and keep enum?

jpmann (approved):

amangupta26 (approved):

PR #991, opened 2023-08-07
Update model naming pattern

novaturient95:


scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -234,7 +237,8 @@ def validate_name(asset_config: assets.AssetConfig) -> int:
     asset_name = asset_config.name
 
     # Check against generic naming pattern
-    if not COMMON_NAME_PATTERN.match(asset_name):
+    if ((asset_config.type is assets.AssetType.MODEL and not MODEL_NAME_PATTERN.match(asset_name))
+            or (asset_config.type is not assets.AssetType.MODEL and not COMMON_NAME_PATTERN.match(asset_name))):
```
elesel, line 241: Consider this to make it simpler, and then remember to bump the version in setup.py and update the CHANGELOG.
```suggestion
    if not ((asset_config.type is assets.AssetType.MODEL and MODEL_NAME_PATTERN.match(asset_name))
            or COMMON_NAME_PATTERN.match(asset_name)):
```

PR #993, opened 2023-08-07
Validate source tree for config/yml files and file size limitation

lykelly19:


scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Fail if any asset.yml in the tree
+    for asset_yml_file in util.find_files(input_dirs, "asset.yml"):
+        logger.log_error(f"{asset_yml_file} should be named asset.yaml")
+        error_count += 1
+
+    # Fail if any spec.yml in the tree
+    for spec_yml_file in util.find_files(input_dirs, "spec.yml"):
+        logger.log_error(f"{spec_yml_file} should be named spec.yaml")
+        error_count += 1
```
elesel, line 44: We probably don't have to be this picky because asset.yaml could refer to a spec file named anything. Consider removing this one.

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Scan every file in the source tree to be 1 MB or less
+    for file in util.find_files(input_dirs, "*"):
+        if os.path.getsize(file) > BYTES_IN_A_MEGABYTE:
```
elesel, line 48:
```suggestion
        if file.stat().st_size > BYTES_IN_A_MEGABYTE:
```

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,70 @@
...
+            logger.log_error(f"{file} should not be over 1 MB")
```
elesel, line 49: Would something like this be clearer?
```suggestion
            logger.log_error(f"{file} is too large (over 1MB)")
```

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,64 @@
...
+            logger.log_error(f"{spec_config_file} does not have a corresponding asset.yaml")
+            error_count += 1
+
+    # Fail if any asset.yml in the tree
+    for asset_yml_file in util.find_files(input_dirs, "asset.yml"):
+        logger.log_error(f"{asset_yml_file} should be named asset.yaml")
+        error_count += 1
+
+    # Scan every file in the source tree to be 1 MB or less
+    for file in util.find_files(input_dirs, "*"):
```
elesel, line 41: What would you think about moving all of the other checks inside this loop so the directory tree only needs to be traversed once? [`Path.name`](https://docs.python.org/3/library/pathlib.html) gives you the name of the file, excluding its path.

lykelly19: I like this approach as it'll still allow for each check to be done separately in this for-loop

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,65 @@
...
+    Returns:
+        bool: True if files were successfully validated, otherwise False.
+    """
+    error_count = 0
+
+    for file in util.find_files(input_dirs, "*"):
+
+        # Check that every spec.yaml has asset.yaml next to it
+        if file.name == "spec.yaml":
+            asset_config_file = file.parent / "asset.yaml"
```
elesel, line 32: Please use [this constant](https://github.com/Azure/azureml-assets/blob/ed2b533f1d4970621932dae773aff82d08261e3d/scripts/azureml-assets/azureml/assets/config.py#L110) instead of hard-coding asset.yaml. Also applies to the error message two lines down.

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,65 @@
...
+            if not asset_config_file.exists():
+                logger.log_error(f"{file} does not have a corresponding asset.yaml")
+                error_count += 1
+
+        # Fail if any asset.yml in the tree
+        if file.name == "asset.yml":
```
elesel, line 38: You could use an `elif` here instead.

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,64 @@
...
+        bool: True if files were successfully validated, otherwise False.
+    """
+    error_count = 0
+
+    for file in util.find_files(input_dirs, "*"):
+        if file.name == "spec.yaml":
+            # Check that every spec.yaml has asset.yaml next to it
+            asset_config_file = file.parent / assets.DEFAULT_ASSET_FILENAME
+            if not asset_config_file.exists():
+                logger.log_error(f"{file} does not have a corresponding asset.yaml")
```
elesel, line 34:
```suggestion
                logger.log_error(f"{file} does not have a corresponding {assets.DEFAULT_ASSET_FILENAME}")
```

scripts/azureml-assets/azureml/assets/validate_tree.py
```diff
@@ -0,0 +1,64 @@
...
+                error_count += 1
+        elif file.name == "asset.yml":
+            # Fail if any asset.yml in the tree
+            logger.log_error(f"{file} should be named asset.yaml")
```
elesel, line 38:
```suggestion
            logger.log_error(f"{file} should be named {assets.DEFAULT_ASSET_FILENAME}")
```

elesel (approved):

PR #995, opened 2023-08-07
Update scoring script chat completion format

svaruag:


srsaggam (approved):
Consider putting back asserts if needed.

PR #998, opened 2023-08-08
Add two promptflow samples as custom models in registry

dorisjoy:
Add below two promptflow samples as custom models in registry, these two modes are stored in this [storage account ](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/96aede12-2f73-41cb-b983-6d11a904839b/resourceGroups/shared-promptflow-rg/providers/Microsoft.Storage/storageAccounts/promptflowsamples/storagebrowser)

Usage: Promptflow service will download these two models from registry and show them as promptflow samples in different workspaces.


assets/models/system/promptflow-chat-with-wikipedia/asset.yaml
```diff
@@ -0,0 +1,4 @@
+extra_config: model.yaml
+spec: spec.yaml
+type: model
+categories: ["Models"]
```
vizhur, line 4: This should be category of an asset to be properly grouped here https://github.com/Azure/azureml-assets/wiki/models-documentation. Like Bert,  GPT. Please also provide description

dorisjoy: Use "prompt flow" category to group prompt flow related modesl. Add description for these two models. We will refine the description before we decided to release these two model to prod registry. Here we only want to go through the e2e path in test registry.

vizhur (approved):

PR #1008, opened 2023-08-09
Pin langchain in rag envs to 0.0.220 

tot0:
To ensure that faiss langchain indexes are PromptFlow compatible.


nanditamohan (dismissed):

gjwoods:
Looks like the Rag Env is using 0.0.149 - should we sync? assets/large_language_models/rag/environments/rag/context/conda_dependencies.yaml

- langchain==0.0.149

nanditamohan (approved):

PR #1017, opened 2023-08-10
Bump all llm_rag components again so they will pick up new langchain in azureml-staging

tot0:


shivamMg (commented):
Do we also need to update langchain in: `assets/large_language_models/rag/environments/rag/context/conda_dependencies.yaml`

shivamMg (approved):

PR #1054, opened 2023-08-16
Model card test type and min-sku changes

sarthaks95:


novaturient95:
Do we want to republish models on a new version (will need to increment current version) or waiting for metadata update supported for models?

cc: @HrishikeshGeedMS

skasturi:
I would prefer metadata only update rather than having to re-publish.

skasturi (dismissed):

skasturi (approved):

assets/models/system/finiteautomata-bertweet-base-sentiment-analysis/spec.yaml
```diff
@@ -4,8 +4,8 @@ path: ./
 properties:
   SHA: 924fc4c80bccb8003d21fe84dd92c7887717f245
   datasets: ''
-  evaluation-min-sku-spec: 2|0|7|14
-  evaluation-recommended-sku: Standard_DS2_v2
+  evaluation-min-sku-spec: 8|0|28|56
```
haritha-bendapudi, line 7: Why is the minimum number of cpu cores 8 while the recommended sku is DS4_v2?

haritha-bendapudi (approved):

PR #1056, opened 2023-08-16
Implement component to publish metrics to Azure monitor

yetamsft:


assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
```diff
@@ -0,0 +1,62 @@
+$schema: http://azureml/sdk-2-0/SparkComponent.json
+type: spark
+
+name: model_monitor_metric_outputter
```
mastloui-msft, line 4: should this be model_monitor_metric_publisher

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
```diff
@@ -0,0 +1,62 @@
...
+  signal_metrics:
+    type: mltable
+    mode: direct
+  monitor_name:
+    type: string
+  signal_name:
+    type: string
+  location:
+    type: string
+  workspace_resource_id:
```
mastloui-msft, line 24: do you have that information from within the component?

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+inputs:
+  signal_metrics:
+    type: mltable
+    mode: direct
+  monitor_name:
+    type: string
+  signal_name:
+    type: string
+  location:
```
mastloui-msft, line 22: Is there a way we don't need to be location aware for publishing custom metrics?

assets/model_monitoring/components/src/model_monitor_metric_publisher/azure_monitor_metric_publisher.py
```diff
@@ -0,0 +1,104 @@
+import datetime
+import json
+import requests
+
+from pyspark.sql import Row
+from typing import List
+
+from shared_utilities.log_utils import log_utils
+
+def publish_metric(
```
mastloui-msft, line 10: We should probably only publish metrics from the metric publisher component - the outputter only creates the monitoring metric files right?

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
yetamsft: I did find the `workspace_resource_id` from the runtime env vars (`os.environ['AZUREML_WORKSPACE_SCOPE']`), assuming the workspace running the job is always going to be the same where the metrics go. Though I am unsure how reliable this is in future. To me, it is easier to construct (or control) in MoMo.

assets/model_monitoring/components/src/model_monitor_metric_publisher/azure_monitor_metric_publisher.py
yetamsft: Yes, the outputter changes are not relevant in PR (simple rename), which I can move out to another one for ease to review.

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
yetamsft: Similar to the workspace_resource_id, there is some env var we can grok this infromation from, namely `Env var AZUREML_SERVICE_ENDPOINT: https://eastus2euap.api.azureml.ms`. But again we do not how sure we can depend on them. Thoughts?

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
yetamsft: Fixed.

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
yetamsft: After a second thought, both of the variables seem to be generally useful for any jobs. For simplicity of the caller, I changed to pull the location and workspace resource id from the env vars, with some assertion to make sure they exists and are valid.

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
yetamsft: Same to the comment above. Updated to extract from the env var.

assets/model_monitoring/components/src/model_monitor_metric_publisher/azure_monitor_metric_publisher.py
yetamsft: Moved the irrelevant changes out.

assets/model_monitoring/components/src/model_data_collector_preprocessor/run.py
```diff
@@ -20,13 +20,6 @@
 from shared_utilities.constants import MDC_CORRELATION_ID_COLUMN, MDC_DATA_COLUMN
 
 
-def _format_date_string(format: str, date_to_format: str) -> datetime:
```
chupadhy, line 23: Nice moving it to a util

assets/model_monitoring/components/src/model_monitor_metric_publisher/run.py
```diff
@@ -0,0 +1,57 @@
...
+    arg_parser.add_argument("--monitor_name", type=str)
+    arg_parser.add_argument("--signal_name", type=str)
+    arg_parser.add_argument("--data_window_start", type=str)
+    arg_parser.add_argument("--data_window_end", type=str)
+    args = arg_parser.parse_args()
+
+    metrics: List[Row] = read_mltable_in_spark(args.signal_metrics).collect()
+
+    try:
+        azureml_service_endpoint = os.environ['AZUREML_SERVICE_ENDPOINT']
```
chupadhy, line 31: Who sets these envs? Also, it is possible to not have an online endpoint associated with the model monitor no? Then do we expect the env var to be an empty string or something like that?

assets/model_monitoring/components/src/model_monitor_metric_publisher/run.py
```diff
@@ -0,0 +1,57 @@
...
+        workspace_resource_id = os.environ['AZUREML_WORKSPACE_SCOPE']
+    except KeyError as err:
+        log_error(f"Failed to find a required environment variable, error: {str(err)}.")
+        raise
+
+    location = extract_location_from_aml_endpoint_str(azureml_service_endpoint)
```
chupadhy, line 37: Is it possible to Get on WorkspaceResourceId to get the location info instead?

yetamsft: No, I don't think so. `WorkspaceResourceId` doesn't contain location info, which is like, `/subscriptions/{sub_id}/resourceGroups/{rg_name}/providers/Microsoft.MachineLearningServices/workspaces/{ws_name}`.

assets/model_monitoring/components/src/model_monitor_metric_publisher/run.py
yetamsft: I think this is set by Execution Service (or AEther), which is also used for syncing AML run metadata etc.
This is not the "online endpoint" resource, but the AML service (cluster) endpoint, which usually looks like "https://eastus2.api.azureml.ms". I believe this is quite stable and unlikely to change. Of course, we can revisit if we see a problem later.

assets/model_monitoring/components/model_monitor/model_monitor_metric_publisher/spec.yaml
```diff
@@ -0,0 +1,63 @@
+$schema: http://azureml/sdk-2-0/SparkComponent.json
+type: spark
+
+name: model_monitor_metric_publisher
+display_name: Model Monitor - Metric Publisher
```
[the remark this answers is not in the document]
yetamsft: Not that I am aware of. Done.

chupadhy (approved):

mastloui-msft (approved):

PR #1087, opened 2023-08-23
Fix ffmpeg image vulnerability by replacing base image with ubuntu22.10

romanlutz:
Due to CVE https://ubuntu.com/security/notices/USN-5958-1 we have to upgrade ffmpeg and related packages. That's not possible in ubuntu20.04 so this PR gets us to use ubuntu22.10. There's no published base image from AzureML for this, so I copied over the code from https://github.com/Azure/AzureML-Containers/blob/master/base/cpu/openmpi4.1.0-ubuntu22.04/Dockerfile with two small modifications. This worked fine locally.


imatiach-msft (dismissed):

assets/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
```diff
@@ -12,19 +108,19 @@ ENV PATH $AZUREML_CONDA_ENVIRONMENT_PATH/bin:$PATH
...
 RUN pip install --pre 'azure-ai-ml' 'azure-storage-blob<=12.13.0'
 
 # Install azureml packages
-RUN pip install 'azureml-core=={{latest-pypi-version}}' \
-                'azureml-mlflow=={{latest-pypi-version}}' \
+RUN pip install 'azureml-core' \
+                'azureml-mlflow' \
                 'azureml-rai-utils==0.0.3'
 
-RUN pip install 'azureml-dataset-runtime=={{latest-pypi-version}}' \
```
imatiach-msft, line 26: this worries me a bit since the env management team always recommends to pin these to {{latest-pypi-version}}

romanlutz: Oh, this was actually just removed for the local validation build. It doesn't like placeholders :-) For this env definition it's fine since the build in the repo knows how to take care of it. I had the change to revert this locally, just forgot to push. Thanks for the reminder!

kicha0 (approved):

PR #1088, opened 2023-08-23
Update metrics outputter to accept group as an optional column

yetamsft:
TODO: unit test the added scenario

### DIFF
--- a/assets/model_monitoring/components/src/model_monitor_metric_outputter/builder/metric_output_builder.py
+++ b/assets/model_monitoring/components/src/model_monitor_metric_outputter/builder/metric_output_builder.py
@@ -44,7 +44,8 @@ class MetricOutputBuilder:
                     root_group = metric_dict[SIGNAL_METRICS_GROUP]
                     group_names.append(root_group)
 
-                    if metric_dict[SIGNAL_METRICS_GROUP_DIMENSION].lower() != AGGREGATE:
+                    # If group dimension is not specified, default is Aggregate.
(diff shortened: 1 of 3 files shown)


assets/model_monitoring/components/src/model_monitor_metric_outputter/builder/metric_output_builder.py
```diff
@@ -35,34 +35,31 @@ def _build(self, metrics: List[Row]) -> dict:
...
-                    group_names.append(node_id)
+            try:
+                metric_name = metric_dict[SIGNAL_METRICS_METRIC_NAME]
+
+                group_names = []
+                if SIGNAL_METRICS_GROUP in metric_dict:
+                    root_group = metric_dict[SIGNAL_METRICS_GROUP]
+                    group_names.append(root_group)
 
                     if metric_dict[SIGNAL_METRICS_GROUP_DIMENSION].lower() != AGGREGATE:
```
yetamsft, line 47: TODO: need to check whether GROUP_DIMENSION exists here. Currently, it throws KeyError, which is a bug.

chupadhy: Done

assets/model_monitoring/components/tests/unit/test_metric_output_builder.py
```diff
@@ -77,6 +78,118 @@ def test_metrics_with_1_level_groups(self):
             },
         }
 
+    def test_empty_groups_and_group_dimensions_success(self):
```
chupadhy, line 81: This change will be merged as part of another PR

PR #1098, opened 2023-08-24
Hrishikesh/model metadata updates

HrishikeshGeedMS:
This PR add new features to asset_create pipeline, that will help to update model_metadata if the model already exists in the system registry.


scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +153,52 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+        model_config: assets.ModelConfig,
+        registry_name: str, 
+        update: AssetVersionUpdate = None,
+    ):
+    """Update the mutable metadata of already registered Model asset."""
+    try:
+        ml_client = MLClient(credential= AzureCliCredential(), registry_name=registry_name)
+        model = ml_client.models.get(name= model_name, version= model_version)
+
+        if update is not None: ## to update older version of models
```
novaturient95, line 171: will there be a case when update is not None?

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +153,52 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+        model_name: str, 
+        model_version: str, 
+        spec_path: Path, 
+        model_config: assets.ModelConfig,
+        registry_name: str, 
+        update: AssetVersionUpdate = None,
+    ):
+    """Update the mutable metadata of already registered Model asset."""
+    try:
+        ml_client = MLClient(credential= AzureCliCredential(), registry_name=registry_name)
```
novaturient95, line 168: we can get rid of mlclient operation here and just make use of model_config and spec_path

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +153,52 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+
+            model_description_file_path = Path(spec_path).parent / model_config.description
+            logger.print(f"model_description_file_path {model_description_file_path}")
+            if os.path.exists(model_description_file_path):
+                with open(model_description_file_path) as f:
+                    model.description = f.read()
+            else:
+                logger.print(f"description file does not exist for model {model_name}")
+
+        ml_client.models.create_or_update(model)
```
novaturient95, line 202: lets make use of CLI as much as possible? may be add asset_upate function similar to asset_create?

scripts/azureml-assets/azureml/assets/model/model_utils.py
HrishikeshGeedMS: Yes,update will be used only when we want to update older versions of model from ADO deploy configs file.
For latest models, updates will be none and we will use only spec files to get tags, description information

scripts/azureml-assets/azureml/assets/model/model_utils.py
HrishikeshGeedMS: az ml model update does not use spec path as such ref- https://learn.microsoft.com/en-us/cli/azure/ml/model?view=azure-cli-latest#:~:text=Global%20Parameters-,az%20ml%20model%20update,-Update%20a%20model


Also if we use CLI, It uses commands like 

az ml model update -- set.tags.tag1=value1 -- set.tags.tag2=value1 --remove.tags.tag3

for multiple tags it would be very cumbersome to maintain, similarly for removing some tags as well. It will be more error prone.It will be very long command.

Using sdk makes it quite simple just update the tags for model object and just update.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +153,51 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
         return None, False
+
+
+def update_model_metadata(
+        model_name: str,
+        model_version: str,
+        spec_path: Path,
+        model_config: assets.ModelConfig,
+        registry_name: str,
+        update: AssetVersionUpdate = None):
```
elesel, line 164: Please consider calling this function with just an AssetVersionUpdate object (instead of spec_path and model_config), which can either come from a deployment config file or be created using values from a spec.yaml file.

elesel:
You'll also need to increment azureml-assets's setup.py version and update the CHANGELOG.md file in order for this change to reach PyPI.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +153,51 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+
+                    # Delete tags
+                    if update.tags.delete:
+                        for k in update.tags.delete:
+                            model.tags.pop(k, None)
+
+        else:  # update the model using spec files for latest version
+            with open(spec_path) as f:
+                model_spec = YAML().load(f)
+                model.tags = model_spec["tags"]
```
novaturient95, line 191: null check or tags are always going to be there?

scripts/azureml-assets/azureml/assets/model/model_utils.py
novaturient95: yeah make sense. Thanks Hrishikesh

novaturient95: we would need mlclient to fetch current models details and though we can do that via cli as well, but unnecessary parsing would be needed,. For tags, we would need to build command str iterating through all the keys. Current approach make sense to me. The reason I asked to check, is because we had faced some issue in creating mlclient earlier (that could be because if using DefaultAzureCredential. I am not very sure of its reason as of now). But since we are also creating AzureCLICredential for registry utils and its working, I guess we should be okay

scripts/azureml-assets/azureml/assets/model/model_utils.py
novaturient95: Ignore this comment

scripts/azureml-assets/azureml/assets/model/model_utils.py
HrishikeshGeedMS: Added in update object. Thank you

scripts/azureml-assets/azureml/assets/model/model_utils.py
HrishikeshGeedMS: Done. Thankyou Louie

scripts/release/asset_publish.py
```diff
@@ -402,6 +405,49 @@ def create_asset(
...
+            model_description_file_path = Path(spec_path).parent / model_config.description
+            logger.print(f"model_description_file_path {model_description_file_path}")
+            if os.path.exists(model_description_file_path):
+                with open(model_description_file_path) as f:
+                    description = f.read()
+        except Exception as e:
+            logger.log_error(f"Failed to get description for model {model_name}: {e}")
+
+        update = AssetVersionUpdate(versions=[model_version],
+                                    tags=tags,
```
skasturi, line 438: Do we update even if loading of tags/description fail?

skasturi: Could you please add UTs to cover such cases?

HrishikeshGeedMS: No, if tags,description won't load it will get set as None, and for none type values wont get affected. It will only update if and only if it get valid description and tag

scripts/azureml-assets/azureml/assets/deployment_config.py
```diff
@@ -58,7 +58,7 @@ class AssetTags:
     """
 
     add: Dict[str, str] = None
-    replace: Dict[str, str] = None
+    replace: Dict[str, Union[list, str]] = None
```
elesel, line 61: Please revert this. Tag values are strings. If any other datatypes need to be handled, conversion should be upstream.

scripts/azureml-assets/azureml/assets/deployment_config.py
```diff
@@ -201,8 +201,8 @@ def should_create(self, asset_type: assets.AssetType, asset_name: str) -> bool:
 class TagsSchema(Schema):
     """Tags schema."""
 
-    add = fields.Dict(fields.Str(), fields.Str())
-    replace = fields.Dict(fields.Str(), fields.Str())
+    add = fields.Dict(fields.Str())
+    replace = fields.Dict(fields.Str())
```
elesel, line 205: Please revert this. Tag values are strings. If any other datatypes need to be handled, conversion should be upstream.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +152,37 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+
+
+def update_model_metadata(
+        model_name: str,
+        model_version: str,
+        registry_name: str,
+        update: AssetVersionUpdate):
+    """Update the mutable metadata of already registered Model."""
+    try:
+        ml_client = MLClient(credential=AzureCliCredential(), registry_name=registry_name)
```
elesel, line 164: Consider providing an optional ml_client arg to the function and only create the MLClient if the arg is none, or just create the MLClient outside of this function entirely. Creating MLClient is somewhat expensive and results in multiple calls to retrieve the token and call registry discovery.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +152,37 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+                    if update.tags.add:
+                        for k, v in update.tags.add.items():
+                            model.tags[k] = v
+
+                    # Delete tags
+                    if update.tags.delete:
+                        for k in update.tags.delete:
+                            model.tags.pop(k, None)
+
+        ml_client.models.create_or_update(model)
```
elesel, line 185: Consider identifying whether any changes were made to the model object, in which case an update wouldn't be necessary.

scripts/release/asset_publish.py
```diff
@@ -402,6 +405,49 @@ def create_asset(
...
+        except Exception as e:
+            logger.log_error(f"Failed to get tags for model {model_name}: {e}")
+
+        # get updated description for model
+        try:
+            model_description_file_path = Path(spec_path).parent / model_config.description
+            logger.print(f"model_description_file_path {model_description_file_path}")
+            if os.path.exists(model_description_file_path):
+                with open(model_description_file_path) as f:
+                    description = f.read()
```
elesel, line 433: Retrieval of the contents of the description file should be added to the ModelConfig class.

scripts/azureml-assets/azureml/assets/deployment_config.py
novaturient95: @elesel tag values supports list and dict now as well for all asset types

scripts/azureml-assets/azureml/assets/model/model_utils.py
novaturient95: Updated. Could you please help to review

scripts/azureml-assets/azureml/assets/deployment_config.py
novaturient95: this was probably updated very recently

scripts/release/asset_publish.py
novaturient95: Done

scripts/azureml-assets/azureml/assets/deployment_config.py
novaturient95: okay I found one issue. Though you can push a list, but a translation to string happens and eventually it's a list. Updating

scripts/azureml-assets/azureml/assets/deployment_config.py
novaturient95: Done

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -554,10 +555,22 @@ def _publish(self) -> Dict[str, object]:
...
+    @property
+    def description(self) -> Dict[str, object]:
+        """Model description."""
+        if self._description_file_path and not self._description:
+            model_description_file_path = self._file_path / self._description_file_path
+            try:
+                with open(model_description_file_path) as f:
+                    self._description = f.read()
+            except Exception as e:
+                print(f"Error in reading description from {model_description_file_path}. Error: {e}")
```
novaturient95, line 571: importing logger was causing circular dependency error. using print

scripts/release/asset_publish.py
```diff
@@ -402,6 +406,47 @@ def create_asset(
...
+                tags = model_spec.get("tags", {})
+                # convert tag value to string
+                for name, value in tags.items():
+                    if isinstance(value, list):
+                        value = str(value)
+                    elif isinstance(value, dict):
+                        value = json.dumps(value)
+                    elif not isinstance(value, str):
+                        raise Exception(f"Invalid value type: {type(value)} for tag name {name}")
+                    tags[name] = value
```
elesel, line 431: Interesting. Does the CLI support non-string tag values this way too?

scripts/azureml-assets/azureml/assets/config.py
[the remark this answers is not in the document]
novaturient95: updated. Could you please help to recheck

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +143,46 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+    model_name: str,
+    model_version: str,
+    update: AssetVersionUpdate,
+):
+    """Update the mutable metadata of already registered Model."""
+    try:
+        model = mlclient.models.get(name=model_name, version=model_version)
+
+        need_update = False
+        if update is not None:  # to update older version of models
```
elesel, line 159: I think you can remove this line.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +143,46 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+                if update.tags.replace is not None:
+                    updated_tags = update.tags.replace
+                elif update.tags.add is not None:
+                    for k, v in update.tags.add.items():
+                        updated_tags[k] = v
+                elif update.tags.delete is not None:
+                    for k in update.tags.delete:
+                        updated_tags.pop(k, None)
+
+            if updated_tags != model.tags:
```
elesel, line 172: You should be able to nest this under `if update.tags`.

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +143,46 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+                logger.print("tags has been updated.")
+                model.tags = updated_tags
+                need_update = True
+
+            if update.description is not None and model.description != update.description:
```
elesel, line 177: I don't think `update.description` can ever be `None` because it's initialized to the empty string in the class.

scripts/azureml-assets/azureml/assets/model/model_utils.py
novaturient95: yeah !!! this will always be a non null value for current flow

scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -151,3 +143,46 @@ def prepare_model(spec_path, model_config, registry_name, temp_dir):
...
+                    for k, v in update.tags.add.items():
+                        updated_tags[k] = v
+                elif update.tags.delete is not None:
+                    for k in update.tags.delete:
+                        updated_tags.pop(k, None)
+
+            if updated_tags != model.tags:
+                logger.print("tags has been updated.")
+                model.tags = updated_tags
+                need_update = True
```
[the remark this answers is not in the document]
novaturient95: I would need to add this for all 3 case (add, replace, delete). Would it be better to keep just at one place?

scripts/azureml-assets/azureml/assets/model/model_utils.py
novaturient95: Done

elesel (dismissed):

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -554,9 +557,19 @@ def _publish(self) -> Dict[str, object]:
         return self._yaml.get('publish')
 
     @property
-    def description(self) -> Dict[str, object]:
+    def _description_file_path(self) -> Path:
         """Model description."""
```
skasturi, line 561: (nit) description file path

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -554,9 +557,19 @@ def _publish(self) -> Dict[str, object]:
         return self._yaml.get('publish')
 
     @property
-    def description(self) -> Dict[str, object]:
+    def _description_file_path(self) -> Path:
```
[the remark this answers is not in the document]
elesel: Validation would fail before this returns a `None`. The function could just be simplified to `return self._file_path / self._publish.get('description')`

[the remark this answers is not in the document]
novaturient95: Updated

scripts/azureml-assets/azureml/assets/config.py
novaturient95: Done

elesel (approved):

PR #1113, opened 2023-08-26
Responsible AI: unwrap sklearn models when possible

romanlutz:
Mirroring change from https://github.com/Azure/RAI-vNext-Preview/pull/201 to avoid having regression models that fail RAIInsights validation because `hasattr(model, 'predict_proba')` is True (because an object is returned) rather than False.

### DIFF
--- a/assets/responsibleai/tabular/components/src/create_rai_insights.py
+++ b/assets/responsibleai/tabular/components/src/create_rai_insights.py
@@ -177,6 +177,10 @@ def main(args):
             model_path=args.model_input,
         )
 
+    # unwrap the model if it's an sklearn wrapper
+    if model_estimator.__class__.__name__ == "_SklearnModelWrapper":
+        model_estimator = model_estimator.sklearn_model
+
     constructor_args = create_constructor_arg_dict(args)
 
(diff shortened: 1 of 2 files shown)


assets/responsibleai/tabular/components/src/create_rai_insights.py
```diff
@@ -177,6 +177,10 @@ def main(args):
             model_path=args.model_input,
         )
 
+    # unwrap the model if it's an sklearn wrapper
+    if model_estimator.__class__.__name__ == "_SklearnModelWrapper":
```
Advitya17, line 181: Nit: maybe enum for the string?

Advitya17 (approved):

gaugup (approved):

assets/responsibleai/tabular/components/src/create_rai_insights.py
romanlutz: You probably mean constant?
For consistency with the preview repo I'm going to leave this unaddressed. We can add it in the next PR.

vinuthakaranth (approved):

PR #1125, opened 2023-08-29
torch update to 1.13.1

kicha0:
for vulnerability GHSA-47fc-vmwq-366v

### DIFF
--- a/assets/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
+++ b/assets/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
@@ -33,7 +33,8 @@ RUN pip install 'shap==0.40.0' \
                 'setuptools>=65.5.1' \
                 'numpy==1.22.0' \
                 'scipy==1.10.0' \
-                'statsmodels==0.14.0'
+                'statsmodels==0.14.0' \
+                'torch==1.13.1'
 
 RUN pip freeze
 


assets/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
```diff
@@ -33,9 +33,10 @@ RUN pip install 'shap==0.40.0' \
                 'setuptools>=65.5.1' \
                 'numpy==1.22.0' \
                 'scipy==1.10.0' \
-                'statsmodels==0.14.0'
+                'statsmodels==0.14.0' \
+                'torch==1.13.1'
```
imatiach-msft, line 37: this needs to be fixed in automl and not in our image - also, the fix is more complex than this, since we need to install it in conda

imatiach-msft (changes requested):
this issue needs to be fixed in automl instead, we are just downstream consumers of their packages

kicha0:
> this issue needs to be fixed in automl instead, we are just downstream consumers of their packages

I agree and it is understood. However, the issue is that this vulnerability have a due date of 09/01 which is this Friday. Before that if tests are passing I think we should address the issue, what do you think?

Added work item here to remove the pinning as automl updates their dependency.

assets/responsibleai/environments/responsibleai-vision-ubuntu20.04-py38-cpu/context/Dockerfile
kicha0: we've had success pinning statsmodel to a higher version by pinning the package in docker after conda installation is done. This pin for statsmodel fixed vulnerability due to automl requiring a lower version.

kicha0: yes and I agree this needs to be fixed in automl. We created a thread in automl channel, I think we should escalate there as well but before then we should also fix it here since this has a due date of 09/01.

imatiach-msft (approved):

PR #1145, opened 2023-08-31
upgrade changes in t5-small model

MallikharjunaThota123:


skasturi:
Looks like there are conflicts. please check.

skasturi (changes requested):
:clock1:

PR #1152, opened 2023-08-31
[MoMo] Fix the empty string issue in run metrics publisher

chupadhy:


assets/model_monitoring/components/src/model_monitor_metric_outputter/runmetrics_publisher.py
```diff
@@ -46,12 +46,12 @@ def _publish_metric_to_run(self, metrics: dict, metric_step: int):
         timeseries = metrics[TIMESERIES]
         metric_names = timeseries[TIMESERIES_METRIC_NAMES]
 
-        if VALUE in metrics:
+        if VALUE in metrics and metrics[VALUE] is not None and metrics[VALUE] != "":
```
yetamsft, line 49: Do we still want to publish if the value is `None`? I think we can skip the entire payload in this case since there isn't much value publishing the thresholds for blank data. 
This might be a question in theory only. I think we assume the metric computation always gives a value.

yetamsft (approved):

assets/model_monitoring/components/src/model_monitor_metric_outputter/runmetrics_publisher.py
chupadhy: Right - let me create a follow-up item to handle this gracefully. AK said pass in float('nan')

PR #1173, opened 2023-09-06
Add another 14 promptflow models into system registry

dorisjoy:
1. Add another 14 promptflow models into system registry.
2. Update model properties and model description for existing 4 promptflow model.

### DIFF
--- a/assets/promptflow/models/how-to-use-functions-with-chat-models/asset.yaml
+++ b/assets/promptflow/models/how-to-use-functions-with-chat-models/asset.yaml
@@ -1,0 +1,4 @@
+extra_config: model.yaml
+spec: spec.yaml
+type: model
+categories: ["prompt flow"]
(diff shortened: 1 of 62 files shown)


assets/promptflow/models/how-to-use-functions-with-chat-models/asset.yaml
```diff
@@ -0,0 +1,4 @@
+extra_config: model.yaml
```
DaweiCai, line 1: how-to-use-functions-with-chat-models the naming is not consistent with others. Is this a template? Can we rename it to template-functions?

dorisjoy: It is not a template. It shows in the gallery:
image
Just keep it as the example shown. And later we can discuss whether need to change the example name or not.

skasturi:
@dorisjoy Could you please move all the models to "assets/models" folder? Thank you.

PR #1175, opened 2023-09-06
Add support for custom data loaders [ON HOLD]

iamrk04:
This PR adds support for custom data loaders in the dataset downloader component. This gives the flexibility to download datasets from anywhere using the HF data loader script template.


assets/aml-benchmark/components/dataset-downloader/spec.yaml
```diff
@@ -9,19 +9,19 @@ version: 0.0.1
...
-    description: If a specific split of the HuggingFace dataset to download, specify the split name; specify 'all' to download all splits. Leave it null if url is specified.
-    optional: True
-  url:
-    type: string
-    description: URL of the dataset to download. Must be null if dataset_name is specified.
+    description: If a specific split of the dataset to download, specify the split name; specify 'all' to download all splits.
+    optional: False
+  script_path:
+    type: uri_file
+    description: Path to the dataset loading script. Must follow the HuggingFace dataset loading script template.
```
arun-rajora, line 24: Do we need to provide the link to the sample template? Also, are we allowed to add such links to external sources?

assets/aml-benchmark/components/src/dataset_downloader/loaders/math.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) Microsoft Corporation.
```
arun-rajora, line 1: Why are we hosting these scripts in our component code? I thought they would be input to the component? Shouldn't they be a data-asset in the azureml registry?

How do we invoke these scripts?

assets/aml-benchmark/components/src/dataset_downloader/main.py
```diff
@@ -47,7 +48,10 @@ def parse_args() -> argparse.Namespace:
         help="Split of the dataset to download; specify 'all' to download all splits.",
     )
     parser.add_argument(
-        "--url", type=str, default=None, help="URL of the dataset to download."
```
arun-rajora, line 50: Are we removing the functionality to download from a URL? Can you check if none of the P1 datasets were validated using it?

arun-rajora (changes requested):

assets/aml-benchmark/components/src/dataset_downloader/loaders/math.py
```diff
@@ -0,0 +1,100 @@
...
+# Licensed under the MIT License.
```
arun-rajora, line 2: Also, Eric mentioned about CELA requirements for adding a notice.txt file. Please make sure those requirements are properly implemented.

assets/aml-benchmark/components/src/dataset_downloader/loaders/math.py
iamrk04: With having these loader scripts in the component, people can directly use them by using the custom loader file_name, for example `mmlu`, in the component's argument `dataset_name`. Also, this eliminates the need to host these scripts in the registry.

assets/aml-benchmark/components/dataset-downloader/spec.yaml
iamrk04: That's exactly the reason I did not include the link. @EricWrightAtWork, do we have clarity on this, your thoughts?

assets/aml-benchmark/components/src/dataset_downloader/loaders/math.py
iamrk04: @EricWrightAtWork, could you please help in understanding what needs to be included to meet the requirements? Thanks!

assets/aml-benchmark/components/src/dataset_downloader/loaders/math.py
arun-rajora: Just a thought- is it possible that the component the script is in this repo (not in component) and we provide a link to it as URL_FILE input? That way, it becomes clear to the user how download is working?

iamrk04: yes, this will be better. I will add these loaders outside the component's code, and the url of the script can be sent in the `script_path` argument.

assets/aml-benchmark/components/src/dataset_downloader/main.py
iamrk04: We are good with this.

assets/aml-benchmark/components/src/dataset_downloader/loaders/math.py
arun-rajora: Please add a test for it (if not already done). We can make sure it works. Thanks!

iamrk04: test is already added.

PR #1192, opened 2023-09-10
[Model Monitor] - Add samples to metric capability to metrics outputter

mastloui-msft:


assets/model_monitoring/components/tests/unit/test_dict_utils.py
```diff
@@ -0,0 +1,219 @@
...
+            },
+        }
+
+        right = None
+
+        with pytest.raises(Exception):
+            merge_dicts(left, right)
+
+    def test_merge_with_int_values(self):
+        """Test merging dictionaries with a None dictionary."""
```
yetamsft, line 53: nit: int value

yetamsft (approved):
LGTM!

assets/model_monitoring/components/tests/unit/test_dict_utils.py
mastloui-msft: I'll address in a follow-up!

PR #1197, opened 2023-09-11
Detection model version upgrade to update License

gauravrajguru:


assets/models/system/deformable_detr_twostage_refine_r50_16x2_50e_coco/spec.yaml
```diff
@@ -19,4 +19,4 @@ tags:
     apply_deepspeed: 'false'
     apply_ort: 'false'
   task: object-detection
-version: 4
+version: 5
```
rjaincc, line 22: Version bump is not required. We have not released version 4 of these models yet.

rjaincc (approved):
We don't need this version bump for OD models. 
Only for IS.

rjaincc (approved):

haritha-bendapudi (approved):

PR #1207, opened 2023-09-11
Support copy of changed files

novaturient95:


skasturi (approved):

elesel:
Looks good! Please also bump the azureml-assets package version and update the CHANGELOG so that this change can get into PyPI.

novaturient95:
> Looks good! Please also bump the azureml-assets package version and update the CHANGELOG so that this change can get into PyPI.

Done

elesel (approved):

PR #1213, opened 2023-09-12
[Model Monitoring] add violations samples

hawestra:
Add GSQ violations as samples output

https://ml.azure.com/experiments/id/1a0a9a51-a26e-44b3-92d2-dcfd2d1fe6aa/runs/sharp_street_67c10lhdwn?wsid=/subscriptions/e0fd569c-e34a-4249-8c24-e8d723c7f054/resourceGroups/hawestra-rg/providers/Microsoft.MachineLearningServices/workspaces/hawestra-ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

### DIFF
--- a/assets/model_monitoring/components/generation_safety_quality/annotation_compute_histogram/spec.yaml
+++ b/assets/model_monitoring/components/generation_safety_quality/annotation_compute_histogram/spec.yaml
@@ -4,7 +4,7 @@ type: spark
 name: gsq_annotation_compute_histogram
 display_name: Annotation - Compute Histogram
 description: Compute annotation histogram given a deployment's model data input.
-version: 0.3.7
+version: 0.3.8
 is_deterministic: false
 inputs:
   production_dataset: 
@@ -66,6 +66,24 @@ outputs:
   histogram:
     type: mltable
     mode: direct
+  samples_index:
+    type: mltable
+    mode: direct
+  groundedness_violations:
+    type: mltable
+    mode: direct
+  fluency_violations:
+    type: mltable
+    mode: direct
+  coherence_violations:
+    type: mltable
+    mode: direct
+  similarity_violations:
+    type: mltable
+    mode: direct
+  relevance_violations:
+    type: mltable
+    mode: direct
 
 conf:
   spark.driver.cores: 1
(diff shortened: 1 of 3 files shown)


assets/model_monitoring/components/generation_safety_quality/annotation_compute_histogram/spec.yaml
```diff
@@ -66,6 +66,24 @@ outputs:
   histogram:
     type: mltable
     mode: direct
+  samples_index:
```
romanlutz, line 69: With this kind of input signature change perhaps it's time for v0.4.0?

assets/model_monitoring/components/generation_safety_quality/annotation_compute_histogram/spec.yaml
```diff
@@ -66,6 +66,24 @@ outputs:
...
+    type: mltable
+    mode: direct
+  groundedness_violations:
```
romanlutz, line 72: Did you consider just "violations" and then mark the kind of violation in the output data? Just wondering whether this kind of explosion of outputs makes sense for every possible metric.

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1285,11 +1291,20 @@ def _parse_responses(
             print(f"Failed decoding examples: {sample_examples}")
             num_failed += 1
 
+    # if the number of expected samples does not match the output samples
+    # there was an issue getting the response and we cannot accurately map
+    # the indices with the input data
+    index_mapping = job.prompt_data.input_idx
```
romanlutz, line 1297: Aren't those just 1-20, for example, for every batch? I might be misremembering.

hawestra: I've only tested on a dataset with 14 but there's two batches and I get 0-9 in one batch and 10-13 in the next. I can test with a larger dataset to confirm but seems like it's not

assets/model_monitoring/components/generation_safety_quality/annotation_compute_histogram/spec.yaml
hawestra: yeah right now it needs to be like this (ie one table for every metric) for the UI to consume

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
hawestra: just tested with a dataset that has 55 entries and there was an input_idx for each entry even though the batches were smaller than that

imatiach-msft (approved):

mastloui-msft (approved):

PR #1217, opened 2023-09-12
Check ACS Index Name in Validate Deployments

gdyre:


assets/large_language_models/rag/components/src/validate_deployments.py
```diff
@@ -466,6 +466,18 @@ def validate_acs(acs_config, activity_logger: Logger):
     connection_id_acs = os.environ.get(
         "AZUREML_WORKSPACE_CONNECTION_ID_ACS", None)
 
+    index_name = acs_config["index_name"]
```
gjwoods, line 469: Should we check the name is in the values?

gjwoods: I believe this is for Pythonese (?parseltongue?) for c# TryGetValue
```suggestion
    index_name = acs_config.get("index_name", None)
```

gjwoods (changes requested):

assets/large_language_models/rag/components/src/validate_deployments.py
gdyre: Good point. Will fix

gjwoods (approved):

PR #1221, opened 2023-09-13
Conditional_pipeline_component

vikasagrawal-ms:
Making import pipeline conditional. if a model already exists in the azureml or azureml-meta registry. We skip all downline components and pass the pipeline with required info in the logs


assets/training/model_management/src/run_model_download.py
```diff
@@ -99,18 +88,17 @@ def run():
     logger.info(f"Update existing model: {update_existing_model}")
 
     if update_existing_model == "false":
-        validate_if_model_exists(model_id)
+        if validate_if_model_exists(model_id):
+            return False
```
novaturient95, line 92: Should we have a false scenario to download model from registry and registration details and setting that to pipeline output?

vikasagrawal-ms: we would need a new component to do that.

PR #1250, opened 2023-09-18
Add dbcopilot_faiss_e2e and dbcopilot_acs_e2e

ricardrao:


assets/large_language_models/dbcopilot/components/llm_dbcopilot_create_promptflow/spec.yaml
```diff
@@ -4,7 +4,7 @@ tags:
   Preview: ""
 name: llm_dbcopilot_create_promptflow
 display_name: LLM - Create DBCopilot Prompt Flow
-version: 0.0.23
+version: 0.0.24
```
xia-xiao, line 7: why do you need to upgrade this component?

PR #1264, opened 2023-09-19
mlflow model conversion from HF for inpainting task

gauravrajguru:
test pipelines: https://ml.azure.com/experiments/id/310732e9-14db-4e9e-a1d7-d19233140c42?wsid=/subscriptions/dbd697c3-ef40-488f-83e6-5ad4dfb78f9b/resourcegroups/grajguru-rg/workspaces/grajguru-ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47


assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/text_to_image/utils.py
```diff
@@ -5,13 +5,40 @@
 
 import base64
 from io import BytesIO
+import io
```
rjaincc, line 8: Would be changing after Phani's PR?

gauravrajguru: yep, refactored it.

rjaincc (approved):

rjaincc:
@gauravrajguru - Can we have links to the pipeline runs with these changes?

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/convertors.py
```diff
@@ -314,7 +316,77 @@ def save_as_mlflow(self):
...
+        ]
+        super()._save(
+            mlflow_model_wrapper=mlflow_model_wrapper,
+            artifacts_dict=artifacts_dict,
+            conda_env=conda_env_file,
+            code_path=code_path,
+        )
+
+
+class TextToImageInpaintingMLflowConvertor(PyFuncMLFLowConvertor):
```
PhaniShekhar, line 329: Could you please explain why we need this additional class, why not StableDiffusionInpaintingMlflowConvertor inherit from PyFuncMLFLowConvertor

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/text_to_image/stable_diffusion_inpainting_mlflow_wrapper.py
```diff
@@ -0,0 +1,127 @@
...
+            f"But, found text prompt length {len(text_prompts)}, image length {len(images)} and "
+            f"mask_image length {len(mask_images)}"
+        )
+
+        generated_images = []
+        nsfw_content = []
+        if self._batch_output_folder:
+            # Batch endpoint
+            for image, mask_image, text_prompt in zip(images, mask_images, text_prompts):
+                output = self._pipe(
```
PhaniShekhar, line 94: Why not pass all the images together into self._pipe() similar to online endpoint and loop later on the outputs.images to save them to batch_output_folder ?

assets/training/model_management/src/azureml/model/mgmt/processors/factory.py
```diff
@@ -121,20 +122,29 @@ def create_mlflow_convertor(model_dir, output_dir, temp_dir, translate_params):
...
+
+    STABLE_DIFFUSION_TASK_MAP = {
+        PyFuncSupportedTasks.TEXT_TO_IMAGE.value: StableDiffusionMlflowConvertor,
+        PyFuncSupportedTasks.TEXT_TO_IMAGE_INPAINTING.value: StableDiffusionInpaintingMlflowConvertor,
+    }
 
     def create_mlflow_convertor(model_dir, output_dir, temp_dir, translate_params):
         """Create MLflow convertor for diffusers."""
         misc = translate_params["misc"]
         if misc and SupportedTextToImageModelFamily.STABLE_DIFFUSION.value in misc:
```
PhaniShekhar, line 136: Why do we need this model_family parameter in misc ?

PhaniShekhar (changes requested):

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/text_to_image/stable_diffusion_inpainting_mlflow_wrapper.py
MadhuM02: In case of batch, all the inputs are passed at once. it will run OOM if we invoke pipe with all the images. For online endpoint it's up to the user to send the optimal batch size.

PhaniShekhar: What is the user input for batch endpoints in inpainting case?
For CLIP, we are using csv files containing base64 image column and label column.
We are controlling the batch size by limiting the number of rows in a single csv file and passing a folder of csv files to the batch endpoint. 
Wondering if that would work here ?

rjaincc: @PhaniShekhar : 
- For batch endpoints the batch size is the #CSVs given as input, NOT #rows in the csv. Are you handling it differently for CLIP? 
- The primary reason for looping over images one by one is OOM issue as Madhu mentioned earlier.

PhaniShekhar: Yes. so effective batch size to the model would be (min_batch_size in batch input) * (# rows per csv).
In our example, we are using csv files with 10 rows each and min_batch_size of 2. We also called this out and added a comment to change these values in case of OOM.
This would give some sort of a control for the user on the batch size to the model right ?

MadhuM02: let's say if the max batch size for a particular deployment is 10, user has to divide his data into multiple csv's with 10 rows and put 1 as mini batch size. just wondering if its user friendly??

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/convertors.py
MadhuM02: we need to add this model https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0 ,which is a texttoimage model but instead of StableDiffusionPipeline, it uses DiffusionPipeline. we could inherit from the middle class. Note: this example for SD, not SD inpainting.

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/text_to_image/stable_diffusion_inpainting_mlflow_wrapper.py
MadhuM02: updated to pass all the images at one go for batch also. Will have to update the notebooks to call out the same to users.

assets/training/model_management/src/azureml/model/mgmt/processors/factory.py
MadhuM02: for handling cases like, Diffusion pipeline and SD pipeline both support TEXT_TO_IMAGE Tasks. to differentiate between different model_families for the same task

PhaniShekhar (approved):

rjaincc (approved):

PR #1283, opened 2023-09-21
Update DNNVisionGPU Env in azureml-assets

shubhamiit:
Updating DNN Vision env in azureml-assets to include the latest changes.

### DIFF
--- a/assets/training/vision/environments/automl-dnn-vision-gpu/context/Dockerfile
+++ b/assets/training/vision/environments/automl-dnn-vision-gpu/context/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/azureml/openmpi4.1.0-cuda11.1-cudnn8-ubuntu20.04:{{latest-image-tag}}
+FROM mcr.microsoft.com/azureml/openmpi4.1.0-cuda11.6-cudnn8-ubuntu20.04:{{latest-image-tag}}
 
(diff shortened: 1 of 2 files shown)


rjaincc (approved):

rjaincc:
@shubhamiit - Let's please update component versions as well in the same PR.

yiyoulin:
do we have a successful run associated with this? the reason why I'm asking is that, simply upgrading the torch will break our code in yolo. so in my PR: https://msdata.visualstudio.com/Vienna/_git/AzureMlCli/pullrequest/1145455 I have some edits in utils.py. 

would love to know if in this PR it's also needed, if not, why. :) thanks!

PhaniShekhar:
> do we have a successful run associated with this? the reason why I'm asking is that, simply upgrading the torch will break our code in yolo. so in my PR: https://msdata.visualstudio.com/Vienna/_git/AzureMlCli/pullrequest/1145455 I have some edits in utils.py.
> 
> would love to know if in this PR it's also needed, if not, why. :) thanks!

Agree with Yiyou's comments. If we have edits in package (dnn-vision or other packages), they have to be released to pypi before we can release this environment.
When we release this environment, it picks up the latest pypi azureml packages. If some required changes (like Yiyou mentioned) are not present in the current pypi version, it could break scenarios.

PhaniShekhar (changes requested):
Please wait for azureml package release.

shubhamiit:
Abandoning this PR as it will not work current pypi package as per team's comment

PR #1290, opened 2023-09-21
add prompt sanitization for dsmii chat models in scoring script

srsaggam:
Throw error when user inputs special tags for Llama chat models.

### DIFF
--- a/assets/training/model_management/environments/foundation-model-inference/context/score.py
+++ b/assets/training/model_management/environments/foundation-model-inference/context/score.py
@@ -579,6 +579,7 @@ def get_request_data(
             })
         )
 
+
 def build_chat_completion_prompt(data: List[str]) -> dict:
     """
     Build a chat completion prompt based on a list of conversation data.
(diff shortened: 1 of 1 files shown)


msp8955 (dismissed):

assets/training/model_management/environments/foundation-model-inference/context/score.py
```diff
@@ -626,6 +636,11 @@ def build_chat_completion_prompt(data: List[str]) -> dict:
     return result
 
 
+def _sanitize_content(content: str) -> None:
```
svaruag, line 639: nit: the function is named like it should return a sanitized string.. how about add the strip part here, and let it return that (and raise if token in content like you have?)

svaruag (dismissed):

assets/training/model_management/environments/foundation-model-inference/context/score.py
srsaggam: renamed function to reflect the logic better

svaruag (approved):

msp8955 (approved):

esrao (approved):

PR #1294, opened 2023-09-22
Bhavanatumma/ta2614748

BhavanaTumma:
### Description
Add mlflow transformers conversion support for NLP models

### Test

Validations: [page](https://microsoftapc-my.sharepoint.com/:fl:/g/personal/bhavanatumma_microsoft_com/ER0xlMWwtYxOgGgnPV0oDa0B-UITZAY8VTfxJ-PPGRo-PQ?e=1ATsBI&nav=cz0lMkZwZXJzb25hbCUyRmJoYXZhbmF0dW1tYV9taWNyb3NvZnRfY29tJmQ9YiUyMWVqRkxWNFZtZDBTMGZzZGxwa0pWbHkwTDVrbmNyaUZCbnMwZGctblFRM05OVzk0Ym1LQ2xRWkVpdkdTWUJnVkgmZj0wMUZCT1RQWEk1R0dLTUxNRlZSUkhJQTJCSEhWT1NRRE5OJmM9JTJGJmE9TG9vcEFwcCZwPSU0MG1zJTJGb2ZmaWNlLWZsdWlkLWNvbnRhaW5lcg%3D%3D)


assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -193,29 +196,58 @@ def _save(
...
+            pip_pkgs = self._get_curated_environment_pip_package_list()
+            conda_deps = CondaDependencies.create(conda_packages=None,
+                                                  python_version=python_version,
+                                                  pip_packages=pip_pkgs,
+                                                  pin_sdk_version=False)
+            curated_conda_env = conda_deps.as_dict()
+
+            model_pipeline = transformers.pipeline(task=self._task, model=model)
+
+            mlflow.transformers.save_model(
```
novaturient95, line 211: what if this fails with OOM? Are we expecting customers to provide compute sufficient enough to load the model? Should we fail before in this scenario?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -412,4 +480,4 @@ def save_as_mlflow(self):
         hf_conf[HF_CONF.HF_PRETRAINED_CLASS.value] = self._hf_model_cls.__name__
         hf_conf[HF_CONF.HF_TOKENIZER_CLASS.value] = self._hf_tokenizer_cls.__name__
```
BhavanaTumma, line 482: Remove segregate false and move to a different method for hf transformers

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: Maybe we can handle this in the validation component?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -253,6 +285,42 @@ def _update_conda_dependencies(self, package_details):
             yaml.safe_dump(conda_dict, f)
             logger.info("updated conda.yaml")
 
+    def _get_curated_environment_pip_package_list(self) -> List[str]:
```
novaturient95, line 288: Could you please share context on this?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: loading with device_map="auto"

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: The base environment is using a pkg called apex, whose version is not in pypi, hence when we let mlflow create the conda file from the base env, it picks up apex for certain models and fails the local validation as it is unable to find the package in pypi. We know exactly what packages are needed and which ones need to be pinned for version, this method creates the conda yaml file that we require and that is sent to mlflow.transformers.save_model()

assets/common/components/mlflow_model_local_validation/spec.yaml
```diff
@@ -17,6 +17,7 @@ command: >-
   --model-path ${{inputs.model_path}}
   $[[--test-data-path ${{inputs.test_data_path}}]]
   $[[--column-rename-map "${{inputs.column_rename_map}}"]]
+  $[[--task-name ${{inputs.task_name}}]]
```
novaturient95, line 20: could we read task_type from MLmodel? both OSS and hftrasnsformers v2 has task emitted to MLModel to be able to load in future

assets/common/src/scripts/validate_model.py
```diff
@@ -91,12 +93,15 @@ def _load_and_prepare_data(test_data_path: Path, mlmodel: Dict, col_rename_map:
...
 
-def _load_and_infer_model(model_dir, data):
+def _load_and_infer_model(model_dir, data, task_name):
     if data is None:
         print("Data not shared. Could not infer the loaded model")
         return
 
     try:
+        if task_name == SUMMARIZATION:
+            os.environ["MLFLOW_HUGGINGFACE_USE_DEVICE_MAP"] = "False"
```
novaturient95, line 103: this is specific to oss transformers? Please add a check to distinguish b/w hftransformers and oss transformers?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -175,7 +177,65 @@ def _save(
...
+            pip_pkgs = self._get_curated_environment_pip_package_list()
+            conda_deps = CondaDependencies.create(conda_packages=None,
+                                                  python_version=python_version,
+                                                  pip_packages=pip_pkgs,
+                                                  pin_sdk_version=False)
+
+            curated_conda_env = conda_deps.as_dict()
+
+            # handle OSS trust_remote_code value
+            trust_remote_code_val = False
```
novaturient95, line 197: can we make use of hf_pipeline_args?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -175,7 +177,65 @@ def _save(
...
+                    and model_args.get('trust_remote_code', False)):
+                trust_remote_code_val = True
+
+            # handle OSS model loading with device map
+            try:
+                model_pipeline = transformers.pipeline(task=self._task, model=model, device_map="auto",
+                                                       trust_remote_code=trust_remote_code_val)
+                logger.info("OSS model loaded with device_map auto")
+            except Exception as e:
+                logger.error("OSS Model load failed with exception: {}, reloading without device_map auto".format(e))
```
novaturient95, line 213: can you add a comment why this exception is needed? how not passing device_map="auto" plays etc and what scenario are we enabling from here. Is there a way to know that error would happen prior to calling transformers.pipeline with `device_map="auto"` and not get into exception at all?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -175,7 +177,65 @@ def _save(
...
+                model_pipeline = transformers.pipeline(task=self._task, model=model, trust_remote_code=trust_remote_code_val)
+
+                # pass in signature for a text-classification model
+            if self._task == SupportedNLPTasks.TEXT_CLASSIFICATION.value:
+                inputs = Schema([ColSpec(DataType.string)])
+                outputs = None
+                params = ParamSchema([ParamSpec("return_all_scores", "boolean", default=True)])
+                self._signatures = ModelSignature(inputs=inputs, outputs=outputs, params=params)
```
novaturient95, line 221: handle in derived class?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -190,39 +250,36 @@ def _save(
                 self._model_dir, **self._hf_conf.get(HF_CONF.HF_TOKENIZER_ARGS.value, {})
             )
 
-        # Set experimental flag
-        if self._experimental:
-            logger.info("Experimental features enabled for MLflow conversion")
```
novaturient95, line 195: removing experimental?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -190,39 +250,36 @@ def _save(
...
-            self._hf_conf["exp"] = True
+            # set metadata info
+            metadata = fetch_mlflow_acft_metadata(base_model_name=self._model_id,
+                                                  is_finetuned_model=False)
```
novaturient95, line 255: this is needed for both OSS and HFTransformers flavor. Could you please check with @rjaincc?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -260,6 +317,42 @@ def _update_conda_dependencies(self, package_details):
...
+    def _get_curated_environment_pip_package_list(self) -> List[str]:
+        """
+        Retrieve the packages using 'conda list' command.
+
+        :return: A List of the pip package and the corresponding versions.
+        """
+        import subprocess
+        import json
+
+        PIP_LIST = ['mlflow', 'accelerate', 'cffi', 'dill', 'google-api-core', 'numpy',
```
novaturient95, line 329: is this an exhaustive list?

novaturient95, line 329: could we maintain a config like `conda.yaml` instead?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
novaturient95: using device_map="auto" can also lead to OOM in cases when combined memory is short for a model? Please let me know if this not correct. 

Don't think this is a blocker per se, but we need to handle this and make sure we throw a safe exception here

assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -28,6 +28,15 @@ inputs:
     description: Huggingface model id (https://huggingface.co/<model_id>). A required parameter for Huggingface model framework. Can be provided as input here or in model_download_metadata JSON file.
     optional: true
 
+  model_flavor:
```
novaturient95, line 31: can we handle it internally? We can make OSS default and if it fails use hftransformers as we've discussed previously?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -187,6 +174,85 @@ def _save(
         metadata = fetch_mlflow_acft_metadata(base_model_name=self._model_id,
                                               is_finetuned_model=False,
                                               base_model_task=self._task)
+        if self._model_flavor == "OSS":
```
novaturient95, line 177: we can keep common code in _save and move both OSS model save handling in a separate func as you've done hftv2

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -249,6 +313,47 @@ def _update_conda_dependencies(self, package_details):
...
+        import subprocess
+        import json
+
+        PIP_LIST = ['accelerate', 'cffi', 'dill', 'google-api-core', 'numpy',
+                    'packaging', 'pillow', 'protobuf', 'pyyaml', 'requests', 'scikit-learn',
+                    'scipy', 'sentencepiece', 'torch', 'mlflow']
+        ADD_PACKAGE_LIST = ['torchvision==0.14.1', 'transformers==4.34.1']
+
+        if self._task == SupportedTasks.AUTOMATIC_SPEECH_RECOGNITION.value:
+            ADD_PACKAGE_LIST.append(['soundfile==0.12.1', 'librosa==0.10.1', 'diffusers==0.21.4',
```
novaturient95, line 331: There is a conda.yaml for whisper model, since it has hardcoded version, we dont wanna use it?

assets/common/components/mlflow_model_local_validation/spec.yaml
skasturi: @novaturient95 What about models that support multiple tasks?

novaturient95: @skasturi for that scenario as well, we can or should infer from MLmodel especially if its getting logged in case of OSS transformers as well as hftv2

assets/common/components/mlflow_model_local_validation/spec.yaml
```diff
@@ -35,7 +36,24 @@ inputs:
     optional: true
     description: |
       Provide mapping of dataset column names that should be renamed before inferencing.
-      eg: col1:ren1; col2:ren2; col3:ren3 
+      eg: col1:ren1; col2:ren2; col3:ren3
+
+  task_name:
+    description: A Hugging face task on which model was trained on
+    enum:
+      - text-classification
```
vikasagrawal-ms, line 44: what about vision model releated tasks
image-object-detection and image-instance-segmentation

vikasagrawal-ms (commented):
Check for image tasks

assets/common/components/mlflow_model_local_validation/spec.yaml
vikasagrawal-ms: got it added in ES.

vikasagrawal-ms (approved):

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
novaturient95: Lets move to specific params changes within specific task classes. In this case: NLPMLflowConvertor

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
novaturient95: have we went into exception for any model or on any compute per se so far with all models have imported

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
novaturient95: we have config, model, tokenizer and pipeline args right? that is meant to be used for passing any init params.

novaturient95: looks like there is trust remote code handling here: https://github.com/Azure/azureml-assets/blob/2fee96555655df5e4ae2c18d1b7f1b839d6e8198/assets/training/model_management/src/run_model_preprocess.py#L144C1-L155C108


Could you please check if change here is also needed?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
novaturient95: we can device_map="auto". I was saying that even with that, there is a chance that model would not be loaded. How are we dealing with that scenario? We should fail it safely by estimating model memory req and what is available

HrishikeshGeedMS (dismissed):

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: The code mentioned above is setting the values, whereas the code added in the PR is checking that all the values if set are provided to the transformers pipeline

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: approach removed as it was causing issues

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: Did that for text-gen, text classification is automated

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -187,6 +174,84 @@ def _save(
...
+                                                      requirements_file, pip_requirements)
+        else:
+            self._save_in_hftransformersv2_flavor(metadata, conda_env,
+                                                  code_paths, input_example,
+                                                  requirements_file, pip_requirements)
+
+        # pin pycocotools==2.0.4
+        self._update_conda_dependencies({"pycocotools": "2.0.4"})
+
+        if self._task == SupportedTasks.AUTOMATIC_SPEECH_RECOGNITION.value:
```
novaturient95, line 195: Could we provide addn conda dependencies from respective Task classes?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -410,6 +515,16 @@ def save_as_mlflow(self):
...
 
+        if self._task == SupportedNLPTasks.TEXT_GENERATION.value:
+            inputs = Schema([ColSpec(DataType.string)])
+            outputs = Schema([ColSpec(DataType.string)])
+            params = ParamSchema([ParamSpec("top_p", "float", default=1.0),
+                                  ParamSpec("temperature", "float", default=1.0),
+                                  ParamSpec("max_new_tokens", "integer", default=50),
+                                  ParamSpec("do_sample", "boolean", default=True),
+                                  ParamSpec("return_full_text", "boolean", default=True)])
+            self._signatures = ModelSignature(inputs=inputs, outputs=outputs, params=params)
```
novaturient95, line 526: there is a specific method to get ModelSignature. Could you please move it there

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -410,6 +515,16 @@ def save_as_mlflow(self):
         self._hf_config_cls = self._hf_config_cls if self._hf_config_cls else AutoConfig
         self._hf_tokenizer_cls = self._hf_tokenizer_cls if self._hf_tokenizer_cls else AutoTokenizer
 
+        if self._task == SupportedNLPTasks.TEXT_GENERATION.value:
```
novaturient95, line 518: Add tests for these?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -249,6 +313,47 @@ def _update_conda_dependencies(self, package_details):
...
+        output_str = process.stdout.decode('ascii')
+        output_json = json.loads(output_str)
+        pip_list = []
+        for pkg in output_json:
+            pkg_name = pkg['name']
+            pkg_version = pkg['version']
+            if pkg_name in PIP_LIST:
+                pip_list.append(pkg_name + "==" + pkg_version)
+
+        for pkg in ADD_PACKAGE_LIST:
```
novaturient95, line 351: nit: can be simplified to `pip_list += ADD_PACKAGE_LIST` or better `pip_list.extend(ADD_PACKAGE_LIST)`

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -249,6 +313,47 @@ def _update_conda_dependencies(self, package_details):
...
+                                     'jsonpickle==3.0.1', 'more-itertools==9.1.0', 'cryptography==41.0.1'])
+
+        conda_list_cmd = ["conda", "list", "--json"]
+        try:
+            process = subprocess.run(conda_list_cmd, shell=False, check=True,
+                                     stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        except (FileNotFoundError, subprocess.CalledProcessError) as err:
+            logger.warning('subprocess failed to get dependencies list from conda with error: {}'.format(err))
+            return []
+        output_str = process.stdout.decode('ascii')
```
novaturient95, line 342: catch exception?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -249,6 +313,47 @@ def _update_conda_dependencies(self, package_details):
...
+
+        :return: A List of the pip package and the corresponding versions.
+        """
+        import subprocess
+        import json
+
+        PIP_LIST = ['accelerate', 'cffi', 'dill', 'google-api-core', 'numpy',
+                    'packaging', 'pillow', 'protobuf', 'pyyaml', 'requests', 'scikit-learn',
+                    'scipy', 'sentencepiece', 'torch', 'mlflow']
+        ADD_PACKAGE_LIST = ['torchvision==0.14.1', 'transformers==4.34.1']
```
novaturient95, line 328: why hardcode?

novaturient95: we would want to make use of model-mgmt env to get he dependencies in all case.

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -187,6 +174,84 @@ def _save(
...
+        self._update_conda_dependencies({"pycocotools": "2.0.4"})
+
+        if self._task == SupportedTasks.AUTOMATIC_SPEECH_RECOGNITION.value:
+            self._update_conda_dependencies({"ffmpeg": "4.2.2"})
+
+    def _save_in_oss_flavor(self, model, code_paths, input_example, pip_requirements, metadata):
+        # create a conda environment for OSS transformers Flavor
+        python_version = platform.python_version()
+        pip_pkgs = self._get_curated_environment_pip_package_list()
+        conda_deps = CondaDependencies.create(conda_packages=None,
```
novaturient95, line 202: from what I understand, the reason for creating new conda env here with specific dependencies is to load the model for saving to oss first and then persisting dependencies in conda.yaml. But I see this env being created, and model still being loaded into current component env (model-mgmt). If this is working in all case, would it make sense to not create conda env at all here and just provide dict as part of model save to oss?

novaturient95: I am also wondering how loading the model in current env works but when conda is inferred by default for oss it adds some internal dependencies which break future valdn. 

Was this verified again recently if that is still the case?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -187,6 +174,84 @@ def _save(
...
+
+                self._save_in_hftransformersv2_flavor(metadata, conda_env,
+                                                      code_paths, input_example,
+                                                      requirements_file, pip_requirements)
+        else:
+            self._save_in_hftransformersv2_flavor(metadata, conda_env,
+                                                  code_paths, input_example,
+                                                  requirements_file, pip_requirements)
+
+        # pin pycocotools==2.0.4
```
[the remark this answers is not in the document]
BhavanaTumma: yes, we require it

assets/common/components/mlflow_model_local_validation/spec.yaml
```diff
@@ -1,7 +1,7 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 
 name: mlflow_model_local_validation
-version: 0.0.6
+version: 0.0.7
```
novaturient95, line 4: nit: move version change out of this PR, so you dont always have to sync with latest and also reduces any error in versioning

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -2,12 +2,15 @@
...
-
+from typing import List
 import transformers
+import platform
+import mlflow
 import os
 import yaml
 from abc import ABC, abstractmethod
 from azureml.evaluate import mlflow as hf_mlflow
+from azureml.core.conda_dependencies import CondaDependencies
```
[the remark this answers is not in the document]
BhavanaTumma: Will be handling this in another PR

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: Moved the pkgs to the conda yaml for whisper, the others we had a conversion on the fact that the version is being taken from the latest env

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: Done

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -187,6 +174,84 @@ def _save(
...
+        tokenizer_args = self._hf_conf.get(HF_CONF.HF_TOKENIZER_ARGS.value, {})
+        model_args = self._hf_conf.get(HF_CONF.HF_MODEL_ARGS.value, {})
+
+        if (config_args.get('trust_remote_code', False)
+                and tokenizer_args.get('trust_remote_code', False)
+                and model_args.get('trust_remote_code', False)):
+            trust_remote_code_val = True
+
+        model_pipeline = transformers.pipeline(task=self._task, model=model,
+                                               trust_remote_code=trust_remote_code_val)
```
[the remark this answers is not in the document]
BhavanaTumma: Let's handle this in a different PR

vikasagrawal-ms (approved):

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -383,21 +495,17 @@ def __init__(self, **kwargs):
...
-                    ]
-                ),
-                outputs=Schema(inputs=[ColSpec(name="text", type=DataType.string)]),
-            )
-
-        return ModelSignature(
-            inputs=Schema(inputs=[ColSpec(name="input_string", type=DataType.string)]),
-            outputs=Schema(inputs=[ColSpec(name="text", type=DataType.string)]),
-        )
+        if self._task == SupportedNLPTasks.TEXT_GENERATION.value:
```
novaturient95, line 498: does text2textgeneration also need to have similar params?

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
```diff
@@ -160,33 +163,108 @@ def _save(
...
+        model = str(tmp_model_dir)
+        logger.info("Loading config")
+        config = self._hf_config_cls.from_pretrained(
+            self._model_dir, **self._hf_conf.get(HF_CONF.HF_CONFIG_ARGS.value, {})
+        )
+        logger.info("Loading tokenizer")
+        tokenizer = self._hf_tokenizer_cls.from_pretrained(
+            self._model_dir, **self._hf_conf.get(HF_CONF.HF_TOKENIZER_ARGS.value, {})
+        )
+
```
novaturient95, line 267: setting signature for hftv2 might be an issue. Could you please confirm once that its not set?

assets/training/model_management/src/run_model_preprocess.py
```diff
@@ -155,6 +157,7 @@ def run():
             logger.warning(f"{TRUST_CODE_KEY} is not provided for hf_config_args. Using {TRUST_CODE_KEY}.")
     preprocess_args["task"] = task_name.lower()
     preprocess_args["model_id"] = model_id if model_id else preprocess_args.get("model_id")
+    preprocess_args["model_flavor"] = model_flavor if model_flavor else "HFtransformersV2"
```
novaturient95, line 160: Please define "HFtransformersV2" under constants. Basically we can add all flavors that is supported

assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -28,6 +28,15 @@ inputs:
...
+    type: string
+    enum:
+      - HFTransformersV2
```
novaturient95, line 34: what about models that make use of pyfunc ex vision models. Could we also get approval from @rjaincc and team pls if they see any concern?

novaturient95 (approved):

novaturient95:
> Code to add OSS flavor to the import model component

would be super helpful to add validation info as part of PR. I understand we have done plenty but still good to have if we have to see the PR back again and confirm scenarios that were validated

assets/training/model_management/src/azureml/model/mgmt/processors/transformers/convertors.py
BhavanaTumma: Not that we know of at the moment

assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
novaturient95: This does not seem to be affecting vision flow. But it will be good to either change property name hf_mlflow_flavor or add pyfunc in the enum options as well.

assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -28,6 +28,15 @@ inputs:
...
+      - OSS
+    default: HFTransformersV2
+    optional: false
```
novaturient95, line 37: we should also think about making it optional if not including pyfunc as favor. For vision models this property would not be used

novaturient95 (dismissed):

novaturient95 (dismissed):

assets/training/model_management/src/run_model_preprocess.py
BhavanaTumma: Will handle in a different PR

vikasagrawal-ms (approved):

novaturient95 (approved):

PR #1308, opened 2023-09-25
[llm] Change to handle new expectation

gjwoods:


assets/large_language_models/rag/components/crack_and_chunk/spec.yaml
```diff
@@ -80,7 +80,7 @@ command: >-
   --input_data ${{inputs.input_data}}
   $[[--input_glob '${{inputs.input_glob}}']]
   $[[--allowed_extensions ${{inputs.allowed_extensions}}]]
-  --output_title_chunk ${{outputs.output_chunks}}
+  --output_chunks ${{outputs.output_chunks}}
```
prakharg-msft, line 83: It will also need changes to pipelines?

gjwoods: Looks like there is no reference
image

tot0 (approved):

PR #1314, opened 2023-09-26
Removing/renaming computes_allow_list  to finetune_compute_allow_list

rjaincc:
1. Removing computes_allow_list from the models where both computes_allow_list and finetune_compute_allow_list are present.
2. Renaming computes_allow_list to finetune_compute_allow_list, where only compute_allow_list is present.

**Q**. Why are we not removing finetune_compute_allow_list and keeping finitune-min-sku-spec?
**A**. The UI is picking up the finetune_compute_allow_list currently.


skasturi (approved):

haritha-bendapudi (approved):

assets/models/system/bert-large-cased/spec.yaml
```diff
@@ -24,28 +24,6 @@ properties:
   languages: en
 tags:
   Preview: ''
-  computes_allow_list:
```
amangupta26, line 27: Should this be renamed to finetune_compute_allow_list?

assets/models/system/bert-large-uncased/spec.yaml
```diff
@@ -24,28 +24,6 @@ properties:
   languages: en
 tags:
   Preview: ''
-  computes_allow_list:
```
amangupta26, line 27: Same here

assets/models/system/bert-large-cased/spec.yaml
rjaincc: We have ` finetune-min-sku-spec: 4|1|28|176` in properties. So we don't need to maintain he allow list.

assets/models/system/bert-large-uncased/spec.yaml
rjaincc: We have ` finetune-min-sku-spec: 4|1|28|176` in properties. So we don't need to maintain he allow list.

haritha-bendapudi (dismissed):

amangupta26 (dismissed):

skasturi (dismissed):

PR #1323, opened 2023-09-27
Update models for inference SKU list

vikasagrawal-ms:
inference-recommended-sku is a complete list of sku's supported for deployment. Any SKU which is not mentioned in the list will not be allowed for deployment. This will form the ground truth for Deployment validations. 
This list is created based on the inference-min-sku-spec already present in the model card.


BhavanaTumma (approved):
Approved.

assets/models/system/bert-base-cased/spec.yaml
```diff
@@ -10,7 +10,18 @@ properties:
   finetune-recommended-sku: Standard_NC24rs_v3
   finetuning-tasks: text-classification, token-classification, question-answering
   inference-min-sku-spec: 2|0|7|14
-  inference-recommended-sku: Standard_DS2_v2
+  inference-recommended-sku: Standard_DS2_v2, Standard_D2a_v4, Standard_D2as_v4, Standard_DS3_v2,
```
rjaincc, line 13: Why not add inference sku spec instead? 
This is a rather long list. Same at other such places.

PR #1327, opened 2023-09-27
Alexka/custom prompt asset

alexkalita:
Add 'prompt' as an asset type in config.py.  This will unblock testing to publish prompt assets to registries.


wayliums (approved):

elesel:
Please also update CHANGELOG and setup.py here, otherwise the change won't get to PyPI: https://github.com/Azure/azureml-assets/tree/main/scripts/azureml-assets

lykelly19 (approved):

PR #1331, opened 2023-09-28
adding invalid ds3 configurations

jpmann:
Adding invalid configurations with deepspeed stage3 -
1. Created a function to invalidate deepspeed stage3 invalid configurations. Added UTs to test the same.
2. Added a new function to calculate the deepspeed stage information. Added UTs for this func. 
3. Added a new function to validate the deepspeed stage3 config file. Currently, only 1 flag is validated whether it is set correctly or not, but in future we can add more checks to this. Added a UT to verify this functionality.


assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
```diff
@@ -442,6 +444,121 @@ def copy_preprocess_args(args: Namespace) -> Namespace:
...
+        dict(
+            invalid_settings=dict(apply_lora=True),
+            fail_run=True,
+            valid_settings=None
+        ),
+        dict(
+            invalid_settings=dict(apply_ort=True),
+            fail_run=True,
+            valid_settings=None
+        ),
```
skanakamedal, line 484: Like for auto_find_batch_size shouldn't we setting the values to false?

assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
```diff
@@ -442,6 +444,121 @@ def copy_preprocess_args(args: Namespace) -> Namespace:
...
+    :param args: User passed args
+    :type Namespace
+    :param do_validate: Validates the deepspeed config file in case of deepspeed stage3
+    :type bool
+    """
+    # Read the default deepspeed config if the apply_deepspeed is set to true without providing config file
+    args.deepspeed = getattr(args, "deepspeed", None) or DEFAULT_DEEPSPEED_CONFIG
+
+    # add validations for deepspeed stage3
+    if do_validate and identify_deepspeed_stage(args.deepspeed) == 3:
```
shivamMg, line 555: nit: deepspeed config can be opened in this setup func and then identify_deepspeed_stage & validate_ds_zero3_config can accept deepspeed_dict directly - avoid opening file twice

assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
jpmann: There are 2 types of validations one that fails the run and 1 that resets the value. _apply_ort_ and _apply_lora_ are set with fail_run=True i.e. unless these settings are enabled, we are going to fail the run. On the other hand, with _auto_find_batch_size=True_, fail_run=False. Here we need the valid configurations to reset the settings.

assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
```diff
@@ -442,6 +444,121 @@ def copy_preprocess_args(args: Namespace) -> Namespace:
...
+                    pii_safe_message=(
+                        "Invalid deepspeed config file. Stage information is missing from the config file."
+                    )
+                )
+            )
+
+    return ds_stage
+
+
+def check_for_invalid_ds_zero3_settings(args: Namespace):
```
amangupta26, line 467: Add log lines appropriately

assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
```diff
@@ -442,6 +444,121 @@ def copy_preprocess_args(args: Namespace) -> Namespace:
...
+                ACFTUserError,
+                pii_safe_message=(
+                    "stage3_gather_16bit_weights_on_model_save should be "
+                    "`true` in deepspeed stage 3 config."
+                )
+            )
+        )
+
+
+def setup_and_validate_deepspeed(args: Namespace, do_validate: bool = True):
```
amangupta26, line 543: Add log lines appropriately

skanakamedal (dismissed):

shivamMg (approved):

amangupta26 (dismissed):

skanakamedal (approved):

PR #1332, opened 2023-09-28
Revert "Revert "Added Llama model folders from azureml-asset repo (#1322)""

vikasagrawal-ms:
Removed the llama models from azureml-asset repo. Duplicate error will not come.


BhavanaTumma (approved):
Approved.

PR #1348, opened 2023-09-29
update mapping to handle more tokens

msp8955:


assets/models/system/Llama-2-13b-chat/spec.yaml
```diff
@@ -23,8 +23,8 @@ tags:
   - Standard_ND96asr_v4
   - Standard_ND96amsr_A100_v4
   sku_to_num_replicas_map:
-    Standard_NC24s_v3: 2
-    Standard_ND40rs_v2: 8
+    Standard_NC24s_v3: 1
+    Standard_ND40rs_v2: 4
```
svaruag, line 27: Will this automatically enable tensor parallelism of 2 (across 4 replicas)?

msp8955: yep, the calculation is in the score script
https://github.com/Azure/azureml-assets/blob/main/assets/training/model_management/environments/foundation-model-inference/context/score.py#L55

svaruag (approved):

PR #1353, opened 2023-10-02
Add OpenAI Support to components

gdyre:


assets/large_language_models/rag/components/src/flow_creation.py
```diff
@@ -135,10 +136,15 @@ def get_connection_name(s):
 
 def get_deployment_and_model_name(s):
     """get_deployment_and_model_name."""
-    deployment_and_model = s.split("/deployment/")[1]
-    deployment_name = deployment_and_model.split("/model/")[0]
-    model_name = deployment_and_model.split("/model/")[1]
-    return (deployment_name, model_name)
+    model_type = s.split("://")[0]
+    if model_type == "open_ai":
```
thomasshupeng, line 140: Are we sure the input is always string, and it consists lower case letters only? "Open_AI://..." is not a possible case?

nanditamohan (approved):

assets/large_language_models/rag/components/src/flow_creation.py
```diff
@@ -135,10 +136,15 @@ def get_connection_name(s):
 
 def get_deployment_and_model_name(s):
     """get_deployment_and_model_name."""
-    deployment_and_model = s.split("/deployment/")[1]
-    deployment_name = deployment_and_model.split("/model/")[0]
-    model_name = deployment_and_model.split("/model/")[1]
-    return (deployment_name, model_name)
+    model_type = s.split("://")[0]
```
gjwoods, line 139: ⁠I wonder if this should be helper function in a util folder (or azureml-rag) and used across multiple components.
FYI- side note- do other components support this?

gdyre: I agree this could work well as a helper function in azureml-rag. The embeddings and QA Gen component do support open ai models, I made the changes a while back

PR #1384, opened 2023-10-05
Mitigate issue with long prompts for LLaVA 7B on V100

rdondera-microsoft:
On V100 machines, long LLaVA prompts may cause GPU OOM errors which the endpoint/server does not recover from. This PR imposes an upper bound on the number of prompt tokens and throws a `ValueError` exception for longer prompts instead of doing inference on them. This effectively upper bounds the number of questions in a dialog to 5-10 (actual number dependent on question and answer length), but ensures that dialogs do not break the server.
On A100 machines, the memory consumption is normally around 22GB for the 13B model, so no issues.

Tested by
1. Running an endpoint with modified code and sending it questions until it raises the `ValueError` exception. (V100, 7B)
2. Running the LLaVA cli client on an A100 machine.

Screenshots with details for the above are in private chat.

### DIFF
--- a/assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/llava/llava_mlflow_wrapper.py
+++ b/assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/llava/llava_mlflow_wrapper.py
@@ -160,11 +160,15 @@ class LLaVAMLflowWrapper(mlflow.pyfunc.PythonModel):
             ).unsqueeze(0).cuda()
             stopping_criteria = KeywordsStoppingCriteria([self._stop_str], self._tokenizer, input_ids)
 
-            # Long prompts cause a GPU OOMs which the server does not recover from. To prevent this, we are using a
-            # length threshold that allows for a small number of question-answer pairs (e.g. 5-10) in each prompt.
-            prompt_length = max([len(i) for i in input_ids])
-            if prompt_length > MAX_PROMPT_LENGTH:
-                raise ValueError(f"Prompt too long: {prompt_length} tokens. Maximum allowed is {MAX_PROMPT_LENGTH}.")
+            # For small models on V100 machines, long prompts cause a GPU OOMs which the server does not recover from.
+            # To prevent this, we are using a length threshold that allows for a small number of question-answer pairs
+            # (e.g. 5-10) in each prompt.
+            if self._model_version in [self.LLAVA_MPT, self.LLAVA_7B]:
+                prompt_length = max([len(i) for i in input_ids])
+                if prompt_length > MAX_PROMPT_LENGTH:
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/llava/llava_mlflow_wrapper.py
```diff
@@ -160,6 +160,16 @@ def predict(self, context: mlflow.pyfunc.PythonModelContext, input_data: pd.Data
             ).unsqueeze(0).cuda()
             stopping_criteria = KeywordsStoppingCriteria([self._stop_str], self._tokenizer, input_ids)
 
+            # For small models on V100 machines, long prompts cause a GPU OOMs which the server does not recover from.
+            # To prevent this, we are using a length threshold that allows for a small number of question-answer pairs
+            # (e.g. 5-10) in each prompt.
+            if self._model_version in [self.LLAVA_MPT, self.LLAVA_7B]:
```
bhimar, line 166: We are assuming users will use V100 for these smaller models? If they use A100 for some reason then we will fail unnecessarily at MAX_PROMPT_LENGTH

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/llava/llava_mlflow_wrapper.py
```diff
@@ -160,6 +160,16 @@ def predict(self, context: mlflow.pyfunc.PythonModelContext, input_data: pd.Data
...
+                prompt_length = max([len(i) for i in input_ids])
+                if prompt_length > MAX_PROMPT_LENGTH:
+                    raise ValueError(
+                        f"Prompt too long: {prompt_length} tokens. Maximum allowed is {MAX_PROMPT_LENGTH}."
```
bhimar, line 170: The user doesn't really have any notion of "Prompt". Should we instead say "Dialogue" or "Conversation"?

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/llava/llava_mlflow_wrapper.py
rdondera-microsoft: Yes, that's the assumption. Correct observation regarding the small model on A100, but why would users do that? They can simply run the large model, which is clearly better than the small one.

assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/llava/llava_mlflow_wrapper.py
rdondera-microsoft: They do know what a prompt is. If not general knowledge, then they can read the model card. Speaking of which, turns out it **is** important to update the model card as per bug 2730582.

bhimar (approved):

PhaniShekhar (approved):

PR #1391, opened 2023-10-05
Update model version

novaturient95:
1. Update model version for published codellama and falcon models as previous version was not properly registered.
2. Recent change in model publishing script caused the models files to surface at the top level instead of mlflow_model_folder

side effect PR: #1341


BhavanaTumma (approved):
Approved.

PR #1395, opened 2023-10-05
Post processor component converged

Nivi09:
Added revised components and required functionalities in post-processor component to get shared as package.


assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+                text = self.remove_prefix(text, prefix)
+        return text
+
+    def apply_strip_characters(self, text: str) -> str:
+        """Remove set of characters from the begining and end the given text."""
+        if self.strip_characters:
+            text = text.strip(self.strip_characters)
+        return text
+
+    def apply_label_map(self, data) -> Union[pd.DataFrame, str]:
```
alpatots, line 274: Nit: type hint for `data`

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+        mlflow.log_metric("missing_completions", self.num_missing_completions)
+        mlflow.log_metric("raw_completion_length_avg", self.raw_completion_length / self.steps)
+        mlflow.log_metric("extracted_completion_length_avg", self.extracted_completion_length / self.steps)
+        mlflow.log_metric("total_completions", self.steps)
+
+        if self.num_missing_completions > 0:
+            logger.warning("For some lines no completions were received")
+
+
+class BabelInferencePostProcessor(InferencePostprocessor):
```
alpatots, line 485: Let's remove this from azureml-assets

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+        """Extract inferences using generic method if no template or custom post-processor is provided."""
+        predicted_data = read_jsonl_files(resolve_io_path(self.prediction_dataset))
+        pred_list = []
+        if self.prediction_column_name in predicted_data[0].keys():
+            key = self.prediction_column_name
+        else:
+            key = key if key else "0"
+        for row in predicted_data:
+            predicted = row.get(key)
+            if isinstance(predicted, list) and len(predicted[0]) > 1:
```
alpatots, line 378: This logic of "if predicted is a list vs. a string" can be extracted into a separate small method (and hence unit tested separately, too)

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+                curr_pred_list = []
+                for i in range(0, len(predicted)):
+                    out_string = predicted[i]
+                    out_string = self.apply_generic_processor(out_string, row)
+                    curr_pred_list.append(out_string)
+                pred_list.append(curr_pred_list)
```
alpatots, line 384: This is a nit, but to avoid code copy/paste you can do 

```python
if isinstance(predicted, str)
    predicted = [predicted] # if not a list, make a list with one element
curr_pred_list = [self.apply_generic_processor(out_string, row) for out_string in predicted] # always a list at this point
```

alpatots: Otherwise it's a bit too many if/else conditions.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+        out_string = self.apply_regex_expr(out_string)
+        out_string = self.apply_strip_characters(out_string)
+        out_string = self.apply_label_map(out_string)
+        return out_string
+
+    def extract_inferences(
+        self, result_df: pd.DataFrame, key: str = None, processor_order: List = None
+    ) -> pd.DataFrame:
+        """Extract inferences using generic method if no template or custom post-processor is provided."""
+        predicted_data = read_jsonl_files(resolve_io_path(self.prediction_dataset))
```
alpatots, line 370: This is small, but I would prefer all the file reading to be done in a separate method, and have a small method that just does the loop of extraction. It's a bit strange for a method to receive one dataset as  already loaded and the other to read on the fly.

assets/aml-benchmark/components/src/inference_postprocessor/runner.py
```diff
@@ -0,0 +1,240 @@
+# ---------------------------------------------------------
```
alpatots, line 1: This probably does not belong to this repo either.

assets/aml-benchmark/components/src/my_history.txt
```diff
@@ -0,0 +1,5213 @@
+ 1/1: import pandas
```
alpatots, line 1: I don't think this `my_history.txt` should be checked in? It looks like a log from code running. 

If it's something your code normally generates, perhaps adding to gitignore is needed.

user_script.json
alpatots, line 1: Why is this file needed?

alpatots (changes requested):

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
```diff
@@ -42,36 +48,56 @@ inputs:
...
+      value needs mapping, followed by key-value pairs containing idtolabel or labeltoid mappers.
+      Example format:
+      {"column_name":"label", "0":"NEUTRAL", "1":"ENTAILMENT", "2":"CONTRADICTION"}
     optional: True
   template:
     type: string
     description: >-
-      Jinja template containing logic to extract prediction.
+      Jinja template containing logic to extract prediction. In case of multiple predictions, logic must be written in a written in
+      format so that it outputs a list of formatted predictions.
```
arun-rajora, line 100: Can we provide a very basic example for it in the description?

assets/aml-benchmark/components/src/inference_postprocessor.zip
arun-rajora, line 1: Important- Seems like this zip file got added accidentally. If so, please remove it. Thanks!

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
-        result_df.to_json(self.result, lines=True, orient='records')
+            if key != self.prediction_column_name:
+                row[self.prediction_column_name] = row.get(key)
+            predicted = row.get(self.prediction_column_name)
+            if isinstance(predicted, list):
+                try:
+                    out_string = jinja_template.render(predicted)
+                    pred_list.append(out_string)
+                except Exception as e:
+                    # "jinja2.exceptions.UndefinedError: 'list object' has no attribute 'split'"
```
arun-rajora, line 327: When is this exception thrown? What is the scenario? Also, can you update the code comment to provide a brief context for it as well? Thanks!

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+    def extract_inferences(
+        self, result_df: pd.DataFrame, key: str = None, processor_order: List = None
+    ) -> pd.DataFrame:
+        """Extract inferences using generic method if no template or custom post-processor is provided."""
+        predicted_data = read_jsonl_files(resolve_io_path(self.prediction_dataset))
+        pred_list = []
+        if self.prediction_column_name in predicted_data[0].keys():
+            key = self.prediction_column_name
+        else:
+            key = key if key else "0"
```
arun-rajora, line 375: Lines 372 to 375 also exists in lines 311-314. Can we just handle this scenario just once?

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+            predicted = row.get(key)
+            if isinstance(predicted, list) and len(predicted[0]) > 1:
+                curr_pred_list = []
+                for i in range(0, len(predicted)):
+                    out_string = predicted[i]
+                    out_string = self.apply_generic_processor(out_string, row)
+                    curr_pred_list.append(out_string)
+                pred_list.append(curr_pred_list)
+            else:
+                out_string = predicted if isinstance(predicted, str) else predicted[0]
```
arun-rajora, line 386: Is it safe to assume that `predicted[0]` won't throw an error here? If `predicted` is an empty array, should we handle it by using a default value like empty string or throw a proper error message?

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+                out_string = self.apply_generic_processor(out_string, row)
+                pred_list.append(out_string)
+        if isinstance(pred_list[0], list) and len(pred_list[0]) > 1:
```
arun-rajora, line 389: * Why does `len(pred_list[0])` have to be > 1. Should it be > 0?
* Also, can we handle the scenario when `pred_list` is empty?

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+            cols = [
+                f"{self.prediction_column_name}_{i+1}" for i in range(len(pred_list[0]))
+            ]
+        else:
+            cols = self.prediction_column_name
+        # result_df[self.prediction_column_name] = pred_list
```
arun-rajora, line 395: nit: Unused code comment. Please remove it.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+        result_df[cols] = pred_list
+        return result_df
+
+    def run(self):  # key: str = None, processor_order: List = None) -> None:
```
arun-rajora, line 399: nit: Unused code comment. Please remove it.

arun-rajora (changes requested):

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
```diff
@@ -29,10 +29,16 @@ inputs:
       Key in ground truth dataset that contains ground truth. If ground_truth_dataset is given, then,
       this is required input.
     optional: True
-  prediction_probabilities_dataset:
-    type: uri_file
+  remove_prefixes:
+    type: string
     description: >-
-      A file that contains the prediction probabilities.
+      A set of string prefixes separated by comma list of string prefixes to be removed from the inference
```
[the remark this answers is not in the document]
Nivi09: @arun-rajora No, because when we say set of prefixes, it could mean the set of characters as we also pass the set of characters in " " and strip remove any of these characters that string starts with, and after removal it checks again if it starts with one of the characters given in set. Whereas string prefix operate differently. For example : "###,>>>" prefixes will first check if  string starts with '###', if yes, then it will be removed and then only the second given prefix is applied. That's the reason that you will find that we apply strip method while stripping characters but we don't apply this method while removing prefixes. I added word string while having a discussion with Sasha where we came to conclusion that prefixes field was there for strings.

assets/aml-benchmark/components/src/inference_postprocessor.zip
Nivi09: @arun-rajora  Thank you. It's due to switching between branches and merging and cleaning, that things got copied from one to another. Thank you pointing it out.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: @arun-rajora Too many methods even to handle a very linear case does not seem appealing to me. That is why I kept it as it is instead of creating a common method.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: @arun-rajora I have never come across any scenario where Null was predicted. Either there would be a text or empty string. And in case of empty strings, it has worked before. Theoretically, I don't think there could be a scenario where nothing gets predicted as we never pass Null to the model.

assets/aml-benchmark/tests/pipelines/inference_postprocessor_pipeline.yaml
```diff
@@ -15,25 +15,22 @@ inputs:
...
+  strip_characters: '\n\n'
+  extract_number: 'first'
   template: '{{prediction}}'
-  extract_value_at_index: 0
   script_path:
     type: uri_file
-    path: ../data/test_custom_postprocessor.py
-  encoder_config: '{"column_name":"label", "0":"NEUTRAL", "1":"ENTAILMENT", "2":"CONTRADICTION"}'
-  
+    path: ../data/sample_ground_truths.jsonl
```
iamrk04, line 26: Just curious if this is what was intended since the expectation of script_path is to accept a python script and not a jsonl file.

iamrk04 (changes requested):

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: @arun-rajora Done. Thanks.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: Done. Thanks.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: command
 name: inference_postprocessor
 display_name: Inference Postprocessor
 description: Inference Postprocessor
-version: 0.0.1
+version: 0.0.1.3
```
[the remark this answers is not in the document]
Nivi09: Thanks. Will change.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
```diff
@@ -42,36 +48,56 @@ inputs:
...
     optional: True
+  find_first:
+    type: string
+    description: >-
+      A list of strings to search for in the inference results. The first occurrence of each string will be extracted. 
+      Must provide a comma-separated list of strings.
+      Example:
+      >>> find_first = "positive,negative"
+      >>> completion = "This is a positive example, not negative"
+      # Output: "positive"
```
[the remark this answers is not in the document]
Nivi09: @iamrk04  This description was taken from babel, but the return is the first found match. I will add this - 
`The first occurrence of each string will be extracted and the occurrence with minimum index will be returned`.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
```diff
@@ -42,36 +48,56 @@ inputs:
...
+    optional: True
+  extract_number:
```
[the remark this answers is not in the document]
Nivi09: I know, I think I was facing the issue at that time and removed it. I will add it again. The previous component 0.1.1.3 has this has enum.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -19,68 +25,106 @@
 
 logger = get_logger(__name__)
 
-jinja2.filters.FILTERS['zip'] = zip
+jinja2.filters.FILTERS["zip"] = zip
 ENV = jinja2.Environment()
 ENV.globals.update(zip=zip)
 
 
+def get_prompt(data: dict, remove_prompt_prefix: bool = True):
```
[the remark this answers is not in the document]
Nivi09: @iamrk04 This function is for Babel use, and they mentioned it as only as `data: Optional[Dict] = None`. That is why I added it as it is, just to minimize the level of change.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
-                "Please provide python script containing your custom postprocessor logic."
-            )
+        if self.user_postprocessor and not self.user_postprocessor.endswith(".py"):
+            mssg = "Please provide python script containing your custom postprocessor logic."
             raise BenchmarkValidationException._with_error(
                 AzureMLError.create(BenchmarkValidationError, error_details=mssg)
             )
 
-    def extract_using_template(self) -> None:
+    def read_ground_truth_dataset(self, result_df) -> pd.DataFrame:
```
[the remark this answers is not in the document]
Nivi09: @iamrk04 Done. Thanks.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
```diff
@@ -90,129 +134,606 @@ def __post_init__(self) -> None:
...
+                        else:
+                            m = parts[-1] if strategy == "last" else parts[0]
+                    try:
+                        # Test that the matched string is a number
+                        np.fromstring(m, sep=" ")
+                        return m
+                    except SyntaxError:
+                        # we matched with something that is not a number
+                        pass
+        return default
```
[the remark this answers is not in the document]
Nivi09: @iamrk04 This is taken from Babel and I dare not to look into simplification right now as first, it might delay the check-ins, second, the test cases created by Babel has passed this. This is something that breaks sometimes, and they modify it. The latest change was made on Sept 25th. For now, I'm taking their experience into account kept it as it is.

assets/aml-benchmark/tests/pipelines/inference_postprocessor_pipeline.yaml
Nivi09: @iamrk04 It's just a placeholder. All the values of the component's fields get set in test_inference_postprocessor.py

assets/aml-benchmark-zipped.zip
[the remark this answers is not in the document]
Nivi09: Done. Thanks.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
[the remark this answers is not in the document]
Nivi09: Done. Thanks.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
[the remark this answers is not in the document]
Nivi09: Done. Thanks.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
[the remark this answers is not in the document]
Nivi09: Done. Thanks.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
Nivi09: Done. Thanks.

assets/aml-benchmark/components/inference-postprocessor/spec.yaml
```diff
@@ -29,10 +29,16 @@ inputs:
       Key in ground truth dataset that contains ground truth. If ground_truth_dataset is given, then,
       this is required input.
     optional: True
-  prediction_probabilities_dataset:
```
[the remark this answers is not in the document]
Nivi09: @arun-rajora Because model prediction component has made ground truth field optional. In case this component becomes a part of inference, then this will be applicable.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: @alpatots I have tested these via test_inference_postprocessor_as_script and test cases are added there. I have tested that each every line gets tested by using debugger.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: Done. Thanks.

assets/aml-benchmark/components/src/my_history.txt
Nivi09: I have removed it. Thanks.

user_script.json
Nivi09: Removed. Thanks.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: @arun-rajora 
1. As I mentioned before, I have never come across any scenario where a language model won't predict anything, and theoretically, language modelling P(generate a word | context) means it will at least generate an empty string. 
2. I have never come across a scenario where single prediction is in the list, model prediction component generally returns a string. So, if I check it as >0, then curr_pred_list will get created and then in the pred_list, it will appear as [["pred_value"]], and what I want is "pred_value" only so that dataframe can be consistent.

assets/aml-benchmark/components/src/inference_postprocessor.zip
Nivi09: Done. Thanks

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
arun-rajora: Why do we need to determine `key` multiple times? Can't we just set it once?

Nivi09: @arun-rajora This `key` is the `completion_key` that Babel uses and to set the current and future development, this is a parameter that I added now. It is not a part of 3P.

assets/aml-benchmark/components/src/inference_postprocessor/runner.py
Nivi09: Added to babel repo.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
arun-rajora: I can imagine when working with endpoints, these assumptions can easily break where the API requests to the endpoints can fail as well. So, it is better if we handle very minor corner cases as well.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
arun-rajora: I can imagine when working with endpoints, these assumptions can easily break where the API requests to the endpoints can fail as well. So, it is better if we handle very minor corner cases as well.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: @alpatots I did not understand and need bit more elaboration on this. Let me know please when can we discuss this.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
Nivi09: Done. Thanks.

assets/aml-benchmark/components/src/inference_postprocessor/inference_postprocessor.py
alpatots: @Nivi09 sorry about not being clear. It's a really small comment, but in general it seems strange (to me) to have a code like:

```python
def f(path_to_data, some_other_data):
    data = read_from(path_to_data)
    do_something(data, some_other_data)
```

I would expect to see 

```python
def f(some_data, some_other_data):
    do_something(data, some_other_data)
```

user_script.json
iamrk04, line 1: Looks like this is committed by mistake. If so, please remove this. Let's be careful with what we are committing. Thanks!

iamrk04 (changes requested):

alpatots (dismissed):

user_script.json
Nivi09: That's done. This was mentioned in Sasha's comment, and I had removed it yesterday. I don't know why it not got reflected. Removed the file again. Thanks.

iamrk04 (approved):

arun-rajora (approved):

PR #1459, opened 2023-10-11
use none check

hawestra:
Check none so that metrics value gets propagated if the value is 0

### DIFF
--- a/assets/model_monitoring/components/tests/unit/test_metric_output_builder.py
+++ b/assets/model_monitoring/components/tests/unit/test_metric_output_builder.py
@@ -14,6 +14,80 @@ from model_monitor_metric_outputter.builder.metric_output_builder import MetricO
 class TestMetricOutputBuilder:
     """Test class for metric output builder."""
 
+    def test_metrics_with_none_values(self, mock_runmetric_client, monitor_name, signal_name):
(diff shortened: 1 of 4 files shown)


yetamsft (approved):

yetamsft:
Could you please bump the version here so that backend can pick it up?
https://github.com/Azure/azureml-assets/blob/main/assets/model_monitoring/components/model_monitor/model_monitor_evaluate_metrics_threshold/spec.yaml

assets/model_monitoring/components/tests/unit/test_metric_output_builder.py
```diff
@@ -14,6 +14,80 @@
...
+                threshold_value=0,
+            ),
+            Row(
+                metric_name="num_calls_with_status_code_500",
+                metric_value=63.0,
+                threshold_value=None,
+            ),
+            Row(
+                metric_name="num_calls_with_status_code_501",
+            ),
```
yetamsft, line 42: Could you please also add a test case for values == `NaN`, just to be explicit?

yetamsft (approved):
Looks good to me!

yunjie-hub (approved):

PR #1462, opened 2023-10-12
Change to hash to map the payload and ground truth

yuanzhuangyuanzhuang:


assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
```diff
@@ -61,26 +61,14 @@ inputs:
     type: string
     optional: False
     description: Connections name for the endpoint.
-  metadata_key:
-    type: string
-    optional: True
-    description: The metadata key that in the request payload that contains label columns. If not provided, the `request_metadata` will be used in the batch benchmark score output to retrieve ground truth.
   label_key:
```
arun-rajora, line 64: nit: Can we rename this label_column_name?

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
```diff
@@ -93,6 +81,10 @@ inputs:
     description: The fallback value that can be used when request payload failed. If not provided, the fallback value will be an empty string.
     type: string
     optional: True
+  metadata_key:
```
arun-rajora, line 84: Is `metadata_key` needed any longer?

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
```diff
@@ -31,6 +31,10 @@ inputs:
         "_batch_request_metadata": ###<_batch_request_metadata>
       }
     optional: False
+  label_key:
```
arun-rajora, line 34: nit: Can this be label_column_name?

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
```diff
@@ -47,14 +51,19 @@ outputs:
   formatted_data:
     type: mltable
     description: Path to the folder where the payload will be stored.
+  output_metadata:
```
arun-rajora, line 54: nit: Can this be renamed as ground_truth_metadata?

assets/aml-benchmark/components/batch-output-formatter/spec.yaml
```diff
@@ -14,12 +14,9 @@ inputs:
     optional: True
     description: The metadata key that stores ground truth `label_key` in the response. If using azureml llama model and pass `_batch_request_metadata`, then this one can left empty or using `request_metadata`.
   label_key:
```
arun-rajora, line 16: nit: Can this be `label_column_name`?

arun-rajora (changes requested):

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
yuanzhuangyuanzhuang: This is still useful for the llama dataset or dataset that customer may want to skip hash. If this one is not enabled, then the hash will be used

assets/aml-benchmark/components/batch-output-formatter/spec.yaml
yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
arun-rajora: >  llama dataset 

I assume you meant llama model? 

If we don't expose this parameter to the user, things would still work for the user, right? Additionally, it reduces the number of parameters the user has to understand. I am thinking we can hide this parameter for now (we can always bring it back based on feedback). What do you think?

assets/aml-benchmark/components/batch-output-formatter/spec.yaml
```diff
@@ -13,13 +13,10 @@ inputs:
...
     description: The metadata key that stores ground truth `label_key` in the response. If using azureml llama model and pass `_batch_request_metadata`, then this one can left empty or using `request_metadata`.
-  label_key:
-    type: string
-    optional: False
-    description: The key that stores the ground truth value.
-  data_id_key:
+  label_column_name:
     type: string
     optional: True
+    description: The key that stores the ground truth value.
```
arun-rajora, line 19: nit: We can update the description as well like you have for `batch-inference-preparer`- "The label column name.".

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
```diff
@@ -31,6 +31,10 @@ inputs:
         "_batch_request_metadata": ###<_batch_request_metadata>
       }
     optional: False
+  label_column_name:
+    type: string
+    optional: True
+    description: The key that stores the ground truth value.
```
arun-rajora, line 37: nit: We can update the description as well like you have for `batch-inference-preparer`- "The label column name.".

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
yuanzhuangyuanzhuang: I am just thinking this can be removed and the only key we will look up is `batch_metadata` in the component output. This will happen only if the customer is using the exact payload which contains `_batch_request_metadata` and we will convert this key to `batch_metadata` automatically in the final output.

assets/aml-benchmark/components/batch-resource-manager/spec.yaml
```diff
@@ -7,6 +7,9 @@ description: Resource Manager for batch inference.
 version: 0.0.1
 
 inputs:
+  wait_input:
+    type: uri_file
+    optional: True
```
iamrk04, line 12: just curious, what is the use of this?

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
arun-rajora: Sounds good. Thanks!

assets/aml-benchmark/components/batch-resource-manager/spec.yaml
yuanzhuangyuanzhuang: I just add a fake input that the component can wait until the run is finished. Then it can execute deletion of the endpoiint

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-output-formatter/spec.yaml
yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
yuanzhuangyuanzhuang: removed

assets/aml-benchmark/components/batch-inference-preparer/spec.yaml
```diff
@@ -47,14 +51,19 @@ outputs:
   formatted_data:
     type: mltable
     description: Path to the folder where the payload will be stored.
+  ground_truth_metadata:
+    type: uri_folder
+    description: Path to the folder where the payload will be stored.
```
arun-rajora, line 56: This has the same description as `formatted_data`- both says path to the folder where the payload will be stored. I am assuming this description has to be updated.

yuanzhuangyuanzhuang: changed

assets/aml-benchmark/components/batch-resource-manager/spec.yaml
iamrk04: How are we using this? This input is not even passed to the `main` script. Am I missing something?

arun-rajora (approved):

assets/aml-benchmark/components/batch-resource-manager/spec.yaml
arun-rajora: @iamrk04 , I believe this input will be connected to some output of inferencing component so that it runs after it has finished. It is marked as optional so that even if inferencing fails, this component can run. That is why it is "wait_input", just to make sure it runs __after__ inferencing is done.

PR #1464, opened 2023-10-12
Switch off generating prompt asset type

lykelly19:


scripts/docs/generate_asset_documentation.py
```diff
@@ -103,7 +103,8 @@ def create_asset_info(asset_config):
         if asset_config.type == AssetType.DATA:
             return DataInfo(asset_config)
         if asset_config.type == AssetType.PROMPT:
-            return PromptInfo(asset_config)
+            return None
+            # return PromptInfo(asset_config)
 
         raise Exception(f"Not supported asset type {asset_config.type}. Use {SUPPORTED_ASSET_TYPES}")
```
elesel, line 109: You could instead check the asset type against SUPPORTED_ASSET_TYPES at the top of this function and warn (instead of raise exception) if it's not supported. I think that would better support benchmarking results.

scripts/docs/generate_asset_documentation.py
```diff
@@ -103,7 +103,8 @@ def create_asset_info(asset_config):
         if asset_config.type == AssetType.DATA:
             return DataInfo(asset_config)
```
elesel, line 104: Not related to this PR, but you could create a dict that maps asset type to the function that generates its documentation, which would greatly simplify the above and probably also take over the function of SUPPORTED_ASSET_TYPES. Something like:

```python
TYPE_TO_DOC_FUNCS = {
    AssetType.ENVIRONMENT: EnvironmentInfo,
    AssetType.COMPONENT: ComponentInfo,
    ...
}
```

and then create_asset_info would do something like:

```python
if asset_config.type not in TYPE_TO_DOC_FUNCS:
    warn and return
return TYPE_TO_DOC_FUNCS[asset_config.type](asset_config)
```

elesel (approved):

PR #1471, opened 2023-10-12
Add RPS and TPS for endpoint compute metrics

yuanzhuangyuanzhuang:


assets/aml-benchmark/components/src/compute_performance_metrics/main.py
```diff
@@ -460,6 +464,8 @@ def main(
                         for val in zip(latency_data, input_output_token_data)
                     ],
                 )
+    if run_timespan is not None:
+        results["request_per_second"] = len(all_data) / run_timespan
```
jialiu103, line 468: request -> requests

jialiu103 (approved):

assets/aml-benchmark/components/src/compute_performance_metrics/main.py
```diff
@@ -260,6 +261,7 @@ def main(
             ]
         )
         latency_data = (end_data - start_data) * 1000
+        run_timespan = np.max(end_data) - np.min(start_data)
```
iamrk04, line 264: @yuanzhuangyuanzhuang 
I think it will be better if we do the computation `tokens per sec` and `requests per sec` for every record, and then take the average. The current approach does not take into account that there can be some delay in between two requests. Although this would rarely happen in batch infer component (we never know about model prediction), we shouldn't be taking chances. What do you think?

yuanzhuangyuanzhuang: I think the problem is that for batch inference, we need to take the concurrent calls into account. The concurrent is currently dynamically adjusted to achieve the best throughput. I am not sure if a good way that we can do this. If we use a fixed number of concurrent calls, this can be achieved by the individual one times the number of the concurrent calls.

yuanzhuangyuanzhuang: changed

iamrk04 (approved):

PR #1477, opened 2023-10-13
update multi workers for mii, fix bugs, make vllm default

msp8955:


assets/training/model_management/environments/foundation-model-inference/context/requirements.txt
```diff
@@ -10,7 +11,7 @@ azureml-inference-server-http=={{latest-pypi-version}}
 azureml-mlflow=={{latest-pypi-version}}
 azureml-ai-monitoring==0.1.0b4
 azure-ai-contentsafety~=1.0.0b1
-cryptography~=41.0.3
+cryptography~=41.0.4
```
novaturient95, line 14: could we remove `cryptography` dep from here, and consume from source?

novaturient95 (approved):

assets/training/model_management/environments/foundation-model-inference/context/requirements.txt
msp8955: we have to pin this because of the vulnerability, otherwise it is pulling lower version of cryptography
similar to the pin from here: https://github.com/Azure/azureml-assets/blob/main/assets/training/general/environments/acpt-pytorch-2.0-cuda11.7/context/requirements.txt

PR #1496, opened 2023-10-17
standardize categorical code

hawestra:
use same categorical code in all places in feature importance computation


assets/model_monitoring/components/src/feature_importance_metrics/feature_importance_utilities.py
```diff
@@ -28,6 +28,40 @@ def convert_pandas_to_spark(pandas_data):
...
+    if pd.api.types.is_integer_dtype(baseline_column):
+        distinct_column_values = len(baseline_column.unique())
+        total_column_values = len(baseline_column)
+        distinct_value_ratio = distinct_column_values / total_column_values
+        if distinct_value_ratio < 0.05:
+            return False
+        else:
+            return True
+    # Log the datatype detected and default to true
+    log_time_and_message(f"Unknown column type: {baseline_column_type}")
```
VivienneTang, line 61: so for all the unknown type of column, we set it as categorical column?

hawestra: yes, there shouldn't be too many of these but lightgbm (the model we use for calculating the feature importances) only really can "understand" simple datatypes ie float, int and all other data types will need to be marked as categorical so that lightgbm will do the conversion of the column to int

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
```diff
@@ -25,7 +25,7 @@
     pass
 
 from feature_importance_metrics.feature_importance_utilities import (
-    compute_categorical_features, convert_pandas_to_spark, log_time_and_message)
+    compute_categorical_features, convert_pandas_to_spark, convert_time_columns, is_categorical_column, log_time_and_message)
```
VivienneTang, line 28: convert_time_columns is missing. Could you add this function?

hawestra: removed thanks!

assets/model_monitoring/components/feature_attribution_drift/feature_attribution_drift_signal_monitor/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: pipeline
 name: feature_attribution_drift_signal_monitor
 display_name: Feature Attribution Drift - Signal Monitor
 description: Computes the feature attribution between a baseline and production data assets.
-version: 0.3.7
+version: 0.3.8
```
VivienneTang, line 7: I think we already have 0.3.9 on main branch

assets/model_monitoring/components/src/feature_importance_metrics/feature_importance_utilities.py
```diff
@@ -28,6 +28,41 @@ def convert_pandas_to_spark(pandas_data):
...
+    :param column_name: the column to determine
+    :type column_name: string
+    :param baseline_data: The baseline data meaning the data used to create the
+    model monitor
+    :type baseline_data: pandas.DataFrame
+    :rtype: boolean
+    """
+    baseline_column = pd.Series(baseline_data[column_name])
+    baseline_column_type = baseline_column.dtype.name
+    if (pd.api.types.is_float_dtype(baseline_column) or
```
yunjie-hub, line 43: Does this include the INT type?

assets/model_monitoring/components/src/feature_importance_metrics/feature_importance_utilities.py
```diff
@@ -28,6 +28,41 @@ def convert_pandas_to_spark(pandas_data):
...
+            pd.api.types.is_datetime64_ns_dtype(baseline_column) or
+            pd.api.types.is_timedelta64_ns_dtype(baseline_column)):
+        return False
+    # treat all datetime types as categorical since LightGBM cannot accept anything but bool, int and float
+    if (pd.api.types.is_object_dtype(baseline_column) or pd.api.types.is_string_dtype(baseline_column)
+            or baseline_column_type == "bool" or pd.api.types.is_datetime64_dtype(baseline_column) or
+            pd.api.types.is_timedelta64_dtype(baseline_column)):
+        return True
+    if pd.api.types.is_integer_dtype(baseline_column):
```
yunjie-hub, line 52: Do we need some default value if it doesn't match any of the "IF"s

yunjie-hub (dismissed):

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: pipeline
 name: data_drift_signal_monitor
 display_name: Data Drift - Signal Monitor
 description: Computes the data drift between a baseline and production data assets.
-version: 0.3.15
+version: 0.3.17
```
yunjie-hub, line 7: should be 16 or 17?

hawestra: https://github.com/Azure/azureml-assets/blob/main/assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml

17 the current version is 0.3.16, I think github isn't great for these PR diffs

yunjie-hub (approved):

PR #1524, opened 2023-10-19
fix hash algo for benchmark output formatter

yuanzhuangyuanzhuang:


assets/aml-benchmark/components/src/utils/online_endpoint/endpoint_utils.py
```diff
@@ -93,7 +93,7 @@ def hash_payload_prompt(
                 if isinstance(prompt[0], dict):
                     prompt = " ".join([p['content'] for p in prompt])
             elif model.is_aoai_model():
-                prompt = payload["messages"][0]["content"]
+                prompt = " ".join([msg.get("content", "") for msg in payload["messages"]])
```
EricWrightAtWork, line 96: You might want to also handle the same where paylod["messages"] is not a list. I'm not sure if it comes up, but it's easy to handle both cases

yuanzhuangyuanzhuang: added for other scenarios

EricWrightAtWork (approved):

PR #1531, opened 2023-10-20
add more supported dataType

VivienneTang:
1. Add more supported dataType when cast dataType
2. If the dataType is not supported, we do not throw runtime exception, instead we set the datatype violation number be the default value 0 and log a warning
3. added unit test for the changes.

### DIFF
--- a/assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
+++ b/assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
@@ -387,7 +387,7 @@ def impute_categorical_with_mode(df: pyspark.sql.DataFrame) -> pyspark.sql.DataF
 
 
 def modify_dataType(data_stats_table) -> pyspark.sql.DataFrame:
-    """Cast DataType() to DataType"""
+    """Cast DataType() to DataType."""
     data_stats_table_mod = data_stats_table.withColumn(
         "dataType",
         when(col("dataType") == "DoubleType()", "double")
--- a/assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
+++ b/assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
@@ -2,8 +2,10 @@
 # Licensed under the MIT License.
 
 """This file contains the core logic for data quality metrics component."""
-from typing import Tuple
+
 import pyspark
+import pyspark.pandas as ps
+import warnings
 from pyspark.context import SparkContext
 from pyspark.sql.session import SparkSession
 from pyspark.sql.types import (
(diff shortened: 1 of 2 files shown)


assets/model_monitoring/components/tests/unit/test_compute_data_quality_metrics.py
```diff
@@ -0,0 +1,113 @@
...
+            ('feature_double', 'double'),
+            ('feature_binary', 'binary'),
+            ('feature_date', 'date'),
+            ('feature_decimal', 'decimal'),
+            ('feature_float', 'float'),
+            ('feature_integer', 'integer'),
+            ('feature_long', 'long'),
+            ('feature_short', 'short'),
+            ('feature_string', 'string'),
+            ('feature_char', 'char(30)'),
```
yunjie-hub, line 51: You tested all the cast be very successful, right?

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
```diff
@@ -29,13 +31,13 @@
     concat,
 )
 from pyspark.ml.feature import Imputer
-import pyspark.pandas as ps
+from typing import Tuple
```
yunjie-hub, line 34: is this an order change?

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
```diff
@@ -376,6 +386,29 @@ def impute_categorical_with_mode(df: pyspark.sql.DataFrame) -> pyspark.sql.DataF
     return df
```
yunjie-hub, line 388: Could you please add some comments here, for a new mapping added her, please added in the supported_datatypes array.

assets/model_monitoring/components/tests/unit/test_compute_data_quality_metrics.py
VivienneTang: Yes, when I test the cast for "char" and "varchar". It showed that I have to case with char(int) and varchar(int), so I added an int in char and varchar

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
VivienneTang: Yes, only to order the import by alphabet order

yunjie-hub: then put the import warnings down?

VivienneTang: Done in the new PR: https://github.com/Azure/azureml-assets/pull/1533

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
VivienneTang: Done in the new PR: https://github.com/Azure/azureml-assets/pull/1533

PR #1541, opened 2023-10-23
Handle device failure fix, incrementing pipeline version, removing unused parameter

sarthaks95:


assets/training/model_evaluation/src/validation.py
```diff
@@ -267,7 +267,13 @@ def validate_Xy(X_test, y_test):
         X_test (_type_): _description_
         y_test (_type_): _description_
     """
-    _validate_data_passed(X_test, "test_data")
+    # _validate_data_passed(X_test, "test_data")
```
guptha23, line 270: this functions is commented. can you please check if this is expected?

guptha23 (approved):

assets/training/model_evaluation/src/validation.py
sarthaks95: That's deliberate.

PR #1544, opened 2023-10-23
Change ground truth column name for benchmark

yuanzhuangyuanzhuang:


assets/aml-benchmark/components/src/batch_output_formatter/result_converters.py
```diff
@@ -106,7 +107,7 @@ def convert_result_ground_truth(self, result: Dict[str, Any]) -> Dict[str, Any]:
             request_payload = self._get_request(result)
             payload_hash = EndpointUtilities.hash_payload_prompt(request_payload, self._model)
             ground_truth = self._lookup_dict.get(payload_hash, '')
-        return {'ground_truth': ground_truth}
+        return {self.ground_truth_column_name: ground_truth}
```
arun-rajora, line 110: Can we add a test for it or is there a test that needs to be updated for this change?

arun-rajora (approved):

assets/aml-benchmark/components/src/batch_output_formatter/result_converters.py
yuanzhuangyuanzhuang: tests added

arun-rajora (approved):

PR #1546, opened 2023-10-23
Adding Document Intelligence support to crack and chunk components

shpengmsft:
Adding optional parameters - doc_intel_connection_id to both crack_and_chunk and crack_and_chunk_and_embed components.


assets/large_language_models/rag/components/crack_and_chunk_and_embed/spec.yaml
```diff
@@ -41,6 +41,10 @@ inputs:
     type: integer
     default: 0
     description: "Number of tokens to overlap between chunks."
+  doc_intel_connection_id:
```
gdyre, line 44: I think this also needs to be added to the command below as well

gdyre: For both specs

gjwoods (approved):

PR #1555, opened 2023-10-24
Add tag validation for prompt assets

alexkalita:
Add validation for tags in prompt assets.  The set of valid tags is configured in a yaml file which is checked into the source tree.  Tags can be marked as required or optional, and they can have a single value (e.g. modality=chat) or multiple values in a comma-separated list (e.g. task=text-classification,summarization).

For now, only prompt and eval assets are checked for tags.  Later, we may expand this to models or other asset types.


scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -399,6 +455,9 @@ def validate_assets(input_dirs: List[Path],
                 if check_build_context:
                     error_count += validate_build_context(asset_config.extra_config_as_object())
 
+            if asset_config.type == assets.AssetType.PROMPT:
```
wayliums, line 458: let's include EvaluationResult type here too

wayliums (approved):

scripts/azureml-assets/azureml/assets/validate_assets.py
alexkalita: Added validation for eval results.  It uses the same list as prompts except prompt has modality set as required.

guoyukathy (approved):

vizhur (approved):

PR #1557, opened 2023-10-25
rename blip2 model

sharma-riti:


assets/models/system/salesforce-blip2-opt-2.7b/spec.yaml
```diff
@@ -1,5 +1,5 @@
 $schema: https://azuremlschemas.azureedge.net/latest/model.schema.json
-name: salesforce-blip2-opt-2.7b
+name: Salesforce-BLIP-2-opt-2.7b
```
PhaniShekhar, line 2: This name should be in sync with the deploy configs.
It is possible that model publish pipeline doesn't support "." and hence changes it to "-", but the pipeline to register assets might support it. I will add you to a thread with Ayush and others, where we discussed similar issue with mot models containing "_" character.

PhaniShekhar (dismissed):

assets/models/system/salesforce-blip2-opt-2.7b/spec.yaml
sharma-riti: updated.. :)

haritha-bendapudi (approved):

PR #1565, opened 2023-10-25
datatype classification

VivienneTang:
1. classify new datatypes into numerical and categorical
2. exclude the boolean type from outofbounds metrics

3. add e2e tests with timestamp and boolean datatype


assets/model_monitoring/components/src/data_quality_statistics/compute_data_quality_statistics.py
```diff
@@ -133,8 +153,11 @@ def compute_data_quality_statistics(df):
     """Compute data quality statistics."""
     dtype_df = get_df_schema(df=df).to_pandas_on_spark()
     unique_vals_df = get_unique_value_list(df=df)
-    max_vals = compute_max_df(df=df.to_pandas_on_spark())
-    min_vals = compute_min_df(df=df.to_pandas_on_spark())
+    # Note: excluding boolean type column as the boolean type do not need to be calculated
```
yunjie-hub, line 156: Is boolean the only type doesn't support max/min value operation here?

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
```diff
@@ -539,6 +539,17 @@ def compute_data_quality_metrics(df, data_stats_table):
         when(col("dataType") == "DoubleType()", "Numerical")
         .when(col("dataType") == "LongType()", "Numerical")
         .when(col("dataType") == "StringType()", "Categorical")
+        .when(col("dataType") == "IntegerType()", "Numerical")
+        .when(col("dataType") == "TimestampType()", "Categorical")
+        .when(col("dataType") == "BooleanType()", "Categorical")
+        .when(col("dataType") == "BinaryType()", "Categorical")
```
yunjie-hub, line 545: can we have some tests with these types

assets/model_monitoring/components/src/data_quality_statistics/compute_data_quality_statistics.py
VivienneTang: As I added supported type list, we support 6 datatypes in pyspark: [StringType, IntegerType, DoubleType, ByteType, LongType, FloatType (float might not be accurate, we suggest double instead)]. For not supported type, we just do not hav max, min value and we just do not generate the outofbands metrics. We do not fail the job or throw exceptions for unsupported types.

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
VivienneTang: Yes, I have unit test for all supported and some unsupported type. The data quality metrics are generated as we expected. no failure or exception for unsupported type. Also we have a e2e test to cover common datatypes.

assets/model_monitoring/components/src/data_quality_statistics/compute_data_quality_statistics.py
```diff
@@ -3,18 +3,20 @@
...
 from pyspark.sql.types import StructType, StructField, StringType
+import pyspark
+import pyspark.pandas as ps
 
 
 # Init spark session
 sc = SparkContext.getOrCreate()
 spark = SparkSession(sc)
 
+supported_datatype_for_max_min_value = ["IntegerType()", "DoubleType()", "ByteType()", "LongType()", "FloatType()"]
```
yunjie-hub, line 18: do we have any warnings for the non-support max-min which should be used for out bound metrics, right?

yunjie-hub (approved):

assets/model_monitoring/components/src/data_quality_statistics/compute_data_quality_statistics.py
VivienneTang: No. It does not add much value for warning in our driver logs because we will have a list of metrics generated as output. The user can view the metrics for each feature in the output easily, they will notice that some feature does not have outofbands easily.

PR #1578, opened 2023-10-27
humaneval_gptchatsupport

SamGos93:
Custom Post Processing logic for Human Eval GPT4.


assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval_postprocessor_template.py
```diff
@@ -0,0 +1,258 @@
+# ---------------------------------------------------------
```
iamrk04, line 1: We can update the file name to `humaneval.py`. Thanks!

iamrk04 (approved):

iamrk04 (changes requested):
Can we fix the script syntax failure?

assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval_postprocessor_template.py
```diff
@@ -0,0 +1,258 @@
...
+            pred_combined_prompt = row["original_prediction"]
+        else:
+            pred_combined_prompt = row["prompt"]+"\n"+row["original_prediction"]
+
+        # Applying regex on the prediction column
+        if regex_exp:
+            pred = apply_regex_expr(pred_combined_prompt, regex_exp)
+        else:
+            pred = pred_combined_prompt
+        failed, _ = code_run(pred,
```
arun-rajora, line 184: I am assuming all of this code to run the output will be cleaned up from the post processor?

arun-rajora (changes requested):

assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval_postprocessor_template.py
SamGos93: Yes, the_code_run() would be removed. The compute metrics component should show whether a prediction passed or failed the test cases. This is not available in compute metrics (atleast the component i used.) This is quite important for debugging IMO.

SamGos93: @EricWrightAtWork suggested we keep these. It would be helpful in debugging unless we have this capability in compute_metrics,

iamrk04 (approved):

PR #1585, opened 2023-10-30
Add sampling validation

hawestra:
1. log when the sampling rate causes the dataset to be too small and update fraction, example https://ml.azure.com/experiments/id/1a0a9a51-a26e-44b3-92d2-dcfd2d1fe6aa/runs/mighty_turtle_601mq93959?wsid=/subscriptions/e0fd569c-e34a-4249-8c24-e8d723c7f054/resourceGroups/hawestra-rg/providers/Microsoft.MachineLearningServices/workspaces/hawestra-ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#
3. add passrate threshold and rating thresholds as input to gsq components


assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_metrics/run.py
```diff
@@ -122,6 +121,8 @@ def run():
         )
     for metric_name in compact_metric_names:
         passrate_threshold = threshold_args[f"{metric_name.lower()}_passrate_threshold"]
+        print(passrate_threshold)
```
gaugup, line 124: Do you want to remove these prints?

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1558,10 +1559,21 @@ def apply_annotation(
         SIMILARITY: [prompt_column_name, completion_column_name, ground_truth_column_name]
     }
     # Sampling
-    production_df = production_df.sample(withReplacement=False, fraction=sample_rate)
-    production_df_with_index = production_df.withColumn("id",
-                                                        row_number()
-                                                        .over(Window.orderBy(monotonically_increasing_id()))-1)
+    production_df_sampled = production_df.sample(withReplacement=False, fraction=sample_rate)
+    if production_df_sampled.count() == 0:
+        print("Not enough data resulting from sample_rate and production dataset. "
```
gaugup, line 1564: We seem to be re-choosing the sample_rate for the user. Should you inform them about that in the print? The print statement looks like it is expecting user to fix their sample_rate parameter. Should this be an exception.

The steps taken from lines 1563 to 1570 seem a little confusing to me.

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_metrics/run.py
hawestra: thanks!

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
hawestra: updated a little. we discussed this and don't want to fail the job in this case so don't want to throw an exception but inform the user that we will not honor their sample_rate

gaugup (dismissed):

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1558,10 +1559,21 @@ def apply_annotation(
...
+              "Using default sampling value. To use custom sample_rate with this dataset, "
+              "try increasing sample_rate value.")
+        # if the production dataset is less than the sample_rate, use all the data. otherwise, default to 5.
+        fraction = float(5 / production_df.count())
```
feng-pan8014, line 1568: maybe we just do df.limit(5), which picks the first 5 rows? 
we can also do sampled_df = df.orderBy(rand()).limit(5) to make it random. but I don't think it worth to do a full random shuffle here which costs money.

force the sample rate to 100% seems too aggressive.

assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: pipeline
 name: generation_safety_quality_signal_monitor
 display_name: Generation Safety & Quality - Signal Monitor
 description: Computes the content generation safety metrics over LLM outputs.
-version: 0.4.4
+version: 0.4.5
```
yunjie-hub, line 7: do you take the latest version?

hawestra: what do you mean ?

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1558,10 +1559,18 @@ def apply_annotation(
...
     }
     # Sampling
-    production_df = production_df.sample(withReplacement=False, fraction=sample_rate)
-    production_df_with_index = production_df.withColumn("id",
-                                                        row_number()
-                                                        .over(Window.orderBy(monotonically_increasing_id()))-1)
+    production_df_sampled = production_df.sample(withReplacement=False, fraction=sample_rate)
+    if production_df_sampled.count() == 0:
+        print("Not enough data resulting from sample_rate and production dataset. "
+              "Using default sampling value of 5. To use custom sample_rate with this dataset, "
```
feng-pan8014, line 1565: maybe mention that we just pick 5 from head (that it is not a randomly sampling)?

feng-pan8014 (dismissed):

assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
yunjie-hub: I remember it's already 0.4.5 now

feng-pan8014 (approved):

PR #1590, opened 2023-10-31
Adding signatures for transformers

Anubha98:
Adding Signatures for transformers (OSS), removing the fixed column names, keeping data types as per evaluation team suggestion.


assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
```diff
@@ -23,6 +23,7 @@
     HfModelTypes,
     MLFlowHFFlavourConstants,
     LOGS_TO_BE_FILTERED_IN_APPINSIGHTS,
+    SUPPORTED_FLAVORS,
```
skanakamedal, line 26: Can we change the name like MLFLOW_FLAVORS?

Anubha98: Fixed

skanakamedal (approved):

PR #1591, opened 2023-10-31
Adding flavor map to model selector args

Anubha98:
Adding flavor map to model selector args

### DIFF
--- a/assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
+++ b/assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
@@ -249,7 +249,7 @@ ACFT_CONFIG = {
 
 FLAVOR_MAP = {
     # OSS Flavor
-    "transformers": { 
+    "transformers": {
         "tokenizer": "components/tokenizer",
         "model": "model",
         "config": "model"
--- a/assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
+++ b/assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
@@ -247,6 +247,25 @@ ACFT_CONFIG = {
 }
 
 
+FLAVOR_MAP = {
+    "transformers" : {  ## OSS flavor
+        "tokenizer" : "components/tokenizer",
+        "model" : "model",
+        "config" : "model"
+    },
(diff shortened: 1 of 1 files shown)


assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
```diff
@@ -247,6 +247,26 @@ class ModelImportConstants:
 }
 
 
+FLAVOR_MAP = {
```
jpmann, line 250: As discussed, add a task to flow these parameters from component script.

assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
```diff
@@ -480,6 +500,9 @@ def main():
         azureml_pkg_denylist_logging_patterns=LOGS_TO_BE_FILTERED_IN_APPINSIGHTS,
     )
 
+    # Adding flavor map to args
+    args.__setattr__("flavor_map", FLAVOR_MAP)
```
jpmann, line 504: nit: we can use `setattr(args, "flavor_map", FLAVOR_MAP)`

assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
```diff
@@ -421,7 +441,7 @@ def model_selector(args: Namespace):
                     mlflow_data = yaml.safe_load(fp)
                 if mlflow_data and "flavors" in mlflow_data:
                     for key in mlflow_data["flavors"]:
-                        if key in ["hftransformers", "hftransformersv2"]:
```
jpmann, line 424: Is this same logic applicable to OSS flavor as well? Are we going to save the same parameters in OSS MLModel file as we are doing for hftransformers MLModel file?

Anubha98: yes

assets/training/finetune_acft_hf_nlp/src/model_selector/model_selector.py
Anubha98: Thanks, Fixed

jpmann (approved):

PR #1593, opened 2023-10-31
Add label_key parameter in prompt crafter

SamGos93:


assets/aml-benchmark/components/prompt_crafter/spec.yaml
```diff
@@ -93,6 +93,11 @@ inputs:
     optional: true
     default: 0
     description: Random seed for sampling few-shots; if not specified, 0 is used.
+  ground_truth_column_name:
+    type: string
+    optional: true
+    default: ''
+    description: Label key/ ground truth column to be passed in the output.
```
arun-rajora, line 100: Can we remove `Label key/` from here? Also, can we update the description to explain why this is optional and when to use it?

assets/aml-benchmark/components/src/prompt_crafter/main.py
```diff
@@ -106,7 +112,8 @@ def main(
     :param n_shots: Number of shots to use for few-shot prompts.
     :param output_pattern: Pattern to use for output prompts.
     :param prompt_pattern: Pattern to use for prompts.
-    :param output_data: Path to jsonl with generated prompts.
+    :param output_file: Path to jsonl with generated prompts.
+    :param ground_truth_column_name: Ground truth column.
```
arun-rajora, line 116: Instead of `Ground truth column`, the param description should say `Ground truth column name`.

assets/aml-benchmark/components/src/prompt_crafter/package/prompt_factory.py
```diff
@@ -190,6 +191,9 @@ def process_row(self, row: Dict, index: int) -> Dict:
         if self.output_pattern is not None:
             output_data['completion'] = self.get_label_from_output_pattern(row)
 
+        if self.ground_truth_column_name is not None and self.ground_truth_column_name in row:
+            output_data['ground_truth_column_name'] = row[self.ground_truth_column_name]
```
arun-rajora, line 195: Instead of adding a column with name `ground_truth_column_name`, can we use `ground_truth`?

assets/aml-benchmark/tests/pipelines/prompt_crafter_pipeline.yaml
```diff
@@ -18,6 +18,7 @@ inputs:
   output_pattern: " {{answerKey}} "
   prompt_type: "completions"
   system_message: "Answer truthfully. "
+  ground_truth_column_name: "answerKey"
```
arun-rajora, line 21: Do we have a test that asserts that the ground truth column is in the output if it is passed in?

arun-rajora (changes requested):

assets/aml-benchmark/components/src/prompt_crafter/package/prompt_factory.py
```diff
@@ -190,6 +191,9 @@ def process_row(self, row: Dict, index: int) -> Dict:
         if self.output_pattern is not None:
             output_data['completion'] = self.get_label_from_output_pattern(row)
 
+        if self.ground_truth_column_name is not None and self.ground_truth_column_name in row:
```
arun-rajora, line 194: If `self.ground_truth_column_name` is not None and not an empty string, we should check `self.ground_truth_column_name` should exist in `row`. If not, we should throw an error saying the column does not exist in the data.

assets/aml-benchmark/components/src/prompt_crafter/package/prompt_factory.py
iamrk04: I think we should use the value that was passed in the `ground_truth_column_name` as the column name for ground_truth in the output file.

arun-rajora: @iamrk04 , just wondering if there might be odds of a collision in name in certain scenarios. We generate prompt and completion columns. If one of these is used in `ground_truth_column_name`, it would be an invalid scenario for the component.

iamrk04: makes sense, let's hardcode this to `ground_truth`. Thanks!

assets/aml-benchmark/components/prompt_crafter/spec.yaml
SamGos93: Updated.

assets/aml-benchmark/tests/pipelines/prompt_crafter_pipeline.yaml
SamGos93: Done. Added multiple tests for it.

mercerchen (commented):
Looks good. We can also remove the legacy get ground truth from output pattern method

alpatots (approved):

SamGos93:
> Looks good. We can also remove the legacy get ground truth from output pattern method

@mercerchen Can you point what you want to remove? 

The output_pattern would be used for few_shot prompts and would be used as ground truth in the absence of the ground_truth_column_name (which is optional). In case, the ground_truth_column_name is present, it would be used as ground_truth in the subsequent components. But the output_pattern would be mandatory and used for few_shot_prompts.

PR #1598, opened 2023-10-31
changing the type of validation file path from str -> default missing_type

jpmann:
defaulting the file path to _argparser_ default value for train, validation and test file paths.

### DIFF
--- a/assets/training/finetune_acft_hf_nlp/src/preprocess/preprocess.py
+++ b/assets/training/finetune_acft_hf_nlp/src/preprocess/preprocess.py
@@ -7,7 +7,8 @@ import json
 import argparse
 from pathlib import Path
 from argparse import Namespace
-
+from typing import Any
+from functools import partial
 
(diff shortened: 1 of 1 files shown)


assets/training/finetune_acft_hf_nlp/src/preprocess/preprocess.py
```diff
@@ -65,9 +73,11 @@ def get_parser():
         default=None,
         help="Train data path",
     )
+    # NOTE that the default is present in both :param `type` and `default`. In case of change, we need to update
+    # in both places
     parser.add_argument(
         "--validation_file_path",
-        type=str,
+        type=partial(default_missing_path, default=None),
```
jpmann, line 80: Does it make sense to just keep it for validation file path?

jpmann: as per suggestions, adding the same check for train and test file paths

amangupta26 (approved):

PR #1624, opened 2023-11-01
Fix timeseries data output bug

yunjie-hub:


hawestra (approved):

RichardLi1437 (approved):
so the bug is already fixed in model_monitor_output_metrics component, this PR just update all the spec.yaml to use the new version of the model_monitor_output_metrics component?

PR #1631, opened 2023-11-02
Added Hellaswag HF's preprocessor logic as custom script.

Nivi09:
Added Hellaswag HF's preprocessor logic as custom script and test case.


assets/aml-benchmark/scripts/custom_dataset_preprocessors/hellaswag_hf.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def _run(input_path: str, output_path: str) -> None:
+    """Entry function to read, run and write the processed the data."""
+    data = _read_jsonl_file(input_path)
+    processed_data = run_processor(data)
+    _write_to_jsonl_file(processed_data, output_path)
+
+
+def preprocess(text):
```
iamrk04, line 55: Can we please specify the type for parameter and return type as well?
Thanks!

assets/aml-benchmark/scripts/custom_dataset_preprocessors/hellaswag_hf.py
```diff
@@ -0,0 +1,81 @@
...
+
+def run_processor(data: List[Dict[str, Any]]) -> Union[pd.DataFrame, List[Dict[str, Any]]]:
+    """Run the custom processor function."""
+    ret_data = []
+    for doc in data:
+        ctx = doc["ctx_a"] + " " + doc["ctx_b"].capitalize()
+        out_doc = {
+            "query": preprocess(doc["activity_label"] + ": " + ctx),
+            "choices": [preprocess(ending) for ending in doc["endings"]],
+            "gold": str(doc["label"]),
```
iamrk04, line 73: `gold` seems a bit weird to store the label. Is there a specific reason we have chosen this?

assets/aml-benchmark/scripts/custom_dataset_preprocessors/hellaswag_hf.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+        ret_data.append(out_doc)
+    return ret_data
+
+
+if __name__ == '__main__':
+    argss = _parse_args()
```
iamrk04, line 80: nit: we should avoid using `argss`, please start using `args` instead.

assets/aml-benchmark/tests/test_dataset_preprocessor.py
```diff
@@ -90,6 +90,10 @@ class TestDatasetPreprocessorComponent:
                 """ {"premise":{{premise}}, "hypothesis":{{hypothesis}},"label":{{label|string}}} """,
                 None, '{"column_name":"label", "0":"NEUTRAL", "1":"ENTAILMENT", "2":"CONTRADICTION"}'
             ),
+            (
+                "hellaswag_hf", Constants.PROCESS_SAMPLE_EXAMPLES_INPUT_FILE, None,
+                os.path.join(Constants.CUSTOM_PREPROCESSOR_SCRIPT_PATH, "hellaswag_hf.py"), None
+            ),
```
iamrk04, line 96: There is no need to test the same thing via script as well as AML run. We should minimize the number of tests that we submit to AML workspace. In this case, we should keep the script test, and eliminate the component test because both will test the same thing.

iamrk04 (changes requested):

assets/aml-benchmark/scripts/custom_dataset_preprocessors/hellaswag_hf.py
```diff
@@ -0,0 +1,81 @@
...
+    processed_data = run_processor(data)
+    _write_to_jsonl_file(processed_data, output_path)
+
+
+def preprocess(text):
+    """Clean the data."""
+    text = text.strip()
+    # NOTE: Brackets are artifacts of the WikiHow dataset portion of HellaSwag.
+    text = text.replace(" [title]", ". ")
+    text = re.sub("\\[.*?\\]", "", text)
```
arun-rajora, line 60: Can we add some comments explaining what happens in "preprocess"?

arun-rajora (changes requested):

assets/aml-benchmark/scripts/custom_dataset_preprocessors/hellaswag_hf.py
Nivi09: Done. Thanks.

assets/aml-benchmark/tests/test_dataset_preprocessor.py
Nivi09: I have seen in some cases, the code works on local via script but fails via component, sometimes related to json conversion of arguments. That is the reason I keep on adding the cases just to make sure that this will work. Although I do test them on local, so I will remove it as I have already confirmed that it works.

assets/aml-benchmark/scripts/custom_dataset_preprocessors/hellaswag_hf.py
Nivi09: Some Hellaswag examples crawled from wikihow have patterns as [header] <some sentence>, [title] <some sentence> or [step] <some sentence>. So, basically we are replacing such patterns in the examples with another strings. 

Do you want me to add this ? 
       `the re.sub("\\[.*?\\]", "", text) removes any text enclosed in square brackets from the input string text.`

arun-rajora: Yes this should be enough.

Nivi09: Thanks.

arun-rajora (approved):

iamrk04 (approved):

PR #1645, opened 2023-11-02
Add new benchmark result assets

yuanzhuangyuanzhuang:
Add new benchmark result assets


jialiu103 (commented):
Let's remove agieval dataset as it doesn't have score for llama 70b model yet.

EricWrightAtWork (approved):

PR #1648, opened 2023-11-03
rename columns if custom column input

hawestra:
If the user has custom column input we need to make sure the schema is consistent with our metaprompt information; rename the columns to match prompt completion context ground_truth in this scenario: 

example:
input data is 
image
we rename the columns so instead of 
image

image


assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1557,12 +1557,18 @@ def apply_annotation(
     if SIMILARITY in metric_names and ground_truth_column_name not in production_df.columns:
         raise ValueError(f"production_dataset must have column: {ground_truth_column_name}")
 
+    # rename columns to prompt, completion, context, ground truth to match metaprompt data
+    production_df = (production_df.withColumnRenamed(prompt_column_name, PROMPT)
```
yunjie-hub, line 1561: I assumed you verified

yunjie-hub (approved):

PR #1663, opened 2023-11-03
fix throttling issue

hawestra:
Fix throttling issue; example fix : https://ml.azure.com/experiments/id/82fe6b93-2b2d-489c-917a-832ee08abef4/runs/loving_drawer_76tp20gn0s?wsid=/subscriptions/15ae9cb6-95c1-483d-a0e3-b1a1a3b06324/resourceGroups/copilot-demo-eastus/providers/Microsoft.MachineLearningServices/workspaces/aistudio_townhall&flight=AiStudio,DeployChatWebapp,SkipAADRegistration&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

### DIFF
--- a/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
+++ b/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
@@ -999,7 +999,7 @@ class _HTTPClientWithRetry:
         self.attempts = n_retry
 
         retry_strategy = Retry(
-            total=n_retry,
+            total=None,
             status=n_retry,
             status_forcelist=[104, 408, 409, 424, 429, 500, 502, 503, 504],
             backoff_factor=backoff_factor,
--- a/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
+++ b/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
@@ -999,7 +999,7 @@ class _HTTPClientWithRetry:
         self.attempts = n_retry
 
         retry_strategy = Retry(
-            total=None,
+            total=n_retry,
             status=n_retry,
             status_forcelist=[104, 408, 409, 424, 429, 500, 502, 503, 504],
             backoff_factor=backoff_factor,
--- a/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
+++ b/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
@@ -1000,9 +1000,11 @@ class _HTTPClientWithRetry:
 
         retry_strategy = Retry(
             total=n_retry,
+            status=n_retry,
             status_forcelist=[104, 408, 409, 424, 429, 500, 502, 503, 504],
             backoff_factor=backoff_factor,
             respect_retry_after_header=True,
+            allowed_methods=frozenset(['GET', 'POST', 'PUT']),
         )
         adapter = HTTPAdapter(max_retries=retry_strategy)
         self.client = requests.Session()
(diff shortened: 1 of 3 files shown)


assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1000,9 +1000,11 @@ def __init__(self, n_retry, backoff_factor):
 
         retry_strategy = Retry(
             total=n_retry,
+            status=n_retry,
             status_forcelist=[104, 408, 409, 424, 429, 500, 502, 503, 504],
             backoff_factor=backoff_factor,
             respect_retry_after_header=True,
+            allowed_methods=frozenset(['GET', 'POST', 'PUT']),
```
hawestra, line 1007: this was the main issue, there was no retry because default allowed methods don't include POST which is what we need.

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1000,9 +1000,11 @@ def __init__(self, n_retry, backoff_factor):
 
         retry_strategy = Retry(
             total=n_retry,
```
yunjie-hub, line 1002: total and status with the same value?

hawestra: image
image

do you think we should set total to None ?

yunjie-hub: I see, I was interpreting the meaning from the parameter name :)

yunjie-hub (approved):

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -1000,9 +1000,11 @@ def __init__(self, n_retry, backoff_factor):
...
+            status=n_retry,
             status_forcelist=[104, 408, 409, 424, 429, 500, 502, 503, 504],
```
feng-pan8014, line 1004: does it make sense to retry on 503?

hawestra: I think so--
image
I'm not the most well versed in http status codes but if the service is temporaryily down, we could retry in the hopes that it is temporary. Also other instance of retry in azureml-assets repo, retries on 503

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
hawestra: yeah it doesn't seem to make a lot of sense imo

akshaya-a (approved):

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
feng-pan8014: got it. it may require a longer backoff though.

feng-pan8014 (approved):

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
hawestra: I was thinking that as well, I'm going to check this one in since i have approvals and have created this WI to come back to this : [Bug 2797021](https://msdata.visualstudio.com/Vienna/_workitems/edit/2797021): [Monitoring][P2]improve backoff strategy

PR #1700, opened 2023-11-07
relax the common column criteria

VivienneTang:
1. If no common column, we raise valueerror exception. Otherwise, the none or empty dataframe cannot be written in the mltable, we will have none type error.
2. If baseline and production data column (of the same name) both in [float, double] group or [int, bigint, long, short] group, we consider them to be the same column

### DIFF
--- a/assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
+++ b/assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
@@ -4,7 +4,7 @@ type: pipeline
 name: data_drift_signal_monitor
 display_name: Data Drift - Signal Monitor
 description: Computes the data drift between a baseline and production data assets.
-version: 0.3.14
+version: 0.3.15
 is_deterministic: true
 
 inputs:
@@ -87,7 +87,7 @@ jobs:
       type: aml_token
   compute_drift_metrics:
     type: spark
-    component: azureml://registries/azureml/components/data_drift_compute_metrics/versions/0.3.7
+    component: azureml://registries/azureml/components/data_drift_compute_metrics/versions/0.3.8
     inputs:
       production_dataset:
         type: mltable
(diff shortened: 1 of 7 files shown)


assets/model_monitoring/components/src/data_drift_compute_metrics/compute_data_drift.py
```diff
@@ -47,7 +47,8 @@ def compute_data_drift_measures_tests(
        len(categorical_columns_names) == 0:
         print("No common columns found between production data and baseline"
```
hawestra, line 48: can we remove the print since we have the value error

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
       type: aml_token
   compute_histogram_buckets:
     type: spark
-    component: azureml://registries/azureml/components/model_monitor_compute_histogram_buckets/versions/0.3.2
+    component: azureml://registries/azureml/components/model_monitor_compute_histogram_buckets/versions/0.3.3
```
yunjie-hub, line 115: do you have any change for this one? or we forget the upgrade this verison?

assets/model_monitoring/components/model_monitor/model_monitor_compute_histogram_buckets/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: spark
 name: model_monitor_compute_histogram_buckets
 display_name: Model Monitor - Compute Histogram Buckets
 description: Compute histogram buckets given up to two datasets.
-version: 0.3.2
+version: 0.3.3
```
yunjie-hub, line 7: So the shared code is used in this component?

assets/model_monitoring/components/tests/unit/test_df_utils.py
```diff
@@ -131,38 +138,54 @@ def test_is_numerical(self):
...
-                                     columns=["age",
-                                              "gender"])
+        baseline_df = create_pyspark_dataframe([(1, "a"), (2, "b")],
+                                               ["id", "name"])
+        production_df = create_pyspark_dataframe([(3, "c"), (4, "d")],
+                                                 ["age", "gender"])
         assert get_common_columns(baseline_df, production_df) == {}
 
         # Test with two dataframes that have one common column
-        baseline_df = pd.DataFrame([(1, "a"), (2, "b")], columns=["id",
```
yunjie-hub, line 147: The old test is using the pandas dataframe?

yunjie-hub (approved):

PR #1703, opened 2023-11-08
add support for copying model artifact with sas token

msp8955:


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -546,6 +547,11 @@ def uri(self) -> str:
         # token if the account does not allow for anonymous access.
         self._uri = f"{self._account_uri}/{self._container_name}/{self._container_path}"
 
+        # Check whether the token is already set. If so, append the token.
```
brutusbuckeye, line 550: Perhaps reword comment to something like:

If a SAS token has been explicitly set, then assume that the URI+token is valid. Simply append the token and return. If no SAS token is set, then proceed to rest of function.

novaturient95 (approved):

lykelly19 (approved):

PR #1704, opened 2023-11-08
change json string to double quoted

RichardLi1437:
1. change json string to double quoted
2. add chat_history back as json string rather than complex type
3. update tests.

### DIFF
--- a/assets/model_monitoring/components/src/model_data_collector_preprocessor/run.py
+++ b/assets/model_monitoring/components/src/model_data_collector_preprocessor/run.py
@@ -11,11 +11,11 @@ from dateutil import parser
 import mltable
 from mltable import MLTable
 import tempfile
-
+import json
 from fsspec import AbstractFileSystem
 from azureml.fsspec import AzureMachineLearningFileSystem
 from datetime import datetime
-from pyspark.sql.functions import lit, col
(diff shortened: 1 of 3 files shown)


yunjie-hub (approved):

assets/model_monitoring/components/src/model_data_collector_preprocessor/run.py
```diff
@@ -171,7 +170,7 @@ def convert_object_to_str(dataframe: pd.DataFrame) -> pd.DataFrame:
         columns = dataframe.columns
         for column in columns:
             if dataframe[column].dtype == "object":
-                dataframe[column] = dataframe[column].astype(str)
+                dataframe[column] = dataframe[column].apply(lambda x: x if type(x) is str else json.dumps(x))
```
ycheng35xo, line 173: should we add a UT case to verify this?

ycheng35xo (approved):

assets/model_monitoring/components/src/model_data_collector_preprocessor/run.py
RichardLi1437: we already have UT in test_mdc_preprocessor.test_extract_data_and_correlation_id() to cover this

PR #1719, opened 2023-11-09
Reverting metadata finetune script changes for FTaaS

skanakamedal:
Sanity run -> https://ml.azure.com/experiments/id/9824cd46-3f00-42c7-a4f6-97e54f7288d2/runs/6b16e259-e5e4-4354-ab5a-42ec1c52e901?wsid=/subscriptions/ed2cab61-14cc-4fb3-ac23-d72609214cfd/resourceGroups/training_rg/providers/Microsoft.MachineLearningServices/workspaces/train-finetune-dev-workspace&flight=itpmerge&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

### DIFF
--- a/assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
+++ b/assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
@@ -28,10 +28,8 @@ from azureml.acft.contrib.hf.nlp.task_factory import get_task_runner
 from azureml.acft.contrib.hf.nlp.utils.common_utils import deep_update
 
 from azureml.acft.accelerator.utils.run_utils import add_run_properties
-from azureml.acft.common_components.model_selector.constants import ModelSelectorDefaults
 from azureml.acft.common_components.utils.error_handling.exceptions import ACFTValidationException
 from azureml.acft.common_components.utils.error_handling.error_definitions import ACFTUserError, ACFTSystemError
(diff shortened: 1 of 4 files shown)


assets/training/finetune_acft_hf_nlp/src/finetune/finetune.py
```diff
@@ -894,10 +883,6 @@ def finetune(args: Namespace):
     args.save_strategy = args.evaluation_strategy
     args.save_steps = args.eval_steps
 
-    args.model_metadata = update_acft_metadata(metadata=metadata,
-                                               finetuning_task=args.task_name,
```
MadhuM02, line 898: why is this revert needed? latest accelerator would expect the model_metadata to be present right?

amangupta26: We need to release Finetuning as a service components which are on older environment as all benchmarking is done on that.

We will re revert post release. Shivam has anyways started a component release for Model catalog

skanakamedal: This is for FTaaS release. Will revert once the release is done. (The FTaaS components depend on env 28 and latest env is 30. We will upgrade FTaaS after Ignite as all our benchmarking is done on env 28)

PR #1760, opened 2023-11-15
Generate SAS tokens for models & output JSON

lykelly19:


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -582,49 +602,76 @@ def uri(self) -> str:
...
+        # If we have already cached the URI, then simply return it
+        if self._uri is not None:
+            return self._uri
+        
+        # Build the simple, non-SAS URI for quick use later in the function.
+        # Its possible that the account URI may need additional tweaking to add a SAS
+        # token if the account does not allow for anonymous access.
+        self._uri = f"{self._account_uri}/{self._container_name}/{self._container_path}"
+        
+        if self.generate_uri(self._uri, self._account_uri, self._container_name, self._token) is not None:
```
elesel, line 648: `self._uri` won't be set at this point, and neither will `self._token`.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -582,49 +602,76 @@ def uri(self) -> str:
...
             return self._uri
+        
+        start_time = datetime.datetime.now(datetime.timezone.utc)
+        expiry_time = start_time + AzureBlobstoreAssetPath.SAS_EXPIRATION_TIME_DELTA
+
+        return self.generate_sas_token(self._uri, self._account_uri, self._container_name, self._storage_name, start_time, expiry_time)
```
elesel, line 654: I think it would help to swap this with line 648 above. Generate the SAS token first, if not already done, and then create the URI using the account URI and token.

scripts/azureml-assets/azureml/assets/config.py
elesel, line 1: Please move the account URI generation code in lines 521-527 into a static method.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -582,49 +608,76 @@ def uri(self) -> str:
...
                 permission=ContainerSasPermissions(read=True, list=True),
                 expiry=expiry_time,
                 start=start_time
             )
 
-            self._uri += sas_token
-            return self._uri
+            return sas_token
 
         except Exception:
```
lykelly19, line 637: Is the way to log an exception here (ex. in cause of auth error) is through a print statement? It seems like logger can't be used because it causes a circular import

elesel: You could print to stderr. I've run into the circular import issue too. :(

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -524,52 +542,62 @@ def __init__(self, storage_name: str, container_name: str, container_path: str):
...
                 timeout=AzureBlobstoreAssetPath.CONTAINER_ANONYMOUS_ACCESS_CHECK_TIMEOUT
             )
 
             if response.status_code >= 200 and response.status_code <= 299:
-                return self._uri
+                return uri
         except Exception:
             # If we fail pass through to the next approach
-            pass
+            return None
```
elesel, line 581: Originally this function would add the SAS token if the test above demonstrated that the container isn't public. I think you'll need to generate and add the token at this point.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -535,23 +534,27 @@ def __init__(self, storage_name: str, container_name: str, container_path: str):
             None,
         )
 
-    @property
-    def uri(self) -> str:
-        """Asset URI."""
-        # If we have already cached the URI, then simply return it
-        if self._uri is not None:
-            return self._uri
+    def get_uri(self, token_expiration_timedelta=datetime.timedelta(hours=1)) -> str:
```
elesel, line 537: Sorry, Kelly. What I meant was this, along with an accompanying import statement above:
```suggestion
    def get_uri(self, token_expiration: timedelta = timedelta(hours=1)) -> str:
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -535,23 +534,27 @@ def __init__(self, storage_name: str, container_name: str, container_path: str):
...
-        self._uri = f"{self._account_uri}/{self._container_name}/{self._container_path}"
+        uri = f"{self._account_uri}/{self._container_name}/{self._container_path}"
 
         # If a SAS token has been explicitly set, then assume that the URI+token is valid.
         # Simply append the token and return. If no SAS token is set, then proceed to rest of function.
         if self.token is not None:
-            self._uri += "?" + self.token
-            return self._uri
+            uri += "?" + self.token
+            return uri
```
elesel, line 557: This should handle the empty string:
```suggestion
            if self.token:
                uri += "?" + self.token
            return uri
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -565,7 +568,7 @@ def uri(self) -> str:
             )
 
             if response.status_code >= 200 and response.status_code <= 299:
-                return self._uri
+                return uri
```
elesel, line 571:
```suggestion
                self._token = ""
                return uri
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -609,23 +613,42 @@ def uri(self) -> str:
                 start=start_time
             )
 
-            self._uri += sas_token
-            return self._uri
+            sas_token = "?" + self._token
+            uri += sas_token
+            return uri
```
elesel, line 620:
```suggestion
            uri += "?" + self._token
            return uri
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -609,23 +613,42 @@ def uri(self) -> str:
...
+        """Container name."""
+        return self._container_name
+
     @property
     def token(self) -> str:
         """Sas token."""
         return self._token
 
+    @property
+    def account_uri(self) -> str:
```
elesel, line 650: Is this needed anymore?

scripts/azureml-assets/azureml/assets/get_tokens.py
```diff
@@ -0,0 +1,79 @@
...
+            continue
+
+        account_name = model_config.path.storage_name
+        container_name = model_config.path.container_name
+
+        if container_name in json_info[account_name]:
+            continue
+
+        _ = model_config.path.get_uri(token_expiration_timedelta=datetime.timedelta(days=1))
+        token = model_config.path.token
```
elesel, line 51: If token is the empty string, convert it to None before it goes into json_info so that in the tokens JSON file None means it's a public container. We don't need a mix of "" and None there when "" is really just used internally to the class to say "I've verified it's public" or maybe comment above) "I broke trying to generate a SAS token".

scripts/azureml-assets/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 ### 🐛 Bugs Fixed
 
+## 1.16.17 (2023-11-20)
```
elesel, line 6:
```suggestion
## 1.16.18 (2023-11-20)
```

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -535,23 +534,28 @@ def __init__(self, storage_name: str, container_name: str, container_path: str):
...
-    def uri(self) -> str:
-        """Asset URI."""
-        # If we have already cached the URI, then simply return it
-        if self._uri is not None:
-            return self._uri
+    def get_uri(self, token_expiration: timedelta = timedelta(hours=1)) -> str:
+        """Get Asset URI.
+
+        :param token_expiration_timedelta: Amount of time until token expiration
+        :type token_expiration_timedelta: datetime.timedelta
```
elesel, line 541:
```suggestion
        :param token_expiration: Amount of time until token expiration
        :type token_expiration: timedelta
```

elesel (approved):

PR #1763, opened 2023-11-16
Update embedding deployment validation.

shpengmsft:
There are 2 changes-
1. Will always retry with proxy URL when we hit HttpResponseError in first attempt to capture more failure reasons. because recent case shows resource "Not Found" error response. 
2. Use poxy resource (sub, rg, ws) to create new CognitiveServicesManagementClient and url template.


shpengmsft:
**scripts-syntax / Check syntax (pull_request)** failure is caused by existing code. (Someone checked in code without fixing all the failed check. :( )

tot0 (approved):

PR #1798, opened 2023-11-21
[Multimodal] Environment ver 2 with vulnerability fix

vivek-dani:
Create acft-multimodal-gpu environment version 2, with vulnerability fix for pyarrow

### DIFF
--- a/assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/Dockerfile
+++ b/assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/Dockerfile
@@ -7,3 +7,6 @@ RUN apt-get -y update
 # Install required packages from pypi
 COPY requirements.txt .
 RUN pip install -r requirements.txt --no-cache-dir
+
+# Vulnerability fix
+RUN pip install datasets==2.14.7
--- a/assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/requirements.txt
+++ b/assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/requirements.txt
@@ -5,7 +5,7 @@ azureml-acft-contrib-hf-nlp=={{latest-pypi-version}}
 azureml-acft-multimodal-components=={{latest-pypi-version}}
 # needed by multimodal
 mltable=={{latest-pypi-version}}
-datasets==2.12.0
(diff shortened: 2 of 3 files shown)


assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/requirements.txt
```diff
@@ -5,7 +5,7 @@ azureml-acft-contrib-hf-nlp=={{latest-pypi-version}}
 azureml-acft-multimodal-components=={{latest-pypi-version}}
 # needed by multimodal
 mltable=={{latest-pypi-version}}
-datasets==2.12.0
+datasets==2.14.7
```
rjaincc, line 8: If we have it in DockerFile, we can remove it from requirements.txt for now. 
Introduce it later once the dependencies are resolved.

rjaincc (approved):

assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/Dockerfile
```diff
@@ -6,4 +6,7 @@ RUN apt-get -y update
 
 # Install required packages from pypi
 COPY requirements.txt .
-RUN pip install -r requirements.txt --no-cache-dir
\ No newline at end of file
+RUN pip install -r requirements.txt --no-cache-dir
+
+# Vulnerability fix
+RUN pip install datasets==2.14.7
```
PhaniShekhar, line 12: Why do we need to install it here, if the correct version is added in requirements.txt

PhaniShekhar (approved):

assets/training/finetune_acft_multimodal/environments/acpt_multimodal/context/Dockerfile
vivek-dani: It'll get overwritten by azureml-metrics package. azureml-metrics uses lower version of datasets which has vulnerability

PhaniShekhar: Thanks for the explanation. Should we add a comment in the Dockerfile about this ?

PhaniShekhar (approved):

PR #1822, opened 2023-11-27
Add conda env for pytest in aml-benchmark

iamrk04:
There are a couple of issues (listed below) with the default asset-test workflow which we want to avoid and hence we are creating our own workflow to make sure new changes do not break the existing functionality:
- default asset-test workflow uses latest python version i.e. python==3.12, and some of the packages (`azureml-dataprep` and `azureml-dataprep-native`) that we depend upon do not have support for python==3.12, which simply breaks the workflow and we are unable to run our tests.
- If we change code in multiple components, it triggers the test multiple times (for multiple components), which is unnecessary. We just want the test to be triggered once as all the tests will be executed for any change anywhere. (Please note we want to run all the tests for any change in the code anywhere, not just for change in component but also for change in test)

**Update: To address the issue, we are now leveraging the conda_environment property for pytest.**


assets/aml-benchmark/components/batch-benchmark-inference/asset.yaml
```diff
@@ -3,6 +3,6 @@ spec: spec.yaml
 categories: ["Benchmarking"]
 test:
```
vizhur, line 4: I'd remove test section completely if you have own test pipeline.

.github/workflows/aml-benchmark-test.yaml
elesel, line 1: Why doesn't the assets-test workflow do what you need?

vizhur (changes requested):
lets not duplicate CIs and make improvements to the current one

.github/workflows/aml-benchmark-test.yaml
elesel: #1945 I believe removes the need for this PR.

iamrk04: Thanks @elesel. Is there a way to also control the python version that is used to run the tests?

elesel: There is now: #1964

iamrk04: thanks @elesel, I have made the changes accordingly.

PR #1824, opened 2023-11-27
Add azureml base image metadata in mlflow model

gauravrajguru:


rjaincc (approved):

assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -132,6 +133,11 @@ inputs:
       Pip dependencies expressed as below. Do not use quotes for passing.
       eg: pkg1==1.0, pkg2, pkg3==1.0
     optional: true
+  
+  inference_base_image:
+    type: string
+    description: The azureml base image to use in model inference. This ACR id is added in metadata of Mlmodel (mlflow).
```
novaturient95, line 139: Does it have to be azureml? What base image signify? In this case if we add image metadata it would be the one that will get used completely? Could we add description to fit it generically and also add details what all it would do.

gauravrajguru: The `inference_base_image` input is added under key "azureml.base_image". The key is used by the deployment service. It has to be azureml image and meant to be used as base image as suggested by the key. 
But the environment, [foundation-model-inference](https://github.com/Azure/azureml-assets/tree/main/assets/training/model_management/environments/foundation-model-inference), can do inferencing alone. Because, it uses online deployment hack by setting scoring script in env `AZUREML_ENTRY_SCRIPT` because their service use this env to find scoring script name.

novaturient95 (approved):

assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
gauravrajguru: Thanks Ayush, Updated the description.

novaturient95 (approved):

novaturient95 (approved):

PR #1827, opened 2023-11-27
Exposing mini batch size for the batch score step

yuanzhuangyuanzhuang:
Exposing mini batch size for the batch score step

### DIFF
--- a/assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
+++ b/assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
@@ -141,7 +141,7 @@ jobs:
   # Preparer
   batch_inference_preparer: 
     type: command
-    component: azureml:batch_inference_preparer:0.0.4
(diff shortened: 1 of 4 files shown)


arun-rajora (approved):

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
```diff
@@ -136,7 +141,7 @@ jobs:
   # Preparer
   batch_inference_preparer: 
     type: command
-    component: azureml:batch_inference_preparer:0.0.3
+    component: azureml:batch_inference_preparer:0.0.4
```
iamrk04, line 144: we can keep the version `0.0.3` for all of the batch components as we have `0.0.2` in `azureml` registry.

yuanzhuangyuanzhuang: all changed back

iamrk04 (approved):

PR #1873, opened 2023-12-04
update path to container for nlp model

sharma-riti:


assets/models/system/automl-text-classification/spec.yaml
```diff
@@ -48,4 +47,4 @@ tags:
     apply_deepspeed: 'true'
```
sharma-riti, line 47: @yiyoulin can we verify these for nlp runtime? ignore if these are already verified...

bhimar (approved):
please make sure that min inference sku is verified and accurate

assets/models/system/automl-text-classification/spec.yaml
yiyoulin: discussed offline for this.

yiyoulin (dismissed):

yiyoulin (approved):

PR #1885, opened 2023-12-06
Model Management Vulnerabilities resolution

HrishikeshGeedMS:


assets/common/environments/python-sdk-v2/context/conda_dependencies.yaml
```diff
@@ -13,4 +13,5 @@ dependencies:
     - azureml-telemetry=={{latest-pypi-version}}
     - cryptography=={{latest-pypi-version}}
     - certifi=={{latest-pypi-version}}
-    - urllib3=={{latest-pypi-version}}
\ No newline at end of file
+    - urllib3=={{latest-pypi-version}}
+
```
novaturient95, line 17: revert? Dont see any change here?

assets/common/environments/python-sdk-v2/context/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM mcr.microsoft.com/azureml/openmpi4.1.0-ubuntu20.04:{{latest-image-tag}}
 
-RUN apt-get update -y && apt-get install openssh-server openssh-client openssl curl -y
+RUN apt-get update -y && apt-get install libc-bin libc-dev-bin libc6 libc6-dev libcurl3-gnutls tar openssh-server openssh-client openssl curl -y
```
novaturient95, line 3: why are we updating libc?

novaturient95: Should not this be updated and resolved in the base image?

HrishikeshGeedMS: If we will not do it here, we will get 10 more vulnerabilites with latest image. Right now both the images are having 0 vuln with these changes in top of base image

PR #1904, opened 2023-12-07
fix evaluation logic for different metrics by name

VivienneTang:
1. add metrics name which need to be use "greater_than" to evaluate the breach.
2. fix the breach logic by metrics name


assets/model_monitoring/components/src/model_monitor_evaluate_metrics_threshold/evaluate_metrics_threshold.py
```diff
@@ -40,17 +64,31 @@ def evaluate_metrics_threshold(
...
+    return send_email_for_breached(metrics_threshold_breached_df, notification_emails, signal_name)
+
+
+def calculate_metrics_breach(metrics_threshold_df: pyspark.sql.DataFrame):
+    """Calculate the breached metrics by the given thresholds."""
+    metrics_threshold_breached_df = metrics_threshold_df.where(
+        (F.col(SIGNAL_METRICS_METRIC_NAME).isin(Metric_Value_Should_Greater_Than_Threshold) &
+         (F.col(SIGNAL_METRICS_METRIC_VALUE) < F.col(SIGNAL_METRICS_THRESHOLD_VALUE))) |
+        (~F.col(SIGNAL_METRICS_METRIC_NAME).isin(Metric_Value_Should_Greater_Than_Threshold) &
+         (F.col(SIGNAL_METRICS_METRIC_VALUE) > F.col(SIGNAL_METRICS_THRESHOLD_VALUE)))
```
yunjie-hub, line 79: how about == ?

assets/model_monitoring/components/src/model_monitor_evaluate_metrics_threshold/evaluate_metrics_threshold.py
```diff
@@ -3,8 +3,32 @@
...
+    SIGNAL_METRICS_METRIC_NAME,
+    SIGNAL_METRICS_METRIC_VALUE,
+    SIGNAL_METRICS_THRESHOLD_VALUE,
+)
 from shared_utilities.event_utils import post_warning_event, post_email_event
+import pyspark
+import pyspark.sql.functions as F
+
+
+Metric_Value_Should_Greater_Than_Threshold = [TWO_SAMPLE_KOLMOGOROV_SMIRNOV_TEST_METRIC_NAME,
```
yunjie-hub, line 24: Can we add the comments here, greater than threshold meaning it's valid and the rest of them is less than right?

assets/model_monitoring/components/src/model_monitor_evaluate_metrics_threshold/evaluate_metrics_threshold.py
VivienneTang: We do not need to consider the "equal" when evaluating the value of the metric towards the threshold. If the value equals the threshold, we consider it not breaching. This is also consistent with the UX criteria.

yunjie-hub (approved):

PR #1905, opened 2023-12-08
 Register Project Wednesday (Playground AYOD) RAG Model in assets

anushree1808:
Register Project Wednesday (Playground AYOD) RAG Model assets in the repo.


assets/promptflow/models/playground-ayod-rag/model.yaml
```diff
@@ -0,0 +1,8 @@
+path:
+  container_name: models
+  container_path: samples/playground_ayod_rag/v1/models
+  storage_name: wedpromptflowragsamples
```
vizhur, line 4: is this a public storage?

anushree1808: The storage has been assigned anonymous blob access as well as anonymous level permission is set at container level. Followed the instruction from [here](https://msdata.visualstudio.com/Vienna/_git/PromptFlow?path=/docs/sharing-your-flows-in-prompt-flow-gallery.md&version=GBgega/flowsharingexamples&_a=preview&anchor=4.1.1-uploading-your-flow-to-azure-blob-container).

vizhur: It might be better to have it private?

vizhur (changes requested):
lets use private storages

assets/promptflow/models/playground-ayod-rag/model.yaml
anushree1808: Hi @vizhur, I have updated the storage account access now:
- removed anonymous blob permission
- add permission for test-agent (storage blob reader)

image

vizhur (approved):

anushree1808:
> lets use private storages

PR #1909, opened 2023-12-08
Model spec validation

novaturient95:
Add model spec validation.

**Test:** 
Successful run: https://github.com/Azure/azureml-assets/actions/runs/7274124738/job/19819407084


scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -574,6 +595,188 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
...
+    if asset_config.type != assets.config.ModelType.MLFLOW:
+        logger.print(
+            f"Bypass validation for {asset_config.name} as model type is: {asset_config.type.value}"
+        )
+        return 0
+
+    # confirm must have tags
+    if not model.tags.get(MLFlowModelTags.TASK):
+        _log_error(asset_config.file_name_with_path, f"{MLFlowModelTags.TASK} missing")
+        return 1
```
lykelly19, line 610: Would it be possible to use something like error_count that many of the other validation methods are using? If the task tag is missing, this function would return that there's 1 error and not consider the rest of the checks in this function

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -574,6 +595,188 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
         return 1
 
 
+def validate_model_spec(asset_config: assets.AssetConfig, validated_model_map: dict):
+    """Validate model spec."""
```
lykelly19, line 599: Could you also add info about the args and return value in the comment please, similar to https://github.com/Azure/azureml-assets/blob/b9a80fd47e2bda0dd37f4666c49b2811cf7e4b08/scripts/azureml-assets/azureml/assets/validate_assets.py#L175C1-L182C8 ?

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -574,6 +595,188 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
...
+    supports_eval = False
+    supports_ft = False
+
+    if not model.tags.get(MLFlowModelTags.EVALUATION_COMPUTE_ALLOWLIST):
+        supports_eval = True
+
+    if not model.tags.get(MLFlowModelTags.FINETUNE_COMPUTE_ALLOWLIST):
+        supports_ft = True
+
+    # check properties
```
lykelly19, line 635: Looks like you have several areas like properties, valid computes, etc. which you could possibly separate into different functions

novaturient95: thanks for suggestion @lykelly19. I've moved common compute validations to a new method.

scripts/azureml-assets/azureml/assets/validate_assets.py
novaturient95: updated. Validating few scenarios

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -126,6 +126,15 @@ class PublishVisibility(Enum):
     UNLISTED = 'unlisted'
 
 
+class SystemRegisty:
+    """Prod System registries where assets lie."""
+
+    AZUREML = "azureml"
+    AZUREML_META = "azureml-meta"
+    AZUREML_MSR = "azureml-msr"
+    NVIDIA_AI = "nvida-ai"
```
lykelly19, line 135:
```suggestion
    NVIDIA_AI = "nvidia-ai"

```

lykelly19, line 135: Just wanted to check if this is a typo?
```suggestion
    NVIDIA_AI = "nvidia-ai"

```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -38,15 +40,20 @@
 # model validations
 MODEL_VALIDATION_RESULTS_FOLDER = "validation_results"
 VALIDATION_SUMMARY = "results.json"
+SUPPORTED_INFERNCE_SKU_FILE_NAME = "supported_inference_skus.json"
```
lykelly19, line 43:
```suggestion
SUPPORTED_INFEReNCE_SKU_FILE_NAME = "supported_inference_skus.json"

```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -38,15 +40,20 @@
 # model validations
 MODEL_VALIDATION_RESULTS_FOLDER = "validation_results"
 VALIDATION_SUMMARY = "results.json"
+SUPPORTED_INFERNCE_SKU_FILE_NAME = "supported_inference_skus.json"
+SUPPORTED_INFERNCE_SKU_FILE_PATH = Path(__file__).parent / "config" / SUPPORTED_INFERNCE_SKU_FILE_NAME
```
lykelly19, line 44:
```suggestion
SUPPORTED_INFERENCE_SKU_FILE_PATH = Path(__file__).parent / "config" / SUPPORTED_INFERENCE_SKU_FILE_NAME

```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -564,16 +589,148 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
...
+    error_count += validate_model_scenario(
+        asset_config.file_name_with_path,
+        model,
+        MLFlowModelProperties.INFERENCE_MIN_SKU_SPEC,
+        MLFlowModelProperties.INFERENCE_RECOMMENDED_SKU,
+        MLFlowModelTags.INFERENCE_COMPUTE_ALLOWLIST,
+    )
+
+    # check valid computes for inference
+    with open(SUPPORTED_INFERNCE_SKU_FILE_PATH) as f:
```
lykelly19, line 694:
```suggestion
    with open(SUPPORTED_INFERENCE_SKU_FILE_PATH) as f:

```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -430,12 +440,22 @@ def validate_tags(asset_config: assets.AssetConfig, valid_tags_filename: str) ->
...
+    validated_asset_config: assets.AssetConfig
+) -> int:
+    """Compare latest model with validation results.
+
+    Args:
+        latest_asset_config (assets.AssetConfig): asset config for latest model
+        validated_asset_config (assets.AssetConfig): asset cofig for validated model
+
+    Returns:
+        int: error_count
```
lykelly19, line 454:
```suggestion
        int: Number of errors.

```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -564,16 +589,148 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
...
+
+    Args:
+        asset_file_name_with_path (Path): file path to model asset
+        model (Model): model loaded from spec
+        min_sku_prop_name str: min sku property name for the scenario
+        recommended_skus_prop_name str: recommended sku property name for the scenario
+        compute_allowlist_tags_name str: compute allowlist tag name for the scenario
+
+    Returns:
+        int: error count
```
lykelly19, line 619:
```suggestion
        int: Number of errors.

```

lykelly19, line 619:
```suggestion
        min_sku_prop_name (str): min sku property name for the scenario
        recommended_skus_prop_name (str): recommended sku property name for the scenario
        compute_allowlist_tags_name (str): compute allowlist tag name for the scenario

    Returns:
        int: Number of errors.

```

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -564,16 +589,148 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
...
         return 1
 
 
+def validate_model_scenario(
+    asset_file_name_with_path: Path,
+    model: Model,
+    min_sku_prop_name: str,
+    recommended_skus_prop_name: str,
+    compute_allowlist_tags_name: str,
+) -> bool:
```
lykelly19, line 608: Should this function be returning a bool (to count this as one check) or an int (error count)?

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -470,17 +491,21 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
...
             if latest_model_path_uri != validated_model_path_uri:
                 logger.log_warning(f"latest_model_config_path_uri: [{latest_model_path_uri}]")
                 logger.log_warning(f"validated_model_config_path_uri: [{validated_model_path_uri}]")
+                error_count += 1
 
             if latest_model_config.description != validated_model_config.description:
                 logger.log_warning("Description does not match, for latest and validated asset")
+                error_count += 1
 
             return 1
```
lykelly19, line 510: I think this line needs to be changed because this would disregard all the error_counts in this if-statement

scripts/azureml-assets/azureml/assets/validate_assets.py
```diff
@@ -470,17 +491,21 @@ def validate_model_assets(latest_asset_config: assets.AssetConfig, validated_ass
             if latest_model_config.type != validated_model_config.type:
                 logger.log_warning(f"latest_model_config_type: [{latest_model_config.type}]")
                 logger.log_warning(f"validated_model_config_type: [{validated_model_config.type}]")
```
lykelly19, line 493: Would it be possible to make clear that the type does not match in the warning message (similar to what you did with the description warning in which you said, "Description does not match, for latest and validated asset"). For example, maybe something like: 

```python
logger.log_warning(f"latest_model_config_type: {latest_model_config.type} does not match validated_model_config_type: {validated_model_config.type}")
```

scripts/azureml-assets/azureml/assets/validate_assets.py
elesel, line 44: I'd suggest that SUPPORTED_INFERENCE_SKU_FILE_NAME provide the path relative to the config.py file (config/supported_inference_skus.json) rather than hard coding it here.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -126,6 +126,15 @@ class PublishVisibility(Enum):
     UNLISTED = 'unlisted'
 
 
+class SystemRegisty:
+    """Prod System registries where assets lie."""
```
elesel, line 130: The use of "lie" here feels odd to me.
```suggestion
    """Prod system registries."""
```

scripts/azureml-assets/azureml/assets/validate_assets.py
elesel, line 1: Please add some unit tests for these changes.

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -126,6 +126,15 @@ class PublishVisibility(Enum):
     UNLISTED = 'unlisted'
 
 
+class SystemRegisty:
```
vizhur, line 129: why do we need this and expose the subset of the registries here?

vizhur (changes requested):
lets chat on Teams

scripts/azureml-assets/azureml/assets/validate_assets.py
novaturient95: Thanks Kelly. Removed return stmt

scripts/azureml-assets/azureml/assets/validate_assets.py
novaturient95: Done

scripts/azureml-assets/azureml/assets/config.py
novaturient95: I wanted to check a few things. there have been errors around model version. And if they get checked at workflow that would be great instead of mistakenly creating a drop and then creating a release and then understanding _"whoooa I did not update the model version."_

Anyway, there was issue with credential here. I would probably add that as separate PR and a gate that validates version for components and models and potentially env which are not auto versioned.

novaturient95: @vizhur these are all public registries. Do you see any issue with specifying it here and cross-referencing details with what asset details in any of these registries?

scripts/azureml-assets/azureml/assets/validate_assets.py
novaturient95: Done

scripts/azureml-assets/azureml/assets/validate_assets.py
novaturient95: Thanks, updated

scripts/azureml-assets/azureml/assets/config.py
novaturient95: I've removed registry names as not needed in this PR. Will start discussion on teams for version check

scripts/azureml-assets/azureml/assets/validate_assets.py
novaturient95: one problem I see with current UT is that it refers to pypi version of azureml-assets and not the one in PR

novaturient95: added assets that would fail

vizhur (dismissed):

novaturient95:
Successful run: https://github.com/Azure/azureml-assets/actions/runs/7274124738/job/19819407084

vizhur (approved):

PR #1928, opened 2023-12-13
Allow python version customization

nick863:
**Problem:** Currently our test system will take the default version of python, used by system, however, this version may not be supported by some packages.
**Solution:** Add new configuration variable to allow version other then the default system.


vizhur (changes requested):
lets not default in the package itself

PR #1972, opened 2023-12-17
Version up bert-large-cased

novaturient95:
update version for bert-large-cased to allow property change as suggested in PR #1970


BhavanaTumma (approved):
Approve

PR #1978, opened 2023-12-18
[Finetune] Script Changes to disable ds3 for phi models

Anubha98:
Results and comparison is captured to state that stage 3 is giving bad results: https://dev.azure.com/msdata/Vienna/_workitems/edit/2824120


jpmann:
PR needs to be merged to azuremlcli first

PR #1982, opened 2023-12-18
Add tag to identify run completed with drift

shreeyaharma:
For customers to identify the RunCompleted event related to model monitoring drift.


assets/model_monitoring/components/src/model_monitor_evaluate_metrics_threshold/evaluate_metrics_threshold.py
```diff
@@ -51,6 +50,10 @@ def evaluate_metrics_threshold(
         post_warning_event(error_message)
         if notification_emails is not None and notification_emails != "":
             post_email_event(signal_name, notification_emails, error_message)
+        add_tags_to_run(_get_run_id(),
+                        {
+                            f"azureml_modelmonitor_threshold_breached": error_message
```
yunjie-hub, line 55: Did you check with PM if this key is good or not? Currently, the key doesn't have any signal information.

assets/model_monitoring/components/src/model_monitor_evaluate_metrics_threshold/evaluate_metrics_threshold.py
```diff
@@ -4,8 +4,7 @@
 """This file contains the core logic for data drift evaluate metrics threshold component."""
 
 import pyspark.sql.functions as F
-from shared_utilities.event_utils import post_warning_event, post_email_event
-
+from shared_utilities.event_utils import post_warning_event, post_email_event, add_tags_to_run, _get_run_id
```
yunjie-hub, line 7: You may also need to upgrade the component version as well as the signal component version which would use it

shreeyaharma: sure let me do that

yunjie-hub (dismissed):

yunjie-hub (approved):

PR #1991, opened 2023-12-19
Updating component to use latest environment

vikasagrawal-ms:
Updating component to use latest environment

### DIFF
--- a/assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
+++ b/assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
@@ -1,7 +1,7 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 
 name: convert_model_to_mlflow
-version: 0.0.17
+version: 0.0.18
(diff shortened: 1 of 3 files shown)


assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -1,15 +1,15 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 
 name: convert_model_to_mlflow
-version: 0.0.17
+version: 0.0.18
```
novaturient95, line 4: not needed I guess? can you please compare with components in prod?

vikasagrawal-ms: Environment version upgrade would cause version upgrade, right?

vikasagrawal-ms: got it what you meant

BhavanaTumma (approved):

PR #1998, opened 2023-12-20
Model Evaluation GA - Upgrade to 0.0.21

sarthaks95:


assets/training/model_evaluation/components/evaluate_model/README.md
```diff
@@ -6,7 +6,7 @@ evaluate_model
 
 ### Version 
 
-0.0.20
+0.0.19
```
guptha23, line 9: we are reducing the version of the component, can you please check if this is expected? (also, the version of the pipeline components and individual components is different)

sarthaks95: It is. Evaluate is not being released for GA.

guptha23: can you please check if we still want to reduce the version and if it doesn't create any version conflict?

guptha23 (approved):

ghyadav (approved):

PR #2002, opened 2023-12-20
updave vision model spec

gauravrajguru:


assets/models/system/bytetrack_yolox_x_crowdhuman_mot17-private-half/spec.yaml
```diff
@@ -3,19 +3,56 @@ name: bytetrack_yolox_x_crowdhuman_mot17-private-half
...
+      Standard_NC24ads_A100_v4,
+      Standard_NC48ads_A100_v4,
+      Standard_NC64as_T4_v3,
+      Standard_NC8as_T4_v3,
+      Standard_NC96ads_A100_v4,
+      Standard_ND40rs_v2,
+      Standard_ND96amsr_A100_v4,
+      Standard_ND96asr_v4
+    ]
+version: 4
```
novaturient95, line 58: add new line

novaturient95, line 58: if version was incremented before and model is not yet published, then we can revert back version change

novaturient95 (approved):

assets/models/system/bytetrack_yolox_x_crowdhuman_mot17-private-half/spec.yaml
novaturient95: checked, its 3 in prod

PR #2067, opened 2024-01-04
Updating model card for microsoft-phi-2

Anubha98:
1. changes to finetune config to remove hf_predict_module as phi_predict - to let it use the default predict method.
2. config.json - add eos_token
uploaded to registry link : https://ml.azure.com/registries/azureml-preview-test1/models/microsoft-phi-2/version/10?tid=72f988bf-86f1-41af-91ab-2d7cd011db47#artifacts

Run without hf_predict module: https://ml.azure.com/runs/0bfc52fe-58ee-418c-b2c5-6b10400c11af?wsid=/subscriptions/ed2cab61-14cc-4fb3-ac23-d72609214cfd/resourceGroups/training_rg/providers/Microsoft.MachineLearningServices/workspaces/finetune_dev_southcentralus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

Run without script is giving the expected and similar results
problem: with external script phi_predict.py, there were lot of restrictions regarding return_full_text = false property, fixed max_length, which was not giving right FT results.

### DIFF
--- a/assets/models/system/microsoft-phi-2/model.yaml
+++ b/assets/models/system/microsoft-phi-2/model.yaml
@@ -1,6 +1,6 @@
 path:
   container_name: models
-  container_path: huggingface/microsoft-phi-2/1702485256/mlflow_model_folder
+  container_path: huggingface/microsoft-phi-2/1704358138/mlflow_model_folder
   storage_name: automlcesdkdataresources
   type: azureblob
 publish:


assets/models/system/microsoft-phi-2/model.yaml
```diff
@@ -1,6 +1,6 @@
 path:
   container_name: models
-  container_path: huggingface/microsoft-phi-2/1702485256/mlflow_model_folder
+  container_path: huggingface/microsoft-phi-2/1704358138/mlflow_model_folder
```
novaturient95, line 3: 1. shared model in preview-registry has a lot of temp files (ipynb chk, .amlignore). Please make sure that they are not present in blob storage 
2. If code is not needed for base model inferencing or ft or eval, should we remove it altogether or is there any scenario for which we wanna keep it?

assets/models/system/microsoft-phi-2/model.yaml
```diff
@@ -1,6 +1,6 @@
...
   storage_name: automlcesdkdataresources
   type: azureblob
 publish:
```
novaturient95, line 6: Lets add version change in a separate PR

assets/models/system/microsoft-phi-2/model.yaml
novaturient95, line 3: what was the previous value of `eos_token_id` that has been updated to a new value now?

novaturient95: Will that affect any other scenario?

Anubha98: 1. ipynb extra folder is being deleted, but amlignore is even the part of msr published model  https://ml.azure.com/registries/azureml-msr/models/microsoft-phi-2/version/5?tid=72f988bf-86f1-41af-91ab-2d7cd011db47#artifacts
2. Can't say on eval and base model inferencing either Bhavana/Ghanshyam can tell. I am taking care of FT specific benchmarking related changes.

Anubha98: it was missing from config.json, That will not affect other scenarios.

novaturient95: 1. amlignore should be removed 
2. when we update model path ideally, we should check all scenarios. Anyways, it will be checked as part of model release valdn process. So not a blocker

novaturient95: @BhavanaTumma could you check query 2 about code necessity?

novaturient95: Sounds good

Anubha98: removed amlignore and phi_predict.

novaturient95 (approved):

PR #2072, opened 2024-01-04
update model performance interface for one input dataset

VivienneTang:
update model performance interface for one input dataset
1. make baseline data target column and production data target column to be required
2. remove baseline_data as it is the same as the production_data to reduce the data read


assets/model_monitoring/components/src/shared_utilities/momo_exceptions.py
```diff
@@ -4,7 +4,15 @@
...
-class DataNotFoundError(Exception):
+class UserError(Exception):
+    """Exception raised when it is user error."""
+
+    def __init__(self, message):
+        """Initialize a DataNotFoundError."""
+        super().__init__(f'UserError: {message}')
+
+
+class DataNotFoundError(UserError):
```
yunjie-hub, line 15: Do you have any job with this error shown? we want to check how is the exception thrown and logged

assets/model_monitoring/components/model_performance/model_performance_compute_metrics/spec.yaml
```diff
@@ -20,19 +20,14 @@ inputs:
...
     ]
     description: "Task type"
-  baseline_data:
-    type: mltable
-    mode: direct
   baseline_data_target_column:
     type: string
-    optional: true
     description: "Column name which contains ground truths in provided input for baseline_data. (Optional if we have one column name.)"
   production_data:
```
yunjie-hub, line 26: so this is not the production data but the joined data right? can we add some description?

VivienneTang: Added descriptions.

assets/model_monitoring/components/src/shared_utilities/momo_exceptions.py
VivienneTang: After checking the job, the user error does not help for telemetry. Removed this code change.

yunjie-hub (dismissed):

yunjie-hub (approved):

PR #2076, opened 2024-01-05
Updating phi2 version update

Anubha98:
corrosponding PR for model card change: https://github.com/Azure/azureml-assets/pull/2067


assets/models/system/microsoft-phi-2/spec.yaml
```diff
@@ -33,4 +33,4 @@ tags:
     ]  
   license: other
   task: text-generation
-version: 6
\ No newline at end of file
+version: 7
```
novaturient95, line 36: can you add a new line at end?

Anubha98: Added

novaturient95 (approved):

PR #2079, opened 2024-01-05
Ayushmishra/validate min SKU

novaturient95:


.github/workflows/assets-validation.yaml
```diff
@@ -27,6 +27,10 @@ defaults:
   run:
     shell: bash
 
+permissions:
+  # Required for OIDC login to Azure
+  id-token: write
+
 jobs:
   validate:
     name: Validate
```
lykelly19, line 36: below the line "runs-on: ubuntu-latest", add this line which should solve the access issue
```
    environment: Testing
```

.github/workflows/assets-validation.yaml
```diff
@@ -47,12 +51,28 @@ jobs:
...
-    
+
+      - name: print
+        run: echo $client_id $tenant_id $subscription_id
+        env: 
+          client_id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant_id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription_id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+  
+      - name: Log in to Azure
```
lykelly19, line 62: Could you put this log in section after install dependencies please?

.github/workflows/assets-validation.yaml
novaturient95: Thanks @lykelly19

.github/workflows/assets-validation.yaml
novaturient95: removed

PR #2096, opened 2024-01-09
Update azureml-metrics package to 0.0.42

ghyadav:


assets/model_monitoring/components/model_performance/model_performance_compute_metrics/spec.yaml
```diff
@@ -66,7 +66,7 @@ conf:
     dependencies:
       - python=3.8
       - pip:
-        - azureml-metrics~=0.0.38
+        - azureml-metrics[tabular]~=0.0.42
```
VivienneTang, line 69: In line 74, I added azureml-metrics[promptflow] to quick fix the "no module found" issue. Can you check with the 0.0.42, do we still need the azureml-metrics[promptflow] ? if not, could you please remove that line. Also, could you please update the unittest requirements.txt?

VivienneTang (approved):

PR #2100, opened 2024-01-09
Create correct histogram buckets when data column has a single distinct value

alanpo1:
local e2e test: https://ml.azure.com/experiments/id/15d0d2a8-d3fb-41ef-8609-72af4968f196/runs/elated_battery_dg8cl01pk5?wsid=/subscriptions/5fca341e-4ec3-45bd-b006-a7648d9376c5/resourcegroups/modelmonitoring-e2e-rg/providers/Microsoft.MachineLearningServices/workspaces/momo-e2e-test-ws-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/tests/e2e/resources/mltable_iris_baseline_int_data_type/iris_baseline.csv
```diff
@@ -11,4 +11,4 @@ sepal_length,sepal_width,petal_length,petal_width,target
 4,3,1,5,virginica
 5,3,1,6,versicolor
 4,3,1,5,virginica
-5,3,1,6,versicolor
\ No newline at end of file
+5,3,1.0,6,versicolor
```
ycheng35xo, line 14: This is already the single value test file? Why adding a new one and changing this one?

assets/model_monitoring/components/src/shared_utilities/histogram_utils.py
```diff
@@ -43,6 +43,12 @@ def get_dual_histogram_bin_edges(
 
         bin_width = (max_value - min_value) / num_bins
 
+        # If histogram has only one value then we only need a single bucket and
+        # should skip the for-loop below.
+        if min_value == max_value:
```
ycheng35xo, line 48: Maybe add UT for this as well?

assets/model_monitoring/components/tests/e2e/resources/mltable_iris_baseline_int_data_type/iris_baseline.csv
alanpo1: This is technically not the single-value test file because the corresponding production data (DATA_ASSET_IRIS_PREPROCESSED_MODEL_INPUTS_NO_DRIFT_INT_DATA) for this reference file doesn't have a single-value column and so the test doesn't check for the single distinct value histogram being built correctly. Also, I changed it because currently that column is being inferred as boolean and I don't think that was the intention of the testcase but I can change it back if we want.

perezbecker (approved):

assets/model_monitoring/components/src/shared_utilities/histogram_utils.py
```diff
@@ -43,6 +43,12 @@ def get_dual_histogram_bin_edges(
...
+            all_bin_edges[col] = [min_value, min_value * 1.01]
```
perezbecker, line 49: Not a big deal, but it might be slightly more intuitive for the user if the bin is centered around the single value. The easiest way to achieve this could be to compute a `delta = min_value*0.005` and then define `all_bin_edges[col] = [min_value-delta, min_value+delta]`

perezbecker, line 49: Also note that there is an edge-case for the edge-case. This approach will not work if `min_value==0`, as the bin_edges would remain [0,0]. If you use the `delta` approach above, you can check whether `delta==0`, in which case you force `delta=0.005` or another arbitrary number.

perezbecker, line 49: Another problematic case is when `min_value` is a negative number. For these cases, we need to modify the bin edge assignment, so that the smaller value is assigned first. Maybe the easiest approach that covers all scenarios I mentioned earlier is to define `delta = 0.005 if min_value == 0 else abs(min_value * 0.005)`. Please double-check my logic and let me know if something does not make sense.

alanpo1: I was just considering this case, I think your solution makes sense.

assets/model_monitoring/components/model_monitor/model_monitor_compute_histogram/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: spark
 name: model_monitor_compute_histogram
 display_name: Model Monitor - Compute Histogram
 description: Compute a histogram given an input data and associated histogram buckets.
-version: 0.3.4
+version: 0.3.5
```
ycheng35xo, line 7: Did you make changes in the compute_histogram component? I thought you only changed the compute_histogram_bucket? Let me know if it is not the case.

assets/model_monitoring/components/tests/unit/test_histogram_utils.py
```diff
@@ -0,0 +1,108 @@
...
+            assert all_edges.get(col, None) is not None
+
+            if col == 'col1':
+                assert len(all_edges[col]) == 2
+            else:
+                assert len(all_edges[col]) == self._num_bins_by_struges_algorithm(baseline_df) + 1
+
+            calculate_distinct_values_df = pd.DataFrame({col: all_edges[col]})
+            distinct_df = self.init_spark().createDataFrame(calculate_distinct_values_df)
+            assert distinct_df.distinct().count() == len(all_edges[col])
```
ycheng35xo, line 103: can you also verify the number are correct when the length is 2? i.e. the min-0.005 and min+0.005 value you set.

assets/model_monitoring/components/model_monitor/model_monitor_compute_histogram/spec.yaml
alanpo1: True, my changes affected histogram utils but not necessarily the ones used by compute histograms so I think we can revert.

ycheng35xo (approved):

assets/model_monitoring/components/tests/unit/test_histogram_utils.py
```diff
@@ -0,0 +1,207 @@
...
+                assert all_edges[col][0] == (min_value - expected_delta)
+                assert all_edges[col][1] == (min_value + expected_delta)
+            else:
+                assert len(all_edges[col]) == self._num_bins_by_struges_algorithm(baseline_df) + 1
+
+            calculate_distinct_values_df = pd.DataFrame({col: all_edges[col]})
+            distinct_df = self.init_spark().createDataFrame(calculate_distinct_values_df)
+            assert distinct_df.distinct().count() == len(all_edges[col])
+
+    def test_get_dual_histogram_bin_edges_single_distinct_value_bucket_negative(self):
```
ycheng35xo, line 110: seems most part of the codes are duplicated. Maybe it is better to share the common logic, setting a parameter (positive, zero, negative) as 3 different test cases. (You can do this later)

PR #2108, opened 2024-01-10
raise InvalidInputError for data asset url.

RichardLi1437:
Raise InvalidInputError for:
1. data asset url/id
2. wasbs/abfss url


assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
```diff
@@ -190,10 +190,15 @@ def _set_properties(self, ws: Workspace):
             self.path = matches.group("path").strip("/")
             self._datastore = None
         elif url.scheme == "azureml":
-            if ':' in url.path:  # azureml asset path
+            if ':' in url.path:  # azureml:<data_name>:<version> asset path
                 # asset path should be translated to azureml or hdfs path in service, should not reach here
                 raise InvalidInputError("AzureML asset path is not supported as uri_folder.")
-            else:  # azureml long or short form
+            elif '/data/' in self._base_url:  # azureml:// data asset url
```
alanpo1, line 196: Is this a safe check we can make for distinguishing invalid paths? What if the data uri is in a folder called data?

RichardLi1437: use regex to do the check

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
```diff
@@ -57,11 +57,14 @@ def _get_url(self, scheme=None, store_type=None, relative_path=None) -> str:
...
         """Get credential for this store url."""
-        def get_default_credential():
-            return DefaultAzureCredential() if self._is_secure() else None
-
         if not self._datastore:
-            return get_default_credential()
+            if self._is_secure():
+                raise InvalidInputError(
+                    "abfss, wasbs and https URLs are credential less and thus NOT supported for Model Monitoring job, "
+                    "please use credential data instead.")
```
alanpo1, line 64: Is there any instructions we can give to user for how to add credential to data or is the only option to choose a different data asset with credential?

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
```diff
@@ -190,10 +193,16 @@ def _set_properties(self, ws: Workspace):
...
+            if ':' in url.path:  # azureml:<data_name>:<version> asset path
                 # asset path should be translated to azureml or hdfs path in service, should not reach here
                 raise InvalidInputError("AzureML asset path is not supported as uri_folder.")
-            else:  # azureml long or short form
+
+            data_pattern0 = r"azureml://(subscriptions/([^/]+)/resourceGroups/([^/]+)/workspaces/([^/]+)/)?data/([^/]+)/versions/(.+)"  # noqa: E501
+            data_pattern1 = r"azureml://locations/([^/]+)/workspaces/([^/]+)/data/([^/]+)/versions/(.+)"
+            if re.match(data_pattern0, self._base_url) or re.match(data_pattern1, self._base_url):
+                raise InvalidInputError("AzureML data asset url is NOT supported as input of Model Monitoring job, "
+                                        "please use credential datastore + path instead.")
```
alanpo1, line 204: Is there some more clear instruction we can give? Maybe just tell them to use short form (`azureml:<name>:<version>`) or long form since we know that should work. I think as a user if I got this error on my MoMo job I might not know how to "... use credential datastore + path ..." as input

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
RichardLi1437: For now, we only support credential data. In the future, we need to discuss with PM whether and how we want to support credential less data.

yunjie-hub, line 64: Please use azureml file system path with a credential data store

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
```diff
@@ -57,11 +57,15 @@ def _get_url(self, scheme=None, store_type=None, relative_path=None) -> str:
...
-            return DefaultAzureCredential() if self._is_secure() else None
-
         if not self._datastore:
-            return get_default_credential()
+            if self._is_secure():
+                raise InvalidInputError(
+                    "abfss, wasbs and https URLs are credential less and are NOT supported by Model Monitoring job, "
+                    "please use azureml file system path with credential data store, e.g. "
+                    "azureml://datastores/<credential_datastore_name>/paths/<path_to_data>")
+            else:
```
yunjie-hub, line 66: What's this case?

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
```diff
@@ -190,10 +194,19 @@ def _set_properties(self, ws: Workspace):
...
-            else:  # azureml long or short form
+
+            data_pattern0 = r"azureml://(subscriptions/([^/]+)/resource[gG]roups/([^/]+)/workspaces/([^/]+)/)?data/(?P<data>[^/]+)/versions/(?P<version>.+)"  # noqa: E501
+            data_pattern1 = r"azureml://locations/([^/]+)/workspaces/([^/]+)/data/(?P<data>[^/]+)/versions/(?P<version>.+)"  # noqa: E501
+            matches = re.match(data_pattern0, self._base_url) or re.match(data_pattern1, self._base_url)
+            if matches:
+                raise InvalidInputError(
+                    f"You are using AzureML data url {self._base_url}, but this format of url is NOT supported as "
+                    "input of Model Monitoring job, please use "
+                    f"azureml:{matches.group('data')}:{matches.group('version')} instead.")
```
yunjie-hub, line 208: If the customer specifies this pattern, does it work?

I think the logic here is to verify if it's a V2 dataset

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
RichardLi1437: as the err msg indicates, it's the abfss, wasbs and https URLs, these URLs are essentially credential-less URL and we don't support for now.

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
RichardLi1437: As the comment in line 198, my understanding is service(data svc?) will translate data asset url to datastore+path or hdfs path, so although user provide azureml:<data>:<version>, Spark run time got datastore+path or hdfs path(abfss, wasbs).
I think we don't care if it's v1 or v2 dataset? The input is a url, if the url points to a credential v1 dataset, we can also support it?

alanpo1 (dismissed):

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
RichardLi1437: raise error for all abfs(s), wasb(s) and http(s) URL

assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
RichardLi1437: update error msg.

yunjie-hub (approved):

PR #2154, opened 2024-01-19
check for target_column in feature importance and throw UserError

alanpo1:
local e2e: https://ml.azure.com/experiments/id/15d0d2a8-d3fb-41ef-8609-72af4968f196/runs/ivory_pumpkin_wh1dkkq5yp?wsid=/subscriptions/5fca341e-4ec3-45bd-b006-a7648d9376c5/resourceGroups/modelmonitoring-e2e-rg/providers/Microsoft.MachineLearningServices/workspaces/momo-e2e-test-ws-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/tests/e2e/test_feature_attribution_drift_signal_monitor_e2e.py
```diff
@@ -175,3 +176,18 @@ def test_monitoring_run_use_defaults_empty_production_data_failed(
 
         # empty target data should cause the pipeline to fail
         assert pipeline_job.status == "Failed"
+
+    def test_monitoring_run_no_target_column_reference_data_successful(
+        self, ml_client: MLClient, get_component, test_suite_name
+    ):
+        """Test the scenario where the production data does not contain the target column."""
+        pipeline_job = _submit_feature_attribution_drift_model_monitor_job(
```
ycheng35xo, line 184: I would prefer not to add e2e for this scenario. UT will be better. Because the run time for gated test will be very long if we add a new e2e test for every change.

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
```diff
@@ -11,6 +11,7 @@
     StringType,
     FloatType
 )
+from shared_utilities.momo_exceptions import InvalidInputError
```
ycheng35xo, line 14: you need to change the yaml file as well

assets/model_monitoring/components/tests/e2e/test_feature_attribution_drift_signal_monitor_e2e.py
alanpo1: Yes that makes sense. We could maybe also look into parallelizing our e2e tests since most of them are independent from each other.

ycheng35xo: we can bring this to the scrum

alanpo1: I think the tests are already parallelized so ignore this comment

ycheng35xo (dismissed):

ycheng35xo (approved):

PR #2163, opened 2024-01-22
update the compute feature importances model monitoring component to use TreeExplainer to optimize execution time

imatiach-msft:
update the compute feature importances model monitoring component to use TreeExplainer to optimize execution time

This pull request involves significant changes to the `compute_feature_importance.py` file in the `assets/model_monitoring/components/src/feature_importance_metrics` directory. The changes include replacing the `get_model_wrapper` function with a call to `get_model`, renaming the function to `get_model`, and changing its functionality to create a lightgbm model instead of a model wrapper. The `compute_explanations` function was also modified to use the `TreeExplainer` instead of the `RAIInsights` class for calculating global feature importances.

Main changes:

* [`compute_feature_importance.py`](diffhunk://#diff-722964c327cbff02d1ff2dfa4cb1875410468efc9643ec8c336aae7abda69415L120-R122): Replaced the `get_model_wrapper` function with a call to `get_model`, renamed the function to `get_model`, and changed its functionality to create a lightgbm model instead of a model wrapper. [[1]](diffhunk://#diff-722964c327cbff02d1ff2dfa4cb1875410468efc9643ec8c336aae7abda69415L120-R122) [[2]](diffhunk://#diff-722964c327cbff02d1ff2dfa4cb1875410468efc9643ec8c336aae7abda69415L207-L212) [[3]](diffhunk://#diff-722964c327cbff02d1ff2dfa4cb1875410468efc9643ec8c336aae7abda69415L19-R19)
* `compute_feature_importance.py`: Modified the `compute_explanations` function to use the `TreeExplainer` instead

### DIFF
--- a/assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
+++ b/assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
@@ -16,10 +16,7 @@ from shared_utilities.io_utils import try_read_mltable_in_spark_with_error, save
 from shared_utilities.momo_exceptions import InvalidInputError
 from shared_utilities import constants
 from sklearn.model_selection import train_test_split
-from responsibleai import RAIInsights, FeatureMetadata
-from ml_wrappers.model.predictions_wrapper import (
-    PredictionsModelWrapperClassification,
-    PredictionsModelWrapperRegression)
+from interpret_community.shap.tree_explainer import TreeExplainer
 
 try:
     from lightgbm import LGBMClassifier, LGBMRegressor
@@ -107,8 +104,8 @@ def create_lightgbm_model(X, y, task_type):
     return model
 
 
-def get_model_wrapper(task_type, target_column, train_data):
-    """Create model wrapper using ml-wrappers on which to calculate feature importances.
+def get_model(task_type, target_column, baseline_data):
+    """Create a lightgbm model on which to calculate feature importances.
 
     :param task_type: The task type (regression or classification) of the resulting model
     :type task_type: string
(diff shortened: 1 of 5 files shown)


assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
```diff
@@ -175,19 +158,11 @@ def compute_explanations(model_wrapper, train_data, test_data, categorical_featu
...
-    feature_metadata = FeatureMetadata(categorical_features=categorical_features, dropped_features=[])
-
-    rai_i: RAIInsights = RAIInsights(
-        model_wrapper, train_data, test_data, target_column, task_type, feature_metadata=feature_metadata
-    )
-    log_time_and_message("Created RAIInsights")
-    # Add the global explanations using batching to allow for larger input data sizes
-    rai_i.explainer.add()
-    evaluation_data = train_data.drop([target_column], axis=1)
+    evaluation_data = test_data.drop([target_column], axis=1)
```
ycheng35xo, line 161: why test_data drop here? I think test data only has one column which is target column

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
```diff
@@ -16,10 +16,7 @@
 from shared_utilities.momo_exceptions import InvalidInputError
 from shared_utilities import constants
 from sklearn.model_selection import train_test_split
-from responsibleai import RAIInsights, FeatureMetadata
```
ycheng35xo, line 19: we have such function which manipulates datatype specific to lightgbm. Do we still need such logic or we need some other changes?  https://github.com/Azure/azureml-assets/blob/40f26e2caf91677621af794ebb14f54a238a2979/assets/model_monitoring/components/src/feature_importance_metrics/feature_importance_utilities.py#L40

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
```diff
@@ -16,10 +16,7 @@
...
-from ml_wrappers.model.predictions_wrapper import (
-    PredictionsModelWrapperClassification,
-    PredictionsModelWrapperRegression)
+from interpret_community.shap.tree_explainer import TreeExplainer
```
ycheng35xo, line 19: I feel it is a big change, can you upgrade the component version?

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
imatiach-msft: see line 139 in method get_train_test_data - we downsample the input dataset with line:

train_data, test_data = train_test_split(data, test_size=test_size, random_state=0)

we should have been using the test data but it looks like there was a mistake and train data was used instead.  Both train and test datasets in this case have feature columns and label column.

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
imatiach-msft: yes we still need this I believe, since we still train a lightgbm model on the dataset

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
imatiach-msft: done, incremented component version

ycheng35xo (approved):

assets/model_monitoring/components/src/feature_importance_metrics/compute_feature_importance.py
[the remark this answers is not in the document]
imatiach-msft: we currently train another model on the data (which is coming from some deployed model), and it is always a lightgbm model - so it shouldn't impact other scenarios

PR #2173, opened 2024-01-24
Throw user error when no common columns in compute data drift component

alanpo1:


assets/model_monitoring/components/prediction_drift/prediction_drift_signal_monitor/spec.yaml
```diff
@@ -104,7 +104,7 @@ jobs:
       type: aml_token
   compute_histogram_buckets:
     type: spark
-    component: azureml://registries/azureml/components/model_monitor_compute_histogram_buckets/versions/0.3.5
+    component: azureml://registries/azureml/components/model_monitor_compute_histogram_buckets/versions/0.3.6
```
ycheng35xo, line 107: you also need to upgrade the version for model_monitor_compute_histogram_buckets component

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
```diff
@@ -103,7 +103,7 @@ jobs:
       type: aml_token
   compute_drift_metrics:
     type: spark
-    component: azureml://registries/azureml/components/data_drift_compute_metrics/versions/0.3.10
```
ycheng35xo, line 106: also update data_drift_comput_metrics component

assets/model_monitoring/components/src/shared_utilities/df_utils.py
```diff
@@ -170,3 +182,51 @@ def add_value_if_present(
     if row_has_value(row, row_name):
         dict[target_property_name] = row[row_name]
     return dict
+
+
+def try_get_common_columns_with_warning(
```
ycheng35xo, line 187: when will the warning and ignore be used?

alanpo1: As of right now they are not used but added them for future use if we decide to throw warning in some situations instead of fail job. Or if we don't want to validate for empty common columns in new scenarios.

ycheng35xo (approved):

PR #2178, opened 2024-01-25
added maas-inference tag for mistral models

choudharykartik:
Added maas-inference: true tag for 4 mistral models

### DIFF
--- a/assets/models/system/Mistral-7B-Instruct-v0-1/spec.yaml
+++ b/assets/models/system/Mistral-7B-Instruct-v0-1/spec.yaml
@@ -11,6 +11,7 @@ tags:
   Featured: ""
   Preview: ""
   SharedComputeCapacityEnabled: ""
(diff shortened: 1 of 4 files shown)


novaturient95 (dismissed):

assets/models/system/Mistral-7B-Instruct-v0-1/spec.yaml
```diff
@@ -11,6 +11,7 @@ tags:
   Featured: ""
   Preview: ""
   SharedComputeCapacityEnabled: ""
+  MaaS-inference: true
```
novaturient95, line 14: lets update to all lower case as shared by @skasturi

choudharykartik: Done

novaturient95 (approved):

PR #2188, opened 2024-01-27
fix credential-less blob check

RichardLi1437:


assets/model_monitoring/components/src/model_data_collector_preprocessor/store_url.py
```diff
@@ -67,7 +67,7 @@ def get_credential(self) -> Union[str, ClientSecretCredential, AzureSasCredentia
                 return self._datastore.account_key
             elif self._datastore.credential_type == "Sas":
                 return AzureSasCredential(self._datastore.sas_token)
-            elif self._datastore.credential_type is None:
+            elif self._datastore.credential_type is None or self._datastore.credential_type == "None":
```
yunjie-hub, line 70: upgrade component version

RichardLi1437: updated with version upgrade tool.

ycheng35xo (approved):

assets/model_monitoring/components/tools/spec_version_upgrader.py
```diff
@@ -0,0 +1,173 @@
...
+
+if __name__ == "__main__":
+    # TODO get specs from arguments
+    argparser = ArgumentParser()
+    argparser.add_argument("--specs", help="spec names to upgrade")
+    args = argparser.parse_args()
+    specs = args.specs.split(",") if args.specs else []
+    spec_folder = os.path.abspath(f"{os.path.dirname(__file__)}/..")
+    spec_version_upgrader = SpecVersionUpgrader(spec_folder)
+    spec_version_upgrader.upgrade_versions(specs)
```
ycheng35xo, line 173: so we need to specify the component name(s)? If the name is not specified, it will upgrade all the components? What if the common utils is changed but not component itself changed? Can you add a .md file or add a comment with examples with usage?

assets/model_monitoring/components/tools/spec_version_upgrader.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+if __name__ == "__main__":
+    # TODO get specs from arguments
+    argparser = ArgumentParser()
+    argparser.add_argument("--specs", help="spec names to upgrade")
+    args = argparser.parse_args()
+    specs = args.specs.split(",") if args.specs else []
+    spec_folder = os.path.abspath(f"{os.path.dirname(__file__)}/..")
+    spec_version_upgrader = SpecVersionUpgrader(spec_folder)
```
ycheng35xo, line 172: why not passing the specs in the constructor? It is Todo for now?

assets/model_monitoring/components/tools/spec_version_upgrader.py
RichardLi1437: this tool figure out the subgraph/component dependency/reference, but you need to provide the first level component list.
I can make it to upgrade all component versions if no component provided.

assets/model_monitoring/components/tools/spec_version_upgrader.py
RichardLi1437: the essential for SpecVersionUpgrader is the spec dependencies/references, after spec_folder provided, the tool can figure that out. Once the dependencies is there, the instance can do the upgrade for different specs list, no need to construct the dependency everytime.

yunjie-hub (approved):

PR #2214, opened 2024-01-31
Flow creation- logging error and error handling.

shpengmsft:
[Bug 2930037](https://msdata.visualstudio.com/Vienna/_workitems/edit/2930037): [Promptflow Creation] json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)


assets/large_language_models/rag/components/src/flow_creation.py
```diff
@@ -300,123 +330,178 @@ def main(args, ws, current_run, activity_logger: Logger):
...
-        'properties': {'flow_id': flow_id}})
-    run_properties = current_run.get_properties()
-    parent_run_id = run_properties["azureml.pipelinerunid"]
-    parent_run = ws.get_run(parent_run_id)
-    parent_run.add_properties(
-        {"azureml.promptFlowResourceId": flow_id})
-    activity_logger.info("[Promptflow Creation]: Add into run property Succeed! All operation is done!", extra={
-        'properties': {'flow_id': flow_id}})
+    if "flowResourceId" not in pf_response_json:
+        activity_logger.error(
```
Adarsh-Ramanathan, line 469: Should we also throw here? As it stands, this looks like it'll mark the step as being successful; I'm not sure if that's the desired behavior.

assets/large_language_models/rag/components/src/flow_creation.py
```diff
@@ -300,123 +330,178 @@ def main(args, ws, current_run, activity_logger: Logger):
...
+            f"[Promptflow Creation]: Flow creation failed with Response Code: {response.status_code},"
+            + f" response:{response.text}."
+        )
+    else:
+        flow_id = pf_response_json["flowResourceId"]
```
Adarsh-Ramanathan, line 474: Do we care about this key not being found in this response dict?

liangjason87 (approved):

PR #2234, opened 2024-02-01
Humaneval add support for stripped responses.

arun-rajora:
Humaneval add support for stripped responses.

### DIFF
--- a/assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
+++ b/assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
@@ -176,7 +176,7 @@ def run_humaneval_postprocessor(
         else:
             original_prediction = row["original_prediction"]
             # If spaces were stripped from endpoint responses, add those back.
-            if original_prediction.startswith(" ") or original_prediction.startswith("\t"):
+            if not len(original_prediction) or (len(original_prediction) and original_prediction[0].isspace()):
                 prefix = ""
             else:
                 prefix = "    "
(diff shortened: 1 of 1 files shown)


assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
```diff
@@ -170,13 +170,17 @@ def run_humaneval_postprocessor(
...
-
         if str("def " + row["entry_point"] + "(") in row["original_prediction"]:
             # If the model regenerates the prompt/ function name
             pred_combined_prompt = row["original_prediction"]
         else:
-            pred_combined_prompt = row["prompt"]+"\n"+row["original_prediction"]
-
+            original_prediction = row["original_prediction"]
+            # If spaces were stripped from endpoint responses, add those back.
+            if not len(original_prediction) or (len(original_prediction) and original_prediction[0].isspace()):
```
SamGos93, line 179: What if the prediction starts with 3/5 spaces instead of 4 spaces? Can we have a scenario like that?

SamGos93 (approved):

assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
arun-rajora: Here we check only the first character. If it is some whitespace character, that means endpoint is not stripping them. Else, we add the whitespace.

If there are 3-5 whitespaces, we do nothing, prefix is "".

SamGos93: Yeah, I was wondering if in the future we get a scenario where the first sentence in the prediction is not indented correctly due to some reason which can cause the code to fail. The first sentence after the function declaration should generally start with "    ". So, in case it is not, we should check that and fix it. What do you think? As of now, we do not have this issue, so should be fine.

PR #2237, opened 2024-02-01
Update RAG sample flows to use Index Lookup tool

Adarsh-Ramanathan:
This PR does the following things:
1) Updates the RAG gallery samples to pull from RAG-Team owned `azuremlragsamples` instead of `promptflowsamples`
2) Updates the RAG samples so that they use the Index Lookup tool instead of legacy tools.


DaweiCai (dismissed):

assets/promptflow/models/bring-your-own-data-chat-qna/model.yaml
```diff
@@ -1,7 +1,7 @@
 path:
-  container_name: models
-  container_path: samples/bring_your_own_data_chat_qna/v6/model_folder
-  storage_name: promptflowsamples
```
dorisjoy, line 4: Please also change below to files to increase the model version by 1:
https://github.com/Azure/azureml-assets/blob/main/assets/promptflow/models/bring-your-own-data-chat-qna/spec.yaml
https://github.com/Azure/azureml-assets/blob/main/assets/promptflow/models/bring-your-own-data-qna/spec.yaml
image
change to 7

assets/promptflow/models/bring-your-own-data-chat-qna/model.yaml
```diff
@@ -1,7 +1,7 @@
...
+  container_name: gallery-samples
+  container_path: bring_your_own_data_chat_qna
```
dorisjoy, line 3: The container_path doesn't contain version info. If we found this version example has issue. Do we have place to find the previous working version ?

assets/promptflow/models/bring-your-own-data-chat-qna/model.yaml
Adarsh-Ramanathan: Done!

assets/promptflow/models/bring-your-own-data-chat-qna/model.yaml
Adarsh-Ramanathan: These examples are uploaded to this storage account by a release pipeline, from an ADO repo. Can we not use that as the version control mechanism here, instead of creating multiple versions within the storage account? Git retains history of changes in a more granular fashion, and will allow us to perform reverts more effectively if needed.

Perhaps I'm missing something, but what benefit are we gaining by versioning the samples in storage?

assets/promptflow/.cspell.json
```diff
@@ -66,7 +68,9 @@
       "zelin",
       "embeded",
       "metad",
-      "orignal"
+      "orignal",
```
lykelly19, line 71: Is this supposed to be "original"?

Adarsh-Ramanathan: Probably not. "original" would likely already be in the standard dictionary that the spellchecker uses. I suspect there's some resource or sample in this repo with a typo, and instead of correcting the typo, "orignal" got added to the dict here.

It's not a change that I made, and I don't understand enough about why it's here to change it, unfortunately.

lykelly19 (approved):

dorisjoy (approved):

PR #2266, opened 2024-02-06
SKU list for finetuning / evaluation updated for phi-models

HrishikeshGeedMS:
Since we do not have computes for all the instance type to check in runners, ideally we should add all the possible computes in spec file , to have successful validation.


assets/models/system/microsoft-phi-1-5/spec.yaml
```diff
@@ -44,4 +46,4 @@ tags:
     apply_ort: 'false'
     precision: 16
   task: text-generation
-version: 6
\ No newline at end of file
+version: 7
```
novaturient95, line 49: nit: add new line at the end of spec file

novaturient95 (dismissed):

skasturi (approved):

PR #2288, opened 2024-02-08
Segregate tests + Script to upgrade components

iamrk04:
This PR has the following changes:
- add script to automatically upgrade the required parts of components before kicking off release.
- segregate tests for all aml-benchmark components to parallelly run them, this will allow us to run all the tests in less time.
- improve running command in subprocess to raise the error in a better way, this eases the error message propagation and debugging process.


assets/aml-benchmark/scripts/_internal/upgrade_components.py
```diff
@@ -0,0 +1,214 @@
...
+            file.write(spec_str)
+    except Exception as e:
+        is_error = True
+        error = str(e)
+    return is_error, error, component_path, name
+
+
+def main():
+    """Entry function."""
+    components = _get_all_components_spec()
```
arun-rajora, line 164: Can we create flags to make upgrade environment, ignore resource_manager configurable? If possible, let us specify the environment version to pin or component version to pin to (instead of latest). Thanks!

iamrk04: Added argument env_version` to control env upgrade (more details in readme.md). All the versions are now pinned to a specific version instead of `latest` label. Ignoring components for release is something that we should control within the script, there is a global variable for it which can be changed. Not adding this as an argument at the moment, does not seem to be worth doing at the moment.

assets/aml-benchmark/components/batch-benchmark-inference/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: pipeline
 name: batch_benchmark_inference
 display_name: Batch Benchmark Inference
 description: Components for batch endpoint inference
-version: 0.0.5
+version: 0.0.6
```
arun-rajora, line 7: Can we undo these changes in the spec? Thanks!

arun-rajora (dismissed):

arun-rajora (approved):

PR #2289, opened 2024-02-08
update GSQ to use AzureMLOnBehalfOfCredential from azure-ai-ml

imatiach-msft:
update GSQ to use AzureMLOnBehalfOfCredential from azure-ai-ml


shreeyaharma (dismissed):
If you have time can you pin openai version

Advitya17 (dismissed):

imatiach-msft:
"If you have time can you pin openai version"
done!

PR #2291, opened 2024-02-08
Do None check before publishing metric

yunjie-hub:


assets/model_monitoring/components/src/model_monitor_output_metrics/entities/signals/signal.py
```diff
@@ -46,9 +46,10 @@ def publish_metrics(self, step: int):
             threshold = None
             if "threshold" in run_metric:
                 threshold = run_metric["threshold"]
-            publish_metric(
-                run_metric["runId"], float(run_metric["value"]), threshold, step
-            )
+            if "value" in run_metric and run_metric["value"]:
```
RichardLi1437, line 49: minor: can be simplified as
```py
if run_metric.get("value"):
```

RichardLi1437, line 49: just curious in what scenario we'll get a None metric?

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: pipeline
 name: data_drift_signal_monitor
 display_name: Data Drift - Signal Monitor
 description: Computes the data drift between a baseline and production data assets.
-version: 0.3.30
+version: 0.3.31
```
RichardLi1437, line 7: have you used by version upgrade tool?

RichardLi1437 (approved):

assets/model_monitoring/components/src/model_monitor_output_metrics/entities/signals/signal.py
yunjie-hub: Actually I don't know, I am asking customer

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
yunjie-hub: forget to run

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
```diff
@@ -129,4 +132,11 @@ def publish_metrics(run_id: str, metrics: dict, step: int):
...
     with mlflow.start_run(run_id=run_id, nested=True):
-        mlflow.log_metrics(metrics=metrics, step=step)
+        for _ in range(MAX_RETRY_COUNT):
+            try:
+                mlflow.log_metrics(metrics=metrics, step=step)
+                return
+            except Exception as e:
+                print(f"Exception occurred in publishing metrics: {traceback.format_exc()}")
+                time.sleep(1)
+
```
RichardLi1437, line 142: raise e if all retries fail?

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
```diff
@@ -136,7 +139,6 @@ def publish_metrics(run_id: str, metrics: dict, step: int):
             try:
                 mlflow.log_metrics(metrics=metrics, step=step)
                 return
-            except Exception as e:
-                print(f"Exception occurred in publishing metrics: {traceback.format_exc()}")
+            except Exception:
```
RichardLi1437, line 142: if the run fail because update status from failed to completed, seems we can also catch the exception here? If so, we don't need the outer part of try-except.

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
yunjie-hub: talked offline need catch the exception, otherwise it will affect the following runs

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
yunjie-hub: As we discussed, if we have this logic at the beginning, we should be good.
The reason we need to add outside, it's because there is already some failed run

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
```diff
@@ -103,6 +103,9 @@ def get_or_create_run_id(
         with mlflow.start_run(run_id=_get_or_create_parent_run_id(monitor_name)) as current_run:
             print(f"Current parent run id: {current_run.info.run_id}")
             run_name = f"{signal_name}_{metric_name}"
+            if MLFLOW_RUN_ID in os.environ:
```
RichardLi1437, line 106: this is to delete the MFLOW_RUN_ID env var if it exits, but why we want to do this?

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
RichardLi1437: yes, there are some failed run there, so the mlfow.log_metrics may fail because update run status from failed to completed, but seems this exception will be also caught here and be silence ignored after 3 retries, and continue with next metrics run outside, do we still need to catch exception outside?

RichardLi1437 (approved):

assets/model_monitoring/components/src/shared_utilities/run_metrics_utils.py
RichardLi1437, line 106: nit: better add comments here like: to enforce mlflow to use different runId for metrics calculation job and metrics tracking job.

PR #2303, opened 2024-02-09
Baseline data preprocessor failed and not baseline data passed to nex…

vaibhj:
…t step

### DIFF
--- a/assets/model_monitoring/components/src/model_monitor_feature_selector/run.py
+++ b/assets/model_monitoring/components/src/model_monitor_feature_selector/run.py
@@ -33,12 +33,13 @@ def run():
     if args.input_data_2 is not None:
         input_df2 = try_read_mltable_in_spark(args.input_data_2, "input_data_2")
 
+    # At least we need one input either input_df1 or input_df2 for feature selection.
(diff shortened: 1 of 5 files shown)


assets/model_monitoring/components/src/model_monitor_feature_selector/run.py
```diff
@@ -29,7 +29,9 @@ def run():
     args = parser.parse_args()
 
     input_df1 = try_read_mltable_in_spark(args.input_data_1, "input_data_1")
-    input_df2 = try_read_mltable_in_spark(args.input_data_2, "input_data_2")
+    input_df2 = input_df1
```
ycheng35xo, line 32: if input_df2 is none, then it use same value as input_df1? This is the logic change. If yes, maybe you can add this in line 41 and modify that comment.

vaibhj: This PR trying to cover none check for baseline data as its optional. if either input available feature selection will work and in feature selection input1 is required

ycheng35xo (approved):

PR #2312, opened 2024-02-12
convert missing column validation exceptions to user errors and add tests

imatiach-msft:
convert missing column validation exceptions to user errors and add tests for GSQ model monitoring component

Note that as part of this PR the github action for GSQ tests (several of which were added with this PR and others recently) was separated out from general model monitoring tests for several reasons:
1.) The total time for the model monitoring github action is becoming really long, I think it's 1-2 hours now and I was hitting timeouts sometimes, and it was taking a really long time to wait on PRs to see if they pass tests
2.) The dependencies and installation process is different for GSQ components
3.) It's better to separate out GSQ tests as it is a whole set of different components in general

### DIFF
--- a/.github/workflows/model-monitoring-gsq-ci.yml
+++ b/.github/workflows/model-monitoring-gsq-ci.yml
@@ -1,4 +1,4 @@
-name: model-monitoring-ci
+name: model-monitoring-gsq-ci
 
 on:
   pull_request:
@@ -6,7 +6,7 @@ on:
       - main 
     paths:
       - assets/model_monitoring/**
-      - .github/workflows/model-monitoring-ci.yml
+      - .github/workflows/model-monitoring-gsq-ci.yml
       
 
 concurrency:
@@ -32,7 +32,7 @@ jobs:
   check-execution-context:
     uses: Azure/azureml-assets/.github/workflows/check-execution-context.yaml@main
   run-momo-tests:
-    name: Run Monitoring Component Tests
+    name: Run GSQ Monitoring Component Tests
     runs-on: ubuntu-latest
     needs: check-execution-context
     environment: Testing
@@ -45,6 +45,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.8'
+
+      - name: Install fusepy
+        run: sudo apt-get install -y fuse
+
       - name: Log in to Azure and create resources
         uses: ./.github/actions/create-azure-resources
         with:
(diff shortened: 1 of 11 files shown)


vinuthakaranth (approved):

.github/workflows/model-monitoring-gsq-ci.yml
```diff
@@ -0,0 +1,105 @@
...
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          scripts-setup-dir: ${{ env.scripts_setup_dir }}
+
+      - name: Install dependencies
+        run: pip install -r ${{ env.testsRootPath }}/requirements.txt
+
+      - name: Upgrade pandas, numpy, numba
+        run: pip install pandas==2.0.3 numpy==1.24.4 numba==0.57.1
```
vizhur, line 64: why don't you have this in requirements.txt

imatiach-msft: done, moved to separate file gsq-requirements.txt (the requirements are actually very different between the two set of components)

yunjie-hub (approved):

qusongms (approved):

assets/model_monitoring/components/src/shared_utilities/event_utils.py
```diff
@@ -5,12 +5,18 @@
...
-from azureml.core.authentication import AzureMLTokenAuthentication
-from mlflow import MlflowClient
+import warnings
+try:
+    import mlflow
+    from azureml._restclient.models import BaseEvent
+    from azureml._restclient.run_client import RunClient
+    from azureml._restclient.service_context import ServiceContext
+    from azureml.core.authentication import AzureMLTokenAuthentication
+    from mlflow import MlflowClient
```
yunjie-hub, line 15: Are these required for the metrics calculation?

imatiach-msft: this isn't in the GSQ component, it's in some shared utilities to post messages for dpv2 jobs

assets/model_monitoring/components/pytest.ini
```diff
@@ -2,4 +2,5 @@
 [pytest]
 markers =
     e2e: mark a test as a full e2e test running on a live AML stack.
-    unit: mark a test as a unit test running in memory.
\ No newline at end of file
+    unit: mark a test as a unit test running in memory.
+    gsq_test: mark a test as a generation safety quality component test.
```
alanpo1, line 6: We don't have to do it this PR but we could consider using this new gsq_test marker for our genAI signals and component e2e tests too. (token statistics, genAI preprocessor, etc)

alanpo1:
Doesn't have to be done in this PR but since we're adding a new pytest marker we should consider using the gsq_test marker for all our genAI signals and component e2e tests too. (token statistics, genAI preprocessor, etc). It might help e2e test time to run the "traditional ML" and "genAI" component areas separately going forward as we add more and more components to each area.

assets/model_monitoring/components/src/shared_utilities/event_utils.py
imatiach-msft: ok I removed the imports I think it might work now that I've changed the GSQ CI to test specific files instead of the entire directory

yunjie-hub (approved):

Man-MSFT (approved):

lykelly19 (approved):

PR #2313, opened 2024-02-12
[benchmark] add shared code parsing utils

EricWrightAtWork:
Newer versions of GPT (1106, 0125) often respond to coding prompts with code in markdown tags - this breaks some existing post-processing/parsing logic for 1P (babel) and 3P. Added the following changes:

* Updated the HumanEval custom post-proc script to handle tagged responses


assets/aml-benchmark/tests/test_evaluation_utils.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Tests for common evaluation utilities."""
+
+import pytest
+
+from aml_benchmark.utils.evaluation_utils import extract_python_function, extract_text_from_markdown_tag
```
iamrk04, line 8: This import should fail if the package isn't installed, which mostly is the case. We can fix this for everyone by adding the following statements in `./assets/aml-benchmark/tests/__init__.py`:
```
import sys

from .test_utils import get_src_dir


path = get_src_dir()
if path not in sys.path:
    sys.path.append(path)

```

iamrk04 (changes requested):

assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
```diff
@@ -172,9 +174,9 @@ def run_humaneval_postprocessor(
         gt = "\n" + row["test"] + "\n" + "check(" + row["entry_point"] + ")"
         if str("def " + row["entry_point"] + "(") in row["original_prediction"]:
             # If the model regenerates the prompt/ function name
-            pred_combined_prompt = row["original_prediction"]
+            pred_combined_prompt = extract_text_from_markdown_tag(row["original_prediction"], tag_type='python')
```
arun-rajora, line 177: Should we also handle the scenario where model does not output python and only the backticks with the code in it (so tag_type is empty string)?

assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
```diff
@@ -15,6 +15,8 @@
 from datasets import load_dataset
 from typing import Any, Dict, List, Union
 
+from aml_benchmark.utils.evaluation_utils import extract_text_from_markdown_tag
```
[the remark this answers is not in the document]
EricWrightAtWork: Moved the utility - so, the import is gone.

assets/aml-benchmark/scripts/custom_inference_postprocessors/humaneval.py
EricWrightAtWork: This should already be covered by the function - though not used for that in this post-processor

iamrk04 (approved):

SamGos93 (approved):

PR #2314, opened 2024-02-13
Add GenAI Preprocessor logic

alanpo1:
sample e2e component test: https://ml.azure.com/experiments/id/15d0d2a8-d3fb-41ef-8609-72af4968f196/runs/modest_carpet_b58g2pwb8t?wsid=/subscriptions/5fca341e-4ec3-45bd-b006-a7648d9376c5/resourceGroups/modelmonitoring-e2e-rg/providers/Microsoft.MachineLearningServices/workspaces/momo-e2e-test-ws-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_run.py
```diff
@@ -3,4 +3,138 @@
...
+    """
+    df = df.withColumns(
+        {'end_time': df.end_time.cast(TimestampType()), 'start_time': df.start_time.cast(TimestampType())}
+    )
+
+    df = _promote_fields_from_attributes(df)
+
+    # Cast all remaining fields in attributes to string to make it Map(String, String),
+    # or make it json string, for easier schema unifying
+    df = _convert_complex_columns_to_json_string(df)
```
RichardLi1437, line 71: so your choice is to make it a json string, then please update the comment.

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_run.py
```diff
@@ -3,4 +3,138 @@
...
+def run():
+    """Compute data window and preprocess data from MDC."""
+    # Parse arguments
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--data_window_start", type=str)
+    parser.add_argument("--data_window_end", type=str)
+    parser.add_argument("--input_data", type=str)
+    parser.add_argument("--preprocessed_span_data", type=str)
+    parser.add_argument("--aggregated_trace_data", type=str)
+    args = parser.parse_args()
```
RichardLi1437, line 128: we should have a flag --require_trace_data, if it is set, we do the trace aggregation, otherwise just do the span processing. This is for some metrics it only require span log as input so we can skip trace aggregation to save resource and time.

assets/model_monitoring/components/model_monitor/genai_mdc_preprocessor/spec.yaml
RichardLi1437, line 1: we need 
```yaml
  require_trace_data:
    type; bool
    default: true
```
in `inputs`, if true, aggregate span to trace, if false, skip aggregation just output the span log. The trace output of this component is optional.
This is for some metrics that only require span as input, we can skip trace aggregation to save resource and time.
This is a P2 feature, so you can add this in a separate PR, but please don't miss it.

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_run.py
```diff
@@ -3,4 +3,138 @@
...
+    Args:
+        df: The raw span logs data in a dataframe.
+    """
+    def try_get_df_column(df: DataFrame, name: str):
+        try:
+            return df[name]
+        except AnalysisException:
+            return None
+
+    df = df.withColumns(
```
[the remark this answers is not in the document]
alanpo1: Yes, will do in later PR v1 will have all data.

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_run.py
```diff
@@ -3,4 +3,138 @@
...
+        {
+            key: df_column for key, col_name in _get_important_field_mapping().items()
+            if (df_column := try_get_df_column(df, col_name)) is not None
```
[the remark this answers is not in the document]
alanpo1: Yes I think your logic is correct. I suppose I should handle adding None column if we require it going forward. Will include in v2 PR.

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_run.py
alanpo1: added flag.

assets/model_monitoring/components/model_monitor/genai_mdc_preprocessor/spec.yaml
alanpo1: Will add in v2 PR.

ycheng35xo (approved):

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        self._load_json_tree_lazy = False
+
+    @classmethod
+    def create_tree_from_json_string(cls, json_string: str, load_tree_lazy: bool = False) -> "SpanTree":
+        """Create SpanTree object from "root_span" json string."""
+        obj = cls.__new__(cls)
+        super(SpanTree, obj).__init__()
+        obj._load_json_tree_lazy = load_tree_lazy
+        # Default behavior is to load the whole tree from top level json string.
+        # TODO: Stretch goal will be to load the tree lazily one level at a time to save computation/space.
```
RichardLi1437, line 107: I'm wondering how you want to do this lazy loading, user may not access the tree level by level, in fact, more likely, they may access some specific branches till leaf span and skip other branches in the tree.
If we can do on demand loading, I'm good, but I'm afraid this ends up with each node access require a deserialization of the whole json string, if that's the case, why not just load them all.

alanpo1: When I wrote level I meant node, so I was thinking to have a mechanism to load the json of each node on-demand and skip the rest of the children nodes below it. The children of a node are serialized as a list of json strings so then when we load the tree node they could stay as json strings and only if we want to access the children would we deserialize each node on demand. I've been thinking about it and not sure if this would be worth the extra complexity so I was actually probably going to abandon that mechanism.

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        if self.root_span is None:
+            return
+        for span in self.root_span.__iter__():
+            yield span
+
+    def _from_json_str_repr(self, json_string: str) -> SpanTreeNode:
+        """Convert json string to SpanTree Node fully."""
+        output_node = SpanTreeNode.create_node_from_json_str(json_string)
+        child_subtree_nodes = []
+        for child in output_node.children:
```
RichardLi1437, line 149: I believe we should move this children logic to SpanTreeNode

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _to_json_str_repr(self, curr_span: SpanTreeNode) -> str:
+        """Recursively get tree structure JSON string."""
+        output = curr_span.to_dict()
+        start_time: datetime = output['start_time']  # type: ignore
+        end_time: datetime = output['end_time']  # type: ignore
+        output['start_time'] = start_time.isoformat()
+        output['end_time'] = end_time.isoformat()
+
+        child_subtree_strs = []
```
RichardLi1437, line 163: I believe we should move this children logic to SpanTreeNode

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def to_dict(self) -> dict:
+        """Get dictionary representation of SpanTreeNode."""
+        span_row_dict = self.span_row.asDict()
+        span_row_dict['children'] = self.children
+        return span_row_dict
+
+    def __iter__(self) -> Iterator["SpanTreeNode"]:
+        """Iterate over current span and child spans."""
+        for child_span in self._children:
```
RichardLi1437, line 81: we should use `self._children or []` in case _children is None

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+        obj = cls.__new__(cls)
+        super(SpanTreeNode, obj).__init__()
+        obj._span_row = new_row
+        obj._children = children
+        return obj
+
+    def insert_child(self, span: "SpanTreeNode") -> None:
+        """Insert a child span in ascending time order due to __lt__()."""
+        bisect.insort(self._children, span)
```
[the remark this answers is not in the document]
alanpo1: spark 3.1 uses python 3.8.10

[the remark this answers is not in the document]
alanpo1: Bisect is available yes. The only issue previously mentioned was that bisect didn't have a key-word "key" to set the sort-by ordering in python3.8. That's why I defined a custom __lt__() function

ycheng35xo (approved):

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        """Get the span's children as list."""
+        return self._children
+
+    @children.setter
+    def children(self, value: list) -> None:
+        """Set the span's children."""
+        self._children = value
+
+    @property
+    def span_row(self) -> Row:
```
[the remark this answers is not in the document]
alanpo1: will do in v2

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_run.py
```diff
@@ -3,4 +3,141 @@
...
+        data_window_end: The end date of the data window.
+        input_data: The data asset on which the date filter is applied.
+        preprocessed_span_data: The mltable path pointing to location where the
+        outputted span logs mltable will be written to.
+        aggregated_trace_data: The mltable path pointing to location where the
+        outputted aggregated trace logs mltable will be written to.
+    """
+    store_url = StoreUrl(input_data)
+
+    transformed_df = _genai_uri_folder_to_preprocessed_spark_df(data_window_start, data_window_end, store_url)
```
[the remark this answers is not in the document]
alanpo1: will do in v2

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
alanpo1: will add in v2

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    def show(self) -> None:
+        """Print to stdout a formatted representation of the Span Tree."""
+        if self.root_span is None:
+            return
+        self.root_span.show()
+
+    def to_json_str(self) -> str:
+        """Get tree structure as json string."""
+        if self.root_span is None:
+            return None  # type: ignore
```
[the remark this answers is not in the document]
alanpo1: Probably "null" since the schema is expecting a json string and "null" is what is outputted by json.dumps(None)

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        return self._to_json_str_repr(self.root_span)
```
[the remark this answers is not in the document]
alanpo1: Working on new logic for v2 to first make the nodes into dicts and then json.dumps them.

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
alanpo1: In v2.

assets/model_monitoring/components/src/model_data_collector_preprocessor/span_tree_utils.py
alanpo1: in v2

PR #2359, opened 2024-02-21
Add input schema adaptor to gsq signal graph

alanpo1:
Preprocessor into new GSQ signal:
https://ml.azure.com/runs/2ad28f1e-5ba9-480b-81f5-93a383498043?wsid=/subscriptions/5f08d643-1910-4a38-a7c7-84a39d4f42e0/resourceGroups/model-monitoring-bugbash-2023-05-09-rg/providers/Microsoft.MachineLearningServices/workspaces/momo-bugbash-2023-05-09-ws&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/tests/unit/test_gsq_input_schema_adaptor.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+
+    _expected_gsq_input_schema_extra = StructType(
+        [
+            StructField("trace_id", StringType(), True),
+            StructField("context", StringType(), True),
+            StructField("groundtruth", StringType(), True),
+            StructField("prompt", StringType(), True),
+            StructField("output", StringType(), True),
+            StructField("source", StringType(), True),
```
ycheng35xo, line 41: what does this source mean? Sometimes the output has 2 parameters inside it?

alanpo1: source was just a randomly named extra key that I included in the data to test behavior when there are extra data in the input/output dicts.

ycheng35xo: So basically we want to expand all the fields in "input" and "output" as an individual column in gsq input? And the groundtruth, context, .. those information will be part of the "input"? Is my understanding correct?

alanpo1: Yes that's correct. We cannot know the 'input' and 'output' schema in the adaptor that is up to the user we just expand and pass on to the gsq where they give us the column mapping.

alanpo1:
@RichardLi1437 Do you have more info on if we will be able to get GSQ expected columns logged in the raw mdc logs? For example, the gsq needs 'question', 'answer', 'context', and 'ground-truth' columns to be specified but for a promptflow deployment with MDC enabled will we be sure those columns are logged in the attributes field?

assets/model_monitoring/components/src/shared_utilities/constants.py
```diff
@@ -91,6 +91,10 @@
 MDC_DATA_COLUMN = 'data'
 MDC_DATAREF_COLUMN = 'dataref'
 
+# Column names in GenAI Preprocessor
+GENAI_ROOT_SPAN_SCHEMA_COLUMN = "root_span"
```
imatiach-msft, line 95: nice maybe I can re-use these same constants in GSQ

imatiach-msft (approved):

RichardLi1437:
> @RichardLi1437 Do you have more info on if we will be able to get GSQ expected columns logged in the raw mdc logs? For example, the gsq needs 'question', 'answer', 'context', and 'ground-truth' columns to be specified but for a promptflow deployment with MDC enabled will we be sure those columns are logged in the attributes field?
yes, they should be in input/output of root span. BTW: not necessary question, answer, user should provide column mapping to map their column(s) to question, answer, context, ground truth.

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
```diff
@@ -5,10 +5,55 @@
...
+
+    spark = init_spark()
+    try:
+        input_schema = spark.read.json(df.rdd.map(lambda row: row.input), mode="FAILFAST").schema
+        output_schema = spark.read.json(df.rdd.map(lambda row: row.output), mode="FAILFAST").schema
+    except Exception as ex:
+        if "Malformed records are detected in schema inference. Parse Mode: FAILFAST." in str(ex):
+            raise InvalidInputError(
+                "Failed to parse the input/output column json string for the trace logs provided."
+                " The input/output columns in the inputted production data are not in a parseable" +
```
RichardLi1437, line 36: the trailing + is not necessary.

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
```diff
@@ -5,10 +5,55 @@
...
+    """Adapt the input dataframe schema to fit GSQ input schema."""
+    df_field_names = df.schema.fieldNames()
+
+    # check if we need to adapt the schema
+    if GENAI_ROOT_SPAN_SCHEMA_COLUMN not in df_field_names and GENAI_TRACE_ID_SCHEMA_COLUMN not in df_field_names:
+        return df
+
+    spark = init_spark()
+    try:
+        input_schema = spark.read.json(df.rdd.map(lambda row: row.input), mode="FAILFAST").schema
```
RichardLi1437, line 30: the input is the trace log of our GenAI preprocessor output, right? I'm thinking may be we can only retrieve the input/output schema from the first, say, 100 rows instead of the whole df to save resources, any concern?

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
```diff
@@ -5,10 +5,55 @@
...
+                " The input/output columns in the inputted production data are not in a parseable" +
+                " json string format. Please double-check the data columns are being passed or" +
+                " stored correctly."
+            )
+        raise ex
+
+    df = df.withColumns({
+        'temp_input': from_json(df.input, input_schema),
+        'temp_output': from_json(df.output, output_schema),
+    }).select('trace_id', 'temp_input.*', 'temp_output.*', 'root_span')
```
RichardLi1437, line 45: what happened if there are duplicated columns in temp_input and temp_output? this is an un-supported scenario so far, we may want to raise validation error for now, though it is an unlikely scenario.

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
```diff
@@ -5,10 +5,55 @@
...
+
+    # flatten unpacked json columns to json_string if necessary
+    df = _convert_complex_columns_to_json_string(df)
+
+    df.show(truncate=False)
```
RichardLi1437, line 50: it's ok to keep it for now for debugging, please remember to remove it or set truncate=True before GA.

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
```diff
@@ -5,10 +5,55 @@
...
+    if df.isEmpty():
```
RichardLi1437, line 51: I suppose we have a 1:1 row mapping between the original production_data_df and the adapted df? If so, should we do the df empty check against production_data_df instead of the adapted df?

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
```diff
@@ -22,9 +67,9 @@ def run():
 
     production_data_df = try_read_mltable_in_spark_with_error(args.production_dataset, "production_dataset")
 
-    # TODO: Add input schema adaptor logic. For now just pass data through.
+    transformed_df = _adapt_input_data_schema(production_data_df)
```
RichardLi1437, line 70: rename to adapted_df

assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
```diff
@@ -99,13 +99,28 @@ outputs:
...
+    component: azureml://registries/azureml/components/gsq_input_schema_adaptor/versions/0.0.2
+    inputs:
+      production_dataset:
+        type: mltable
+        path: ${{parent.inputs.production_data}}
+    outputs:
+      adapted_production_data:
+        type: mltable
+    resources:
+      instance_type: ${{parent.inputs.instance_type}}
```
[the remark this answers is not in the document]
alanpo1: As far as I know we don't support changing the `runtime_version` from the user pov so that's why we don't inherit from parent inputs. I don't think any component shares the `runtime_version` they are all static

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
alanpo1: I suppose we can take a sampling of the data, it can take a while to read the json. Should we then catch errors thrown later related to the schema when we try unpacking the string with from_json()?

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
alanpo1: If there are duplicated columns we will fail to save the dataframe a few lines down. I will add validation error

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
alanpo1: We already have those validations when reading from mltable. I actually will remove this isEmpty() check since I added it before including the 'trace_id' and 'root_span' columns in the output df. I was trying to check if the input/output columns have no nested data but that will just be passed down stream and fail in the gsq.

assets/model_monitoring/components/src/generation_safety_quality/input_schema_adaptor/run.py
alanpo1: Do you think we should use a fraction of rows as opposed to a set number? I was thinking sample 20%

assets/model_monitoring/components/src/shared_utilities/df_utils.py
```diff
@@ -240,3 +240,8 @@ def try_get_df_column(df: pyspark_sql.DataFrame, name: str) -> Optional[pyspark_
         return df[name]
     except AnalysisException:
         return None
+
+
+def has_duplicated_columns(df: pyspark_sql.DataFrame) -> bool:
+    """Check if a dataframe has duplicate columns."""
+    return not df.groupBy(df.columns).count().where("count > 1").isEmpty()
```
RichardLi1437, line 247: have you tested this? per my test, this will throw below exception if there are duplicated column names in df:
AnalysisException: [AMBIGUOUS_REFERENCE] Reference `<col_name>` is ambiguous, could be: [`<col_name>`, `<col_name>`].
I think we can simply check if df.columns has duplication. like below:
```py
def has_duplicated_columns(df):
  col_names = df.columns
  col_names_set = set(col_names)
  return len(col_names) > len(col_names_set)
```

RichardLi1437 (approved):

ycheng35xo (approved):

PR #2366, opened 2024-02-22
Added CPU minimal 20.04 python 3.11 and 22.04 python 3.11 

amarahiqbal:
New curated envionments
	- Added minimal-ubuntu22.04-py311-cpu-inference
	- Added minimal-ubuntu20.04-py311-cpu-inference


vizhur (changes requested):
please have MCR and ownership PRs first

PR #2388, opened 2024-02-26
Enabling OssFlavoured tags for the model that supports OSS

HrishikeshGeedMS:


assets/models/system/bert-base-cased/spec.yaml
```diff
@@ -16,6 +16,7 @@ tags:
   license: apache-2.0
   model_specific_defaults : {'apply_deepspeed': 'true', 'apply_lora': 'true', 'apply_ort': 'true'}
   SharedComputeCapacityEnabled: ''
+  mlflow_flavor: transformers
```
novaturient95, line 19: why is this required? For filteration?

Anubha98: Also this will provide a way from SDK to know the flavor, currently there is no way from sdk to know flavor, exception downloading model and check for artifacts

Anubha98: > Also this will provide a way from SDK to know the flavor, currently there is no way from sdk to know flavor, exception downloading model and check for artifacts

Also this will provide a way from SDK to know the flavor, currently there is no way from sdk to know flavor, exception downloading model and check for artifacts

Anubha98 (approved):

novaturient95 (dismissed):

PR #2391, opened 2024-02-26
add evaluation to GSQ model monitoring component output

imatiach-msft:
For new model monitoring UI, GSQ component needs to output a table of question/answer/context input data with corresponding scores.  This PR adds this table as an additional output port in the GSQ model monitoring component.

Update: based on discussion in model monitoring scrum, I've removed the average scores as an output table.  These will just be added as part of the compute metrics component based on the histogram table output.

Screenshot of new output ports:
image

Screenshot of new signal output file with paths to output ports:
image

### DIFF
--- a/assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
+++ b/assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
@@ -134,6 +134,8 @@ jobs:
         type: mltable
       similarity_violations:
         type: mltable
+      evaluation:
+        type: mltable
     resources:
       instance_type: ${{parent.inputs.instance_type}}
       runtime_version: "3.3"
--- a/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
+++ b/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
@@ -337,6 +337,7 @@ def run():
     parser.add_argument("--similarity_violations", type=str, required=True)
 
     parser.add_argument("--workspace_connection_arm_id", type=str, required=True)
+    parser.add_argument("--evaluation", type=str, required=True)
     args = parser.parse_args()
 
     request_args = {
@@ -380,7 +381,8 @@ def run():
         completion_column_name=args.completion_column_name,
         context_column_name=args.context_column_name,
         ground_truth_column_name=args.ground_truth_column_name,
-        violations=violations
+        violations=violations,
+        evaluation=args.evaluation
     )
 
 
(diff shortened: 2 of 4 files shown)


tongyu-microsoft (approved):

assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
```diff
@@ -134,6 +134,8 @@ jobs:
         type: mltable
       similarity_violations:
         type: mltable
+      evaluation:
```
ycheng35xo, line 137: can you make this evaluation as a signal output? i.e. adding another port here https://github.com/Azure/azureml-assets/blob/ed10abd8d33d80a5e447d94872a00cfad6374763/assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml#L97C1-L98C1

I think this does not require extra service change.

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -669,8 +674,7 @@ def mock_metrics_batch(iterator):
                     qca[CONTEXT] = row[CONTEXT]
                 if has_ground_truth:
                     qca[GROUND_TRUTH] = row[GROUND_TRUTH]
-                for metric_name in metric_names:
-                    metric_name_compact = get_compact_metric_name(metric_name)
+                for metric_name_compact in compact_metric_names:
```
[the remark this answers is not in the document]
imatiach-msft: on correlation id/trace id, I can add both actually and even more columns, but I can do all of this in another PR as it's separate from this change

assets/model_monitoring/components/generation_safety_quality/generation_safety_quality_signal_monitor/spec.yaml
imatiach-msft: on making a separate signal output - in that case, I would need to have GSQ pipeline output a second output port.  I think I might need to add another Metric Outputter component in that case:
image

imatiach-msft: so this seems like a separate feature/PR in and of itself, but I think we should talk to Corey about that first

ycheng35xo (approved):

assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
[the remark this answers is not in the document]
alanpo1: The new GenAI schema column names that will be passed to gsq signal are `trace_id` and `root_span` corresponding to the trace id and debug info columns we want to output.

PR #2394, opened 2024-02-26
update code to pass momo job info in spark job

shreeyaharma:


RichardLi1437:
this is to inform Spark job to enable cache? do we have PR in Spark side to handle this new spark conf?

yunjie-hub (approved):

RichardLi1437 (approved):
this is to inform Spark job to enable cache? do we have PR in Spark side to handle this new spark conf?

PR #2403, opened 2024-02-27
add traceid and rootspan passthrough columns to GSQ histogram model monitoring component

imatiach-msft:
Add a whitelist of passthrough columns - specifically traceid and rootspan, as well as the previous correlation_id, to the GSQ histogram component.  These columns will be allowed to pass through to the evaluation dataset output by the component if provided as an input.

### DIFF
--- a/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
+++ b/assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
@@ -37,6 +37,10 @@ COMPLETION = "completion"
 CONTEXT = "context"
 GROUND_TRUTH = "ground_truth"
 CORRELATION_ID = "correlationid"
+TRACE_ID = "traceid"
+ROOT_SPAN = "rootspan"
+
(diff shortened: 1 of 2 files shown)


assets/model_monitoring/components/src/generation_safety_quality/annotation_compute_histogram/run.py
```diff
@@ -37,6 +37,10 @@
 CONTEXT = "context"
 GROUND_TRUTH = "ground_truth"
 CORRELATION_ID = "correlationid"
+TRACE_ID = "traceid"
+ROOT_SPAN = "rootspan"
+
+PASSTHROUGH_COLUMNS = [CORRELATION_ID, TRACE_ID, ROOT_SPAN]
```
ycheng35xo, line 43: nit: this correlation_id is an old schema so should be deprecated

imatiach-msft: I know I'm just supporting everything possible for backwards compat - I don't think this will break anything.  We can always remove it sometime later in like a year when we know it hasn't been used in forever.

ycheng35xo (approved):

PR #2413, opened 2024-02-28
add --all option for SpecVersionUpgrader

RichardLi1437:


assets/model_monitoring/components/tools/spec_version_upgrader.py
```diff
@@ -165,9 +169,16 @@ def _scan_spec_folder(spec_folder: str) -> dict:
 if __name__ == "__main__":
     # TODO add --all to upgrade all components
     argparser = ArgumentParser()
-    argparser.add_argument("--specs", help="spec names to upgrade")
+    argparser.add_argument("-s", "--specs", help="comma separated spec names to upgrade, "
+                           "e.g. model_data_collector_preprocessor,gsq_annotation_compute_histogram")
```
ycheng35xo, line 173: question: for the usage, we just need to add the component level name in the argument, no need to add signal(sub-graph) level, right?

RichardLi1437: yes, the upgrade tool can figure out the dependencies and upgrade the subgraphs automatically if necessary.

PR #2419, opened 2024-02-29
AttributeError when either input data or base line data empty

vaibhj:


assets/model_monitoring/components/src/model_monitor_compute_histogram_buckets/run.py
```diff
@@ -40,6 +40,8 @@ def run():
         return
     elif df1 is None:
         df1 = df2
+    elif df2 is None:
+        df2 = df1
```
ycheng35xo, line 44: I remember in the feature selector you have a same logic, can you make a util function and change for both places?

ycheng35xo: Add you can add UT for it

ycheng35xo (approved):

assets/model_monitoring/components/src/model_monitor_compute_histogram_buckets/run.py
vaibhj: There is little difference between two. In one we do falsy check and other one we do none check. 

Use “if not df1:” when you want to check if df1 is falsy .
Use “if df1 is None:” when you specifically want to check if df1 is None.

PR #2438, opened 2024-03-04
add job cleanup for e2e tests

alanpo1:


assets/model_monitoring/components/tests/e2e/conftest.py
```diff
@@ -609,3 +610,45 @@ def uploaded_version_file_name():
...
+            print(f"job: {job.name}, status: {job.status}, end_status: {job_at_end.status}")
+            if job_at_end.status not in RunHistoryConstants.TERMINAL_STATUSES:
+                ml_client.jobs.begin_cancel(job.name)
+
+    request.addfinalizer(cancel_jobs)
+    yield _submit_job
+
+
+@pytest.fixture(scope="session")
+def cleanup_previous_e2e_tests(ml_client: MLClient, test_suite_name):
```
ycheng35xo, line 640: what's this test_suite_name? experiment name?

assets/model_monitoring/components/tests/e2e/utils/test_publish_components_for_ci.py
```diff
@@ -13,6 +13,6 @@
 class TestPublishComponentsForCI():
     """Class for model-monitoring-ci workflow to upload components and data once before splitting tests to runners."""
 
-    def test_publish_components(self, upload_component_version_file):
+    def test_publish_components(self, upload_component_version_file, cleanup_previous_e2e_tests):
```
ycheng35xo, line 16: Does that mean every time one e2e test is running, the previous unfinished jobs will be cancelled? What if a new test is started before the previous test finish, would they have different experiment names?

assets/model_monitoring/components/tests/e2e/conftest.py
```diff
@@ -609,3 +610,45 @@ def uploaded_version_file_name():
...
+    """Clean up any previously running e2e tests in test suite name."""
+    print(f"Cleaning up past jobs that weren't cancelled in experiment_name={test_suite_name}.")
+    if test_suite_name == "local":
```
ycheng35xo, line 643: we might need to notify team that do not override the experiment name for e2e test, or need to find another way to check the local job vs ci job. For example, submitter identity or maybe add an option parameter is_local which can pass to this function

assets/model_monitoring/components/tests/e2e/conftest.py
alanpo1: test_suite_name is based on the "github_ref", test_suite_name we use as the experiment name. So when the tests run on a PR the github ref is for each PR so something like "merge_2346"

assets/model_monitoring/components/tests/e2e/conftest.py
alanpo1: Yes that's true. I didn't see any of the e2e tests overriding the experiment name but it would be a good idea to mention that for developers. I am not sure how much more work it would be to idenitfy the submitter identity vs just using the test_suite_name fixture

alanpo1: In a way test_suite_name fixture is already the optional parameter since it looks in the environment variables for "GITHUB_REF_NAME" and majority of people don't set that so that is how we know it is local

assets/model_monitoring/components/tests/e2e/utils/test_publish_components_for_ci.py
alanpo1: First see the comment above about the experiment names. Then, no this class module specifically is only ran once during CI setup and it publishes the components and now cleans up the tests. The issue rn is that every time you push a new commit to a PR or run the tests again if they failed it will trigger more e2e tests but the previous ones are still running. This can quickly overload the workspace and the tests will get stuck with not enough quota to run everything. We need to cleanup the previous tests

ycheng35xo: I understand. My question is more like, let's see you have a PR and your PR is running a CI. Then I submitted another PR, which will also trigger a CI, and the cleanup will run. Will this cancel the jobs in your PR? I assume no, right? Only same PR will have same test_suite?

ycheng35xo (approved):

PR #2460, opened 2024-03-07
Update vllm version to 3.3

novaturient95:


novaturient95:
Not needed. FM:30 works for phi model

PR #2464, opened 2024-03-08
Upgrade base image python from 3.8 to 3.10

nvijayrania:
Upgrade base image python from 3.8 to 3.10 for vision environments


nvijayrania:
Abandoned PR

PR #2467, opened 2024-03-08
fix action analyzer bug

ycheng35xo:
bugfix:
1. gsq bug: trace_id and root_span
2. action analyzer pip list change
3. action analyzer parse the signal output folder path incorrectly
4. action analyzer uses attributes to parse value


RichardLi1437 (approved):

assets/model_monitoring/components/src/action_analyzer/action_analyzer_identify_problem_traffic/run.py
```diff
@@ -174,23 +174,19 @@ def parse_debugging_info(root_span):
...
-                if isinstance(queries, str):
-                    text = []
-                    score = []
-                    for lookup_output in lookup_outputs:
-                        text.append(lookup_output["text"])
-                        score.append(float(lookup_output["score"]))
-                    spans_array.append((parent_id, index_content, index_id, queries, TEXT_SPLITTER.join(text), max(score)))  # noqa: E501
+                retrieval_info = json.loads(span.attributes)
+                query = retrieval_info["retrieval.query"]
+                retrieval_documents = json.loads(retrieval_info["retrieval.documents"])
```
RichardLi1437, line 183: this is to retrieve info from attriubutes[retreval.do.document] instead of attributes[inputs]?

RichardLi1437 (approved):

assets/model_monitoring/components/src/action_analyzer/action_analyzer_identify_problem_traffic/run.py
ycheng35xo: yes because for the new runtime, the Retrieval span does not inputs in the attributes. Instead, it has retrieval.documents, retrieval.query

imatiach-msft (approved):

assets/model_monitoring/components/src/action_analyzer/action_analyzer_identify_problem_traffic/run.py
```diff
@@ -174,23 +174,19 @@ def parse_debugging_info(root_span):
                 if not parent_id:
                     print("No look up span found, skip action analyzer.")
                     return None
-                # todo, change this function using utils
-                lookup_span = tree.get_span_tree_node_by_span_id(parent_id)
-                lookup_input = json.loads(lookup_span.input)
-                index_content = lookup_input["mlindex_content"]
+                index_span = tree.get_span_tree_node_by_span_id(parent_id)
+                index_input = json.loads(json.loads(index_span.attributes)["inputs"])
```
imatiach-msft, line 178: this is really weird - calling json.loads twice... but maybe that is correct, just seems strange

ycheng35xo: The raw log is a dict of a dict and serialized. After the first parse, the dict.value is still a jsonstring

PR #2487, opened 2024-03-11
Don't promote inputs/output from attributes in GenAI preprocessor

alanpo1:
Changes in this PR:
Genai Preprocessor:
- no longer promote input/output as first-party columns for span logs. Leave inside of attributes json string.
  - trace logs will still promote input/output

Span_tree_utils:
- Change behavior of input/output properties to load the attributes json and return the available input/outputs if it exists.
- Add documentation note to SpanTreeNode to not use it to deserialize nodes individually.

Sample Run: https://ml.azure.com/runs/amusing_kettle_c17vch2hcc?wsid=/subscriptions/5fca341e-4ec3-45bd-b006-a7648d9376c5/resourcegroups/modelmonitoring-e2e-rg/workspaces/momo-e2e-test-ws-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/src/shared_utilities/span_tree_utils.py
```diff
@@ -101,7 +111,12 @@ def trace_id(self) -> str:
 
     @classmethod
     def create_node_from_dict(cls, span_node_dict: dict) -> "SpanTreeNode":
```
RichardLi1437, line 113: can we rename it to _create_node_from_dict() if this is not intended to be exposed?

RichardLi1437 (approved):

assets/model_monitoring/components/src/shared_utilities/span_tree_utils.py
alanpo1: It's fine to be exposed and could be used in some random cases to build nodes if we so need. The note is because I noticed in token stats v1 PR that Shreeya was using it to manually deserialize each node instead of using the `SpanTree.create_tree_from_json_string()` method which does it all for you.

PR #2488, opened 2024-03-11
use new prompt to calculate metrics

ycheng35xo:
Changes in this PR:
1. update the index score LLM prompt, and use it as the final score
2. add aml deployment id in the action response
3. add a min sample size for grouping the queries
4. add index assetId in the action response


assets/model_monitoring/components/src/action_analyzer/action_analyzer_correlation_test/run.py
```diff
@@ -108,7 +110,7 @@ def perform_ttest(good_answer_scores, bad_answer_scores):
     # Perform Fisher's Exact Test
     t_stat3, p_value3 = stats.fisher_exact(table)
     print(f"Fisher's Exact Tes odds_ratio: {t_stat3}, P-value: {p_value3}")
-    return t_stat1, p_value1
+    return t_stat2, p_value2
```
yunjie-hub, line 114: add comments that we use mannwhitneyu t-test

assets/model_monitoring/components/src/action_analyzer/action_analyzer_correlation_test/run.py
```diff
@@ -144,6 +147,7 @@ def run():
     print(group_set)
     print("===All indexes===")
     print(index_set)
+    data_with_action_metric_score_df = data_with_action_metric_score_df.filter(col(INDEX_SCORE_LLM_COLUMN) > DEFAULT_RETRIEVAL_SCORE)  # noqa: E501
```
yunjie-hub, line 150: This is to filter out the failed one? e.g. we get -1 if the LLM call failed?

assets/model_monitoring/components/src/action_analyzer/action_analyzer_metrics_calculation/run.py
```diff
@@ -50,6 +52,17 @@
 )
 
 
+def _post_process_results(output):
+    parsed_score_response = re.findall(r'\d+', output.split("# Result")[-1].strip())
```
yunjie-hub, line 56: we cannot control the output format of the score response? is this stable?

assets/model_monitoring/components/src/action_analyzer/action_analyzer_metrics_calculation/run.py
```diff
@@ -50,6 +52,17 @@
...
+    if len(parsed_score_response) > 0:
+        score = float(parsed_score_response[0].replace("'", "").strip())
```
yunjie-hub, line 58: Should we add try here

ycheng35xo: I copied this exact parse function from the sdk. Basically, when successfully parsed (regex match), we set it as 1-5. If the result does not match the expected regex, we set it as 0. Then later when we do the T-test, we filter out 0s. 

In my test, 80 samples I saw only one 0 (failure). 

https://msdata.visualstudio.com/Vienna/_git/AzureMlCli?path=/src/azureml-metrics/azureml/metrics/text/rag_evaluation/_rag_evaluation.py&version=GBmaster&line=535&lineEnd=543&lineStartColumn=1&lineEndColumn=29&lineStyle=plain&_a=contents

assets/model_monitoring/components/src/action_analyzer/action_analyzer_correlation_test/run.py
ycheng35xo: set as 0 if failed.

yunjie-hub (approved):

PR #2517, opened 2024-03-15
add code to include new contract values

shreeyaharma:
https://ml.azure.com/experiments/id/4b4e34ef-331f-4b1a-b1b0-88615d6c45ad/runs/ivory_dream_0h6xqptzff?wsid=/subscriptions/1aefdc5e-3a7c-4d71-a9f9-f5d3b03be19a/resourcegroups/TestingTeamProjectRG/providers/Microsoft.MachineLearningServices/workspaces/shrsha-eastus-test&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/src/genai_token_statistics_compute_metrics/constants.py
```diff
@@ -16,7 +16,9 @@
 MODEL_COMPLETION_COUNT = "model_completion_count"
 
 # Constants for attribute keys
-COMPLETION_COUNT_KEY = "llm.token_count.completion"
-PROMPT_COUNT_KEY = "llm.token_count.prompt"
-TOTAL_COUNT_KEY = "     llm.token_count.total"
-MODEL_KEY = "llm.model"
+COMPLETION_COUNT_KEY = ["llm.token_count.completion", "llm.usage.completion_tokens"]
```
alanpo1, line 19: Just for my understanding, are we expecting only one of these keys to appear for each span? If not, what would happen if multiple of these exist in span attributes does it matter which one we count?

alanpo1 (approved):
Just for my knowledge,

assets/model_monitoring/components/src/genai_token_statistics_compute_metrics/constants.py
shreeyaharma: yes only one of these is expected to handle multiple runtime. also even if multiple of them exists which is not possible, we take value from only one making sure we dont compute the value wrong.

assets/model_monitoring/components/src/genai_token_statistics_compute_metrics/constants.py
```diff
@@ -16,7 +16,9 @@
...
+PROMPT_COUNT_KEY = ["llm.token_count.prompt", "llm.usage.prompt_tokens", "embedding.token_count.prompt",
```
RichardLi1437, line 20: I believe we should put the latest schema, I suppose it is "embedding.usage.prompt_tokens" as the first key, to reduce the fallback time.
Please adjust the key sequence for all XXX_KEYS constant list.

assets/model_monitoring/components/src/genai_token_statistics_compute_metrics/metrics_processor.py
```diff
@@ -167,10 +167,16 @@ def calculate_metrics(self, attributes: dict, span_type: str, root_input: str, r
                 "output": root_output
             })
 
+    def get_value_from_attributes(self, attributes, keys):
```
RichardLi1437, line 170: add a default_value instead of a hard code 0 in case no attribute found, can default the default_value to 0 though.

assets/model_monitoring/components/src/genai_token_statistics_compute_metrics/metrics_processor.py
```diff
@@ -167,10 +167,16 @@ def calculate_metrics(self, attributes: dict, span_type: str, root_input: str, r
...
+        """Get usage values from attributes."""
+        for key in keys:
+            if key in attributes:
+                return attributes.get(key)
+        return 0
+
     def has_completion_count(self, span_type):
         """Check if model is completion type."""
         if span_type == "LLM":
```
RichardLi1437, line 179: nit: 
```py
return span_type == "LLM"
```

assets/model_monitoring/components/src/genai_token_statistics_compute_metrics/metrics_processor.py
RichardLi1437: seems this is not addressed?

RichardLi1437 (approved):

PR #2533, opened 2024-03-19
Adding prefix functionality to Chat Prompts in Prompt Crafter Component

SamGos93:
For chat prompts, if prefix was given, then it was added as a separate context with user role. However, this does not work for several models as they expect user/assistant roles in sequence. Also, to provide flexibility to the chat prompts, the prefix is updated as below:

- If prefix is present in chat prompts, it will be added in the beginning of the first few shot message (if n shot data is available) or the beginning of the prompt message (in the context).

- In case the prefix is in the form of role/ context list of dictionary format, the logic remains as before: prefix + few shot+ prompt. This is majorly to ensure that the test cases for Babel work well.

The logic for completion prompts remains the same.


assets/aml-benchmark/components/src/aml_benchmark/prompt_crafter/package/prompt_factory.py
```diff
@@ -303,15 +303,32 @@ def create_prompt(self, row: Dict) -> ChatPrompt:
...
         if self.prefix:
             prefix_messages = self._parse_affix(affix=self.prefix)
-            messages.extend(prefix_messages)
+
+        # Parsing the prompt pattern
+        input_messages = self._parse_affix(affix=self.prompt_pattern)
 
         if self.few_shot_pool is not None and self.n_shots > 0:
             few_shot_messages = self._create_few_shots(row)
+            if len(prefix_messages) == 1 and len(few_shot_messages) > 1:
```
arun-rajora, line 315: Why are we checking if len few_shot_messages > 1? For n_shots = 1, will we handle it differently?

SamGos93: No. Updated it. Thanks!

iamrk04 (approved):

arun-rajora (approved):

PR #2536, opened 2024-03-19
action analyzer add new changes

ycheng35xo:
changes included:
1. modify the sampling logic for good queries
2. add show_topic_name config for disabling the topic name
3. extract query_type and top_k from the debugging info
4. add the query intention by selecting the most confident group for each action
5. change signal to violated metrics in output
6. change min_topic_size to 15% of query length in Bertopic (tested 10%, 25%, 20%. Seems 15 gives better result)

Disable the topic name
image

Default topic name
image


yunjie-hub (approved):

assets/model_monitoring/components/model_monitor_action_analyzer/action_analyzer_identify_problem_traffic/spec.yaml
```diff
@@ -27,16 +27,10 @@ inputs:
     type: string
     optional: false
     description: model name, e.g., name of the deployment for Azure OpenAI
-  prompt_column_name:
+  show_topic_name:
```
yunjie-hub, line 30: why don't we use bool type?

yunjie-hub, line 30: named as "llm_summary_enabled"?

ycheng35xo: bool type has some limitations in the yaml arguments. Fixed by a workaround.

yunjie-hub (approved):

PR #2581, opened 2024-03-25
Bugfix for Output metrics component

alanpo1:


assets/model_monitoring/components/src/model_monitor_output_metrics/entities/signals/data_quality_signal.py
```diff
@@ -111,8 +111,10 @@ def _build_feature_metrics(
             )
             run_metric = {
                 "runId": run_id,
-                "targetValue": float(metric["baseline_metric_value"]),
-                "baselineValue": float(metric["target_metric_value"]),
+                "targetValue": None if metric["target_metric_value"] is None
```
ycheng35xo, line 114: only dataquality has this issue? Would other metrics also has none when parsing to float?

ycheng35xo (approved):

assets/model_monitoring/components/src/model_monitor_output_metrics/entities/signals/data_quality_signal.py
alanpo1: I checked other output metric and we should be fine as they have None checks. If we find this issue in other components we can investigate as they come up as we don't expect None values

PR #2585, opened 2024-03-26
Revert enforcing registry asset URI for components

lykelly19:


scripts/azureml-assets/azureml/assets/publish_utils.py
```diff
@@ -154,6 +154,9 @@ def validate_and_prepare_pipeline_component(
             )
             return False
 
+        # If workspace asset URI is used, use registry we're creating the component in
+        registry = registry or registry_name
```
elesel, line 158: Was this the previous behavior?

elesel (approved):

scripts/azureml-assets/azureml/assets/publish_utils.py
lykelly19: It was originally `registry_name = registry or registry_name` - I was thinking using registry makes more sense than overriding registry_name (This was previous PR: https://github.com/Azure/azureml-assets/pull/2572/files)

elesel (approved):

PR #2592, opened 2024-03-27
Make embeddings_model optional

shpengmsft:
Those components will use Cohere MaaS endpoint to embed data. embeddings_model is not a must to have anymore.


dans-msft (approved):
Looks good!

PR #2609, opened 2024-03-29
Added design modification to read files without jsonl extension and modified custom scripts too as per modification requested.

Nivi09:
Added design modification to read files without jsonl extension and modified custom scripts too as per modification requested.

### DIFF
--- a/assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
+++ b/assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
@@ -113,22 +113,25 @@ def resolve_io_path(dataset: str) -> List[str]:
     return [dataset]
 
 
+def filter_files_with_given_extension(file_paths: List[str], extension: str) -> List[str]:
+    """Filter and returns the list of files with given extension."""
+    return [file_path for file_path in file_paths if file_path.endswith(extension)]
+
+
 def read_jsonl_files(file_paths: List[str]) -> List[Dict[str, Any]]:
     """
-    Read `.jsonl` files if extendion is jsonl else calls `read_files` function for files \
-    that do not endwith jsonl extension but the file content is of type json.
+    Read `.jsonl` file/files and return a list of dictionaries.
 
-    Return a list of dictionaries.
+    If a single path is present in `file_paths` and the file does not have any extension, try \
+    to read it as a `.jsonl` file. This is done to support `uri_file` without extension scenario.
 
     :param file_paths: List of paths to .jsonl files.
     :return: List of dictionaries.
     """
-    if len(file_paths) == 1 and not file_paths[0].endswith(".jsonl"):
-        return read_files(file_paths)
+    if len(file_paths) > 1:
+        file_paths = filter_files_with_given_extension(file_paths, ".jsonl")
     data_dicts = []
     for file_path in file_paths:
-        if not file_path.endswith(".jsonl"):
-            continue
         with open(file_path, 'r') as file:
             for i, line in enumerate(file):
                 try:
@@ -208,24 +211,3 @@ def save_list_to_jsonl_if_path_provided(data: List[Any], path: Optional[str]):
     if path:
         df = pd.DataFrame(data)
         df.to_json(path, orient='records', lines=True)
-
-
-def read_files(file_paths: List[str]) -> List[Dict[str, Any]]:
-    """
-    Read uri_files and return a list of dictionaries.
-
-    :param file_paths: List of paths to uri files.
-    :return: List of dictionaries.
-    """
-    data_dicts = []
-    for file_path in file_paths:
-        with open(file_path, 'r') as file:
-            for i, line in enumerate(file):
-                try:
-                    data_dicts.append(json.loads(line))
-                except json.JSONDecodeError:
-                    mssg = f"Invalid JSON format in line {i + 1} of file '{file_path}'."
-                    raise DataFormatException._with_error(
-                        AzureMLError.create(DataFormatError, error_details=mssg)
-                    )
-    return data_dicts
--- a/assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
+++ b/assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
@@ -115,14 +115,15 @@ def resolve_io_path(dataset: str) -> List[str]:
 
 def read_jsonl_files(file_paths: List[str]) -> List[Dict[str, Any]]:
     """
-    Read `.jsonl` files and return a list of dictionaries.
-
-    Ignores files that do not have `.jsonl` extension. Raises exception if no `.jsonl` files
-    found or if any `.jsonl` file contains invalid JSON.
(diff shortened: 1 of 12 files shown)


assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
```diff
@@ -115,14 +115,16 @@ def resolve_io_path(dataset: str) -> List[str]:
 
 def read_jsonl_files(file_paths: List[str]) -> List[Dict[str, Any]]:
     """
-    Read `.jsonl` files and return a list of dictionaries.
+    Read `.jsonl` files if extendion is jsonl else calls `read_files` function for files \
```
arun-rajora, line 118: nit: Typo in "extendion".

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
```diff
@@ -206,3 +208,24 @@ def save_list_to_jsonl_if_path_provided(data: List[Any], path: Optional[str]):
     if path:
         df = pd.DataFrame(data)
         df.to_json(path, orient='records', lines=True)
+
+
+def read_files(file_paths: List[str]) -> List[Dict[str, Any]]:
```
arun-rajora, line 213: nit: Can we rename it to read_jsonl_files since it will only work for jsonl files else throw an error?

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
```diff
@@ -115,14 +115,16 @@ def resolve_io_path(dataset: str) -> List[str]:
...
-    found or if any `.jsonl` file contains invalid JSON.
+    Return a list of dictionaries.
 
     :param file_paths: List of paths to .jsonl files.
     :return: List of dictionaries.
     """
+    if len(file_paths) == 1 and not file_paths[0].endswith(".jsonl"):
+        return read_files(file_paths)
     data_dicts = []
     for file_path in file_paths:
```
arun-rajora, line 129: nit: Can we use read_files function here? Example-
```
data_dicts = read_files([file_path for file_path in file_paths if file_path.endswith(".jsonl")])
```

assets/aml-benchmark/tests/dataset_preprocessor/test_dataset_preprocessor.py
```diff
@@ -347,9 +347,6 @@ def test_invalid_inputs(
         invalid_dataset_error_mssg = (
             "the following arguments are required: --dataset"
         )
-        invalid_jsonl_dataset_mssg = (
```
arun-rajora, line 350: We would still need tests when no "valid" jsonl file is passed in?

arun-rajora (changes requested):

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
iamrk04: A better description would be:

```
Read `.jsonl` files and return a list of dictionaries.

If a single path is present in `file_paths` and the file does not have any extension, try to read it as a `.jsonl` file. This is done to support `uri_file` without extension scenario.
```

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
iamrk04: Here is a better way to do things without creating confusion.

This function should be renamed to `read_jsonl_file` which will take a single `file_path` and return `dict`. Call this from within `read_jsonl_files`.

assets/aml-benchmark/scripts/custom_dataset_preprocessors/base_preprocessor_template.py
```diff
@@ -20,15 +20,12 @@ def _parse_args():
     return argss
 
 
-def _read_jsonl_file(file_path: str) -> List[Dict[str, Any]]:
-    """Read `.jsonl` file and return a list of dictionaries.
+def _read_input_file(file_path: str) -> List[Dict[str, Any]]:
```
iamrk04, line 23: May I know why does this function even exist? We should try to use the one that we have in `io.py`.

assets/aml-benchmark/scripts/custom_dataset_preprocessors/base_preprocessor_template.py
```diff
@@ -20,15 +20,12 @@ def _parse_args():
...
+    """Read files that have content in the json format and return a list of dictionaries.
```
iamrk04, line 24: This is confusing, we should keep the older function name. we can update the description to:

```
Read `.jsonl` file and return a list of dictionaries.

If the file does not have any extension, try to read it as a `.jsonl` file.
```

iamrk04, line 24: This is confusing, we should keep the older function name. We can update the description to:

```
Read `.jsonl` file and return a list of dictionaries.

If the file does not have any extension, try to read it as a `.jsonl` file.
```

iamrk04 (changes requested):

assets/aml-benchmark/tests/dataset_preprocessor/test_dataset_preprocessor.py
Nivi09: @arun-rajora Why would we need it? This test was for the input_validation method in dataset_preprocessor and inference postprocessor component.

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
Nivi09: @arun-rajora Please elaborate more.

arun-rajora: read_files function is doing similar to what we have starting from line 129. We can simplify all of this using the approach mentioned above. Thanks!

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
Nivi09: Thank you. Updated it.

assets/aml-benchmark/scripts/custom_dataset_preprocessors/base_preprocessor_template.py
Nivi09: Reverted back to the previous version and removed the validation of jsonl file.

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
Nivi09: Made the changes as discussed on the call and removed the read_files function.

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
Nivi09: Removed this function.

arun-rajora (approved):

PR #2632, opened 2024-04-02
added cancel handler

chiragbhatt311:


assets/training/aoai/proxy_components/src/common/cancel_handler.py
```diff
@@ -4,19 +4,30 @@
...
-        self.cancel_triggered = True
+        if self.component is not None:
+            logger.info("calling cancel job for the component")
+            self.component.cancel_job()
+        else:
+            logger.warning("component is none for the cancel handler")
+
+    def register_cancel_handler(component: AzureOpenAIProxyComponent):
+        """Register component cancel job method."""
+        add_custom_dimenions_to_app_insights_handler(logger, component)
```
amangupta26, line 31: why we do this here?

chiragbhatt311: this is to add fields in custom dimension, we can do it in constructor as well.

amangupta26 (approved):

PR #2647, opened 2024-04-04
Add logic for missing span as root_span

alanpo1:


assets/model_monitoring/components/src/shared_utilities/span_tree_utils.py
```diff
@@ -246,6 +248,21 @@ def _construct_span_tree(self, spans: List[SpanTreeNode]) -> Optional[SpanTreeNo
                 parent_span = self.get_span_tree_node_by_span_id(parent_id)
                 if parent_span is not None:
                     parent_span.insert_child(span)
+                else:
+                    # consider the possibility that current span is root_span since
+                    # parent_span is not in the data.
+                    possible_root_spans.append(span)
```
ycheng35xo, line 254: so we have the condition that parent_id exists but there is no parent node?

ycheng35xo (approved):

assets/model_monitoring/components/src/shared_utilities/span_tree_utils.py
alanpo1: Yes, there are several times when this might come up per design or by user setup. Mainly I know that Promptflow has situations where the parent_id is there but the span is not

PR #2667, opened 2024-04-09
update the azure-identity

shpengmsft:
pick the latest azure-identity version.

### DIFF
--- a/assets/large_language_models/rag/environments/rag_embeddings/context/conda_dependencies.yaml
+++ b/assets/large_language_models/rag/environments/rag_embeddings/context/conda_dependencies.yaml
@@ -36,7 +36,7 @@ dependencies:
   - markdown
   - mlflow>=2.6.0
   - mlflow-skinny==2.3.2
-  - msal~=1.22.0
+  - msal~=1.24.0
   - msrest>=0.6.18
   - nltk==3.8.1
   - openai>=0.27.4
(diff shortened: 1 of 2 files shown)


hagh (dismissed):

assets/large_language_models/rag/environments/rag_embeddings/context/conda_dependencies.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - azure-ai-ml=={{latest-pypi-version}}
     # Azure
   - azure-core<2.0.0,>=1.8.0,!=1.22.0
-  - azure-identity==1.12.0
+  - azure-identity=={{latest-pypi-version}}
```
Adarsh-Ramanathan, line 16: Is this value templated in? If so, how do we deal with the case where azure-ai-ml and azure-identity have different versions?

Adarsh-Ramanathan (approved):

assets/large_language_models/rag/environments/rag_embeddings/context/conda_dependencies.yaml
shpengmsft: Yes. These version is filled by build process with the latest version on PyPi at build time.

If there is version dependency issue, the build will break, then we can fix the build break by pinning the versions.

liangjason87 (approved):

PR #2669, opened 2024-04-10
Fetch OIDC token in the background process to avoid expiration issue

iamrk04:
The OID token expires after 5 mins and hence all of our test fails with the following error:
```
ERROR: AADSTS700024: Client assertion is not within its valid time range. Current time: 2024-04-10T19:14:34.7127519Z, assertion valid from 2024-04-10T18:54:41.0000000Z, expiry time of assertion 2024-04-10T18:59:41.0000000Z.
```

I propose to fix this by fetching this token every 4 mins. For more details on the issue, please check [this issue](https://github.com/Azure/azure-cli/issues/28708).

### DIFF
--- a/.github/workflows/assets-test.yaml
+++ b/.github/workflows/assets-test.yaml
@@ -140,7 +140,9 @@ jobs:
           scripts-setup-dir: ${{ env.scripts_setup_dir }}
       
       - name: Test asset
-        run: python -u $scripts_test_dir/test_assets.py -i "${{ matrix.asset_config_path }}" -a $asset_config_filename -p $scripts_test_dir/requirements.txt -r $GITHUB_WORKSPACE/$pytest_reports
+        run: |
+          az account get-access-token --scope https://management.azure.com/.default
+          python -u $scripts_test_dir/test_assets.py -i "${{ matrix.asset_config_path }}" -a $asset_config_filename -p $scripts_test_dir/requirements.txt -r $GITHUB_WORKSPACE/$pytest_reports
 
       - name: Upload test results
         uses: actions/upload-artifact@v3
--- a/.github/workflows/assets-test.yaml
+++ b/.github/workflows/assets-test.yaml
@@ -141,7 +141,7 @@ jobs:
       
       - name: Test asset
         run: |
-          az account get-access-token --scope https://management.azure.com/.default
+          az account get-access-token --scope https://management.azure.com/.default --output none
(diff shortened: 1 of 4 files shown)


.github/workflows/assets-test.yaml
```diff
@@ -138,7 +138,19 @@ jobs:
...
           subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           scripts-setup-dir: ${{ env.scripts_setup_dir }}
-      
+
+      - name: Fetch OID token every 4 mins
+        run: |
+          while true; do
+            token_request=$ACTIONS_ID_TOKEN_REQUEST_TOKEN
+            token_uri=$ACTIONS_ID_TOKEN_REQUEST_URL
+            token=$(curl -H "Authorization: bearer $token_request" "${token_uri}&audience=api://AzureADTokenExchange" | jq .value -r)
```
elesel, line 147: It would help to log the expiration of the federated token here, and ideally use the value to control the sleep timer. Looks like [the federated token is a JWT](https://github.com/Azure/login/blob/e9468bad0a1e2a5863b25f77ea9b6ce0d19f5e41/src/common/LoginConfig.ts#L85).

.github/workflows/assets-test.yaml
```diff
@@ -138,7 +138,19 @@ jobs:
           tenant-id: ${{ secrets.AZURE_TENANT_ID }}
           subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           scripts-setup-dir: ${{ env.scripts_setup_dir }}
-      
+
+      - name: Fetch OID token every 4 mins
```
elesel, line 142: If you want to keep "OID" terminology, please refer to it as "OIDC" instead, or consider this suggestion.
```suggestion
      - name: Refresh login every 4 minutes
```

.github/workflows/assets-test.yaml
iamrk04: Unfortunately, I couldn't see the expiry key in the output from the curl command. It just had "value" and "count" keys in the json.

elesel: You'd have to parse the JWT in the value to see it.

iamrk04: Thanks @elesel

image

elesel (approved):

PR #2672, opened 2024-04-10
add validation logic to not have column with same name

shreeyaharma:


assets/model_monitoring/components/src/model_monitor_data_joiner/run.py
```diff
@@ -83,6 +83,12 @@ def run():
         args.right_join_column
     )
 
+    duplicate_column_names = joined_data_df.columns[joined_data_df.columns.duplicated()]
```
ycheng35xo, line 86: this is just the names not the full column (including the content), right?

ycheng35xo (dismissed):

assets/model_monitoring/components/src/model_monitor_data_joiner/run.py
shreeyaharma: correct

RichardLi1437, line 86: duplicated() is a build in pyspark function for column list? What does it return? can you give me the API manual for this function? Have you tested this method?

assets/model_monitoring/components/src/model_monitor_data_joiner/run.py
```diff
@@ -83,6 +83,11 @@ def run():
         args.right_join_column
     )
 
+    # Check if there are any duplicate column names
+    columns = joined_data_df.columns
+    if len(columns) != len(set(columns)):
+        raise Exception(f"Duplicate column names found: {columns}")
```
ycheng35xo, line 89: do we want to print all the columns but not only the duplicated columns?

assets/model_monitoring/components/tests/e2e/test_data_joiner_e2e.py
```diff
@@ -105,3 +106,20 @@ def test_data_joiner_empty_result_failed(
         )
 
         assert pipeline_job.status == 'Failed'
+
+    def test_data_joiner_common_columns_failed(
```
ycheng35xo, line 110: a minor thought:  would it be easier to test by UT? because adding more and more e2e tests may make the gated tests run very long.

assets/model_monitoring/components/src/model_monitor_data_joiner/run.py
shreeyaharma: good point will change statement

PR #2680, opened 2024-04-12
fetching metrics realtime

chiragbhatt311:


assets/training/aoai/proxy_components/src/finetune_submit.py
```diff
@@ -83,8 +99,6 @@ def submit_finetune_job(self, training_file_id, validation_file_id, model, regis
         """metrics can be retrived only for successful finetune job"""
         finetuned_model_id = finetune_job.fine_tuned_model
         logger.info(f'Fine-tune job: {self.job_id} finished successfully. model id: {finetuned_model_id}')
-        logger.info("fetching training metrics from Azure OpenAI")
```
amangupta26, line 86: I think we should emit here also, to not miss any metrics

assets/training/aoai/proxy_components/src/common/azure_openai_client_manager.py
```diff
@@ -47,7 +47,7 @@ def _get_workspace_subscription_id_resource_group(self) -> str:
         """Get current subscription id."""
         uri = os.environ.get(AzureOpenAIClientManager.MLFLOW_TRACKING_URI, None)
         if uri is None:
-            return None
+            return None, None
```
amangupta26, line 50: Why this is needed? Return type is just `str` for this function

chiragbhatt311: actually return type is tuple[str,str], fixed the return type

assets/training/aoai/proxy_components/src/finetune_submit.py
chiragbhatt311: calling one more time after status becomes terminal

PR #2685, opened 2024-04-12
Dev Tool: Create RAG components with private environment

shpengmsft:
This script will create/update custom components based on the spec.yaml files. 

- Automatically increase version by "0.0.1" if the component already exits.
- 
Example:
Usage:

1. Using the latest version of environment.
 python create_rag_components.py --environment-string=private-embedding-env

2. Using speicific version of environment.
 python create_rag_components.py --environment-string=private-embedding-env:10

3. Force all component to specific version.
 python create_rag_components.py --environment-string=private-embedding-env ----component-version=1.2.3

### DIFF
--- a/assets/large_language_models/utils/create_rag_components.py
+++ b/assets/large_language_models/utils/create_rag_components.py
@@ -1,0 +1,189 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""
+Create RAG components pointing to a custom environment.
+
+Usage:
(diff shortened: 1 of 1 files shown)


jinzhaochangAtMicrosoft (commented):
1) Would this accidentally overwrite the official component in use?
2) Can we specify only one of the component? For example, what if I only want to test `validate_deployments`.
3) Should we copy and paste the azureml-rag repo so that the component can pick up the latest changes? For example, if we update `validate_deployments.py` in azureml-rag repo and then want to test `validate_deployments` component.

Adarsh-Ramanathan (approved):

PR #2752, opened 2024-04-23
Component for benchmarking embedding models

iamrk04:


assets/aml-benchmark/components/benchmark_embedding_model/spec.yaml
```diff
@@ -0,0 +1,71 @@
...
+name: benchmark_embedding_model
+display_name: Benchmark Embedding Model
+description: Component for benchmarking an embedding model via MTEB.
+version: 0.0.1
+
+inputs:
+  endpoint_url:
+    type: string
+    optional: True
+    description: For AOAI, the base endpoint url. For OAI, this will be ignored. For OSS_MaaS, the target url.
```
arun-rajora, line 13: Canf AOAI be same as OSS_MaaS? Do we need different handling of the two?

assets/aml-benchmark/components/benchmark_embedding_model/spec.yaml
```diff
@@ -0,0 +1,71 @@
...
+  deployment_type:
+    type: string
+    optional: False
+    description: The deployment type.
+    enum:
+      - AOAI
+      - OAI
+      - OSS_MaaS
```
arun-rajora, line 21: Same here, can OSS_MaaS be removed and AOAI be used instead?

arun-rajora (changes requested):

assets/aml-benchmark/components/benchmark_embedding_model/spec.yaml
iamrk04: There is a mandatory parameter that Cohere models require, which is not the case with AOAI models. Not sure if other OSS models coming as MaaS will have some other parameter like this. Hence, separated these out. Also, the implementation for calling the endpoint is different for these types.

arun-rajora (approved):

PR #2757, opened 2024-04-23
Remove safe to test label after validation

vizhur:


.github/workflows/check-execution-context.yaml
```diff
@@ -32,6 +32,11 @@ jobs:
         run: |
           echo "::error::This PR was created from a fork and requires the 'safe to test' label to run this check."
           exit 1
+
+      - name: Remove label
+        uses: actions-ecosystem/action-remove-labels@v1
```
elesel, line 37: Please add the check above so that we only attempt to remove the label if it's already present.

vizhur: it will fail above if it is missing

elesel: I think this step would run if the workflow is started manually (`github.event_name == 'workflow_dispatch') and not from a fork.

elesel (approved):

PR #2764, opened 2024-04-23
Allow null input/output in trace aggregator output schema

alanpo1:


assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
```diff
@@ -155,3 +155,26 @@ def _filter_df_by_time_window(df: DataFrame, data_window_start: datetime, data_w
...
+
+def _count_dropped_rows_with_error(
+        original_df_row_count: int,
+        transformed_df_row_count: int,
+        additional_error_msg: str = ""):
+    """Calculate number of dropped rows after transformations and throw error if too many rows lost."""
+    if original_df_row_count <= 0:
+        print("Original df contains no rows. Returning without calculating drop rate.")
+        return
+    drop_rate = abs(transformed_df_row_count - original_df_row_count) / original_df_row_count
```
RichardLi1437, line 168: why abs? Isn't original_df_row_count always >= transformed_df_row_count?

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
```diff
@@ -155,3 +155,26 @@ def _filter_df_by_time_window(df: DataFrame, data_window_start: datetime, data_w
...
+    print(f"Calculated the df drop rate as {drop_rate*100}%. Comparing to threshold criteria...")
+
+    if drop_rate > 0.10 and drop_rate < 0.67:
```
RichardLi1437, line 171: how do we come up with this value(0.67)? We still continue with 2/3 rows invalid and dropped? this value seems too large for me.

assets/model_monitoring/components/src/model_data_collector_preprocessor/trace_aggregator.py
```diff
@@ -8,13 +8,31 @@
...
+    """Check specific validations for aggregated trace entries as necessary."""
+    print("Validating aggregated trace logs...")
+    # some PF logs that result in exception won't have output so need to drop traces that don't have output or input.
+    transformed_df = aggregated_traces_df.dropna(subset=["input", "output"])
+
+    _count_dropped_rows_with_error(
+        aggregated_traces_df.count(),
+        transformed_df.count(),
+        additional_error_msg="Additionally, the step that caused issues was validating trace logs input/output."
+        " Double check the stdout and spark executor logs for debug info to find root cause of issue.")
```
RichardLi1437, line 31: we print out the invalid rows? otherwise how can user check stdout to debug the root cause?

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
alanpo1: I just copied this from a different PR where we suggested 67% as a threshold for failing the input schema adaptor job. I am open to other values, do you have one in mind? We can also consult PM/Data scientists what a good threshold value is for failing the job

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
alanpo1: I think so but It was just to be safe in case someone called this function passing in bad values

assets/model_monitoring/components/src/model_data_collector_preprocessor/trace_aggregator.py
alanpo1: I will add some debug logs for traces that are thrown out. They will be in the executor logs mainly

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_preprocessor_df_schemas.py
```diff
@@ -38,8 +38,8 @@ def _get_aggregated_trace_log_spark_df_schema() -> StructType:
             StructField("session_id", StringType(), True),
             StructField("start_time", TimestampType(), False),
             StructField("end_time", TimestampType(), False),
-            StructField("input", StringType(), False),
-            StructField("output", StringType(), False),
+            StructField("input", StringType(), True),
```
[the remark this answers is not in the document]
alanpo1: Switched the schema back.

assets/model_monitoring/components/src/model_data_collector_preprocessor/trace_aggregator.py
alanpo1: Added more debug logs in aggregation function

alanpo1: Edited the debug message

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
RichardLi1437: firstly, this is an internal API, we should ensure we don't make such mistake, secondly, if that really happen, abs won't ensure a correct value, we should just raise error.

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
RichardLi1437: let's start with 0.33, we can consult PM/DS and update this value if they have a reasonable recommendation.

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
alanpo1: Okay that makes sense. I'll remove abs()

assets/model_monitoring/components/src/model_data_collector_preprocessor/mdc_utils.py
alanpo1: sounds good

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_preprocessor_df_schemas.py
RichardLi1437, line 41: this is all const values, why not define a const like variable, e.g. PREPROCSSED_SPA_SCHEMA, AGGREGATED_TRACE_SCHEMA?
Under current implementation, the const schema is constructed every time you call this function, which is un-necessary

RichardLi1437 (approved):

assets/model_monitoring/components/src/model_data_collector_preprocessor/genai_preprocessor_df_schemas.py
alanpo1: followed the output schema pattern from another component. Can code refactor later

assets/model_monitoring/components/src/model_data_collector_preprocessor/trace_aggregator.py
```diff
@@ -22,33 +22,45 @@
...
-        return None
     return tuple(output_dict.get(fieldName, None) for fieldName in output_schema.fieldNames())
 
 
-def _aggregate_span_logs_to_trace_logs(grouped_row: Row, error_rate_accumulator: Accumulator[int]):
+def _validate_traces_df(aggregated_traces_df: DataFrame) -> DataFrame:
+    """Check specific validations for aggregated trace entries as necessary."""
+    print(f"{logging_prefix_str} Validating aggregated trace logs...")
+    # some PF logs that result in exception won't have output so need to drop traces that don't have output or input.
+    transformed_df = aggregated_traces_df.dropna(subset=["input", "output"])
```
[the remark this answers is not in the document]
alanpo1: Might refactor later. My assumption was going forward we might add more validations in some way so the transformed_df was left broad on purpose

PR #2808, opened 2024-04-30
Publush cohere command and cohere command r plus results

ankushbhatia2:


arun-rajora (approved):

assets/evaluation_results/boolq_cohere_command_r_question_answering/spec.yaml
```diff
@@ -0,0 +1,31 @@
+type: evaluationresult
+name: boolq_cohere_command_r_question_answering
+version: 2.30.04
```
ShipraJain01, line 3: please verify the versions of these assests across..

In my view it should be 1.0.0 since we are publishing the assets for the first time..

ankushbhatia2: For the others as well, the version is the date they were updated/uploaded.

PR #2833, opened 2024-05-03
upgrading component versions to fix phi3 deepcopy issue

jpmann:
Changes
---------
1. Upgraded env for chat completion components.
2. Rest of the component versions remain as it is.
3. Passing "messages" key to the data validation components.
4. Removing the hack to remove the "return_full_text" key in model convertor adapters. This is already handled during preprocessing (pypi release should have these changes).


skanakamedal (approved):

Anubha98 (approved):

amangupta26 (approved):
We can upgrade env for text gen too, but this release is focused for chat completion so can be done in subsequent release as well.

PR #2861, opened 2024-05-08
action detector fix null index id

ycheng35xo:


assets/model_monitoring/components/src/action_analyzer/action_detector_component/run.py
```diff
@@ -47,15 +47,15 @@ def parse_hashed_index_id(root_span: str) -> List[str]:
...
+                    print("No look up span found, skip this span.")
+                    continue
                 index_span = tree.get_span_tree_node_by_span_id(parent_id)
                 index_input = json.loads(index_span.input)
                 index_content = index_input['mlindex_content']
                 hashed_index_id = hash_index_content(index_content)
                 if hashed_index_id:
                     index_list.append(hashed_index_id)
-        return index_list
+        return index_list if index_list != [] else None
```
yunjie-hub, line 58: if you return None here, would it fail later?

yunjie-hub: I see you have a unit test to cover, so explode will fail with empty list but works withe None?

ycheng35xo: later added drop None. So it will be dropped.  Also UT tested this scenario

ycheng35xo: no explode will not fail. The issue is the unique_index list will become ["indexid", nan], which is unexpected.

yunjie-hub (approved):

PR #2864, opened 2024-05-08
Phi-2 model with new vllm image v35

BhavanaTumma:


skasturi (approved):

sanchez-alex:
Can we remove the ds_mii tag as well since that engine is not working anymore with this model?

PR #2910, opened 2024-05-15
Vision dataset downloading

rdondera-microsoft:
Support downloading vision + multimodal datasets in the dataset downloader component.

Implementation
It is not possible to directly save vision HF datasets to JSONL (as is done for NLP) because binary image data is not compatible with this setup. The solution was to keep the JSONL format but use a URL field that refers to image data saved to the default datastore. For each HF dataset, a specific adapter class extracts a label and a PIL image for a dataset instance and common code uploads the image, passes through the fields etc.

Extensions
New vision/multimodal datasets from HF can be added by adding new dataset adapters.

Tests
Ran the benchmarking pipeline e2e in the cloud: https://ml.azure.com/experiments/id/9f7237b4-3232-49d3-ae02-e07326242835?wsid=/subscriptions/dbd697c3-ef40-488f-83e6-5ad4dfb78f9b/resourcegroups/rdondera/providers/Microsoft.MachineLearningServices/workspaces/benchmarking&tid=72f988bf-86f1-41af-91ab-2d7cd011db47 runs "resisc45", "food101", "patch_camelyon", "gtsrb".


assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
```diff
@@ -20,6 +20,10 @@
 
 logger = get_logger(__name__)
 
+DATASTORE_DIRECTORY_URL_TEMPLATE = "AmlDatastore://{datastore_name}/{directory_name}"
+RANDOM_IMAGE_DIRECTORY_TEMPLATE = "images/{random_id}"
+IMAGE_FILE_NAME_TEMPLATE = "image_{image_counter:09d}.png"
+
```
iamrk04, line 26: can we please move them to `constants.py`?

assets/aml-benchmark/components/src/aml_benchmark/dataset_downloader/vision_dataset_adapter.py
```diff
@@ -0,0 +1,127 @@
...
+    def get_pil_image(self, instance):
+        """Extract the instance's image as a PIL image."""
+        return Image.open(io.BytesIO(instance["Path"]["bytes"]))
+
+
+class VisionDatasetAdapterFactory:
+    """Factory for making vision dataset adapters based on dataset names."""
+
+    @staticmethod
+    def get_adapter(dataset: Dataset) -> VisionDatasetAdapter:
```
iamrk04, line 109: Since, we are retuning `None`, can we update the return type from `VisionDatasetAdapter` to `Optional[VisionDatasetAdapter]`?

iamrk04 (changes requested):

assets/aml-benchmark/components/src/aml_benchmark/utils/io.py
rdondera-microsoft: Sure.

assets/aml-benchmark/components/src/aml_benchmark/dataset_downloader/vision_dataset_adapter.py
rdondera-microsoft: Nice catch, thank you! Done. (wondering why this didn't trigger mypy or other analyses)

iamrk04 (approved):

PR #2933, opened 2024-05-17
Skip hidden layer scanned tag valdn for models

novaturient95:
Currently, we are checking hidden layer scanned tag for all models. Skip until we have done for all models & not do it components.


vikasagrawal-ms (approved):

skasturi (approved):

ManojBableshwar (approved):
Confirmed with Sasi that "Checking for "hidden layer" was added to components also.  
This skips that check"

PR #2969, opened 2024-05-24
Add AzureMLOnBehalfOfCredential  to AOAI finetuning component

f-salvetti:
This PR adds AzureMLOnBehalfOfCredential credential to AOAI finetuning component. If getting a token with this credential fails, it will fallback to the ManagedIdentityCredential of the compute


tengwenxuan:
I saw some build error with "Error: This PR was created from a fork and requires the 'safe to test' label to run this check." who can set this label? Is this something that only repo owners can set? thx

PR #2985, opened 2024-05-29
update requests to >=2.32.0 to fix vulnerability

qusongms:
vulnerability issue: GHSA-9wx4-h78v-vm56


vizhur (changes requested):
That seems unnecessary, rebuild will take care of requests

PR #2991, opened 2024-05-30
Generate SAS tokens for data assets

lykelly19:


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -1295,7 +1343,7 @@ def extra_config_with_path(self) -> Path:
         config = self.extra_config
         return self._append_to_file_path(config) if config else None
 
-    def extra_config_as_object(self, force_reload: bool = False) -> EnvironmentConfig:
+    def extra_config_as_object(self, force_reload: bool = False) -> Config:
```
elesel, line 1346: Good catch!

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -1314,6 +1362,8 @@ def extra_config_as_object(self, force_reload: bool = False) -> EnvironmentConfi
                     self._extra_config = EnvironmentConfig(extra_config_with_path)
                 elif self.type == AssetType.MODEL:
                     self._extra_config = ModelConfig(extra_config_with_path)
+                elif self.type == AssetType.DATA:
+                    self._extra_config = DataConfig(extra_config_with_path)
```
elesel, line 1366: Some (small) data assets only use local storage. How should they be handled? Here's one example: https://github.com/Azure/azureml-assets/blob/6446f5fd809a57a4e39cab1d7c56e2c9a809efd6/assets/promptflow/data/ask-wikipedia/asset.yaml

elesel (approved):

scripts/azureml-assets/azureml/assets/config.py
lykelly19: discussed via chat, no extra config will be created for local storage

PR #3002, opened 2024-06-03
use StoreUrl to save spark dataframe instead of azureml file system

alanpo1:
When we are trying to upload data to a DataLakeGen2 datastore we need to first check if the file exists and if not then create and append the data to the file. THat is why I add the extra code path on StoreUrl for `FileSystemClient`. Otherwise, we get a bug that the resource does not exist when trying to upload data to gen2 storage.

**Note:** these are some examples of jobs where I test replacing the action analyzer code with StoreUrl.
**before** changes:
https://ml.azure.com/data/azureml_c2451173-4a01-48f1-ad87-5997129b6f6a_output_data_action_output/1/details?wsid=/subscriptions/1aefdc5e-3a7c-4d71-a9f9-f5d3b03be19a/resourcegroups/yuachengtestrg/workspaces/ai-proj-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

**after** changes:
https://ml.azure.com/data/azureml_70c8600f-7484-46de-a679-c59c9fd90252_output_data_action_output/1/details?wsid=/subscriptions/1aefdc5e-3a7c-4d71-a9f9-f5d3b03be19a/resourcegroups/yuachengtestrg/workspaces/ai-proj-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

**Test with using old StoreUrl code:**
image

**Test with using new create_fil() and append_data() code:**
image


assets/model_monitoring/components/src/action_analyzer/action_analyzer_output_actions/run.py
```diff
@@ -4,20 +4,22 @@
...
-from shared_utilities.io_utils import try_read_mltable_in_spark, np_encoder
+from shared_utilities.store_url import StoreUrl
+from shared_utilities.io_utils import try_read_mltable_in_spark
 import os
-import json
 import uuid
 import datetime
 import copy
 from mlflow import MlflowClient
-from shared_utilities.amlfs import amlfs_upload
```
ycheng35xo, line 20: are we going to remove the dependency on azureml.fsspec? 

1. are we going to remove the amlfs.py?
2. after that, are we going to remove the conda package azureml-fsspec?

alanpo1: 1. No, the create manifest and output metrics components will still need to use functions from amlfs.py that are not supported by StoreUrl such as downloading/uploading whole folders. For these components I have to manually set some env vars to support credential-less data save/read. In most other scenarios I am moving our code to use StoreUrl which will handle credential-less datastore support on-demand in a cleaner way than adhoc env variables initialization.
2. I was considering this but due to the cache issues we have seen, I think I'll leave it in place for now. Can reconsider removing that dependency later on if we can depend on cache being updated.

alanpo1: I am testing to see which components we can remove fsspec dependency from in case we want to remove in this PR.

assets/model_monitoring/components/src/shared_utilities/io_utils.py
```diff
@@ -177,34 +163,10 @@ def read_mltable_in_spark(mltable_path: str):
...
-        # TODO: remove this check block after we switch over to using StoreUrl for saving spark df
-        if isinstance(error, Py4JJavaError):
-            if "Access token couldn't be obtained" in str(error):
-                raise InvalidInputError(
-                    MISSING_OBO_CREDENTIAL_HELPFUL_ERROR_MESSAGE.format(message=error.java_exception.getMessage()))
-        raise error
-
     base_path = folder_path.rstrip('/')
-    output_path_pattern = base_path + "/*.parquet"
+    output_path_pattern = base_path + "/data/*.parquet"
```
RichardLi1437, line 169: why add /data sub folder?

assets/model_monitoring/components/src/shared_utilities/io_utils.py
```diff
@@ -213,13 +175,15 @@ def save_spark_df_as_mltable(metrics_df, folder_path: str, file_system=None):
...
     while True:
-        if _write_mltable_yaml(mltable_obj, folder_path, file_system):
+        if _write_mltable_yaml(mltable_obj, folder_path):
             break
         retries += 1
         if retries >= MAX_RETRY_COUNT:
             raise Exception("Failed to write mltable yaml file after multiple retries.")
         time.sleep(1)
 
+    metrics_df.write.mode("overwrite").parquet(base_path+"/data/")
```
RichardLi1437, line 185: why removed the original try except logic?

assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -234,7 +234,9 @@ def write_file(self, file_content: Union[str, bytes], relative_path: str = None,
         try:
             if isinstance(container_client, FileSystemClient):
                 with container_client.get_file_client(full_path) as file_client:
-                    return file_client.upload_data(file_content, overwrite)
+                    if not file_client.exists():
```
RichardLi1437, line 237: can we refactor the code as:
```py
with _get_file_client(container_client, full_path) as file_client:
  return file_client.upload_data(...)
```
and _get_file_client() defined as:
```py
def _get_file_client(filesystem_client):
  file_client = filesystem_client.get_file_client(_client(full_path)
  if not file_client.exists():
    file_client = filesystem_client.create_file(full_path)
  return file_client
```
You use append_data for a new created file_client, I suppose we can also use upload_data?

assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -254,6 +256,17 @@ def _read_local_file_content(self, relative_path: str = None) -> str:
         with open(full_path) as f:
             return f.read()
 
+    def _create_file_and_append_content(
```
RichardLi1437, line 259: we don't need this method if the refactored code works

assets/model_monitoring/components/src/shared_utilities/io_utils.py
alanpo1: Because when I did the MLTable write first (to enable credential-less with StoreUrl) and then did `spark.write.parquet` in the same location it ended up overwriting and deleting the MLTable, so if we write to the data subfolder we avoid this issue.

assets/model_monitoring/components/src/shared_utilities/io_utils.py
alanpo1: We don't need to catch the credential-less unavailable errors anymore. I flipped the logic order to write the MLTable yaml file first using StoreUrl so if credential-less doesn't work it will fail at that step. Now, when we get to the `spark.write` credential-less should work or it would have failed before.

assets/model_monitoring/components/src/action_analyzer/action_analyzer_output_actions/run.py
```diff
@@ -4,20 +4,22 @@
 """Entry script for Action Analyzer output actions."""
 
 import argparse
+from action_analyzer.contracts.utils.detector_utils import (
```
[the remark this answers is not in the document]
alanpo1: Not sure, I just noticed that the write_file() function was the same in both files so I decided to import it and remove duplicates.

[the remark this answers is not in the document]
ycheng35xo: we did not set a well-organized folder structure when writing the action detector code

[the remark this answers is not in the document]
alanpo1: Will change later

assets/model_monitoring/components/src/shared_utilities/store_url.py
alanpo1: I tried to use this refactor but upload_data() is still an issue on writing to Gen2 storage. I have only got it to work with append_data() and flush_data() with the create_file() client method. But then for future operations after the file is created successsfully we can use upload_data() to modify/overwrite in the FileSystemClient. I'm not sure why this gap exists just that it does

alanpo1: I'll stick to the current code that uses append and flush data. Do you have any other suggestions?

assets/model_monitoring/components/src/shared_utilities/store_url.py
alanpo1: See other comment.

assets/model_monitoring/components/src/shared_utilities/store_url.py
RichardLi1437: ok, maybe add some comments to explain why we have to do in this way.

RichardLi1437 (dismissed):

assets/model_monitoring/components/src/shared_utilities/store_url.py
alanpo1: Added comment

ycheng35xo (approved):

PR #3016, opened 2024-06-05
Enable input column, label column for vision tasks

rdondera-microsoft:
Context (benchmarking the Stable Diffusion model on MSCOCO): https://github.com/Azure/azureml-assets/compare/main...rdondera/sd

Honor column name parameters (inputs, label) passed to the model prediction component for vision tasks. This does not bring benefits for image classification, object detection or instance segmentation, but is needed for image generation tasks, where the input name depends on the model.

Also: clarified how vision dataset processing works in the model evaluation pipeline. The MLTable is assumed to have columns "image_url" and "label" and the Pandas dataframe created from it has columns set according to the input_column_names and label_column_name parameters. When these parameters are not specified, task specific default values are filled in.

Also: brought back changes from https://github.com/Azure/azureml-assets/pull/2888 , which were removed in a subsequent PR.

Tested by running the evaluation pipeline for ViT in the cloud: https://ml.azure.com/experiments/id/9f7237b4-3232-49d3-ae02-e07326242835?wsid=/subscriptions/dbd697c3-ef40-488f-83e6-5ad4dfb78f9b/resourcegroups/rdondera/providers/Microsoft.MachineLearningServices/workspaces/benchmarking&tid=72f988bf-86f1-41af-91ab-2d7cd011db47 , vit_*.
1. vit_params: both input_column_names and label_column_name properly set ("image", "label")
2. vit_params: neither input_column_names nor label_column_name set
3. vit_wrong_params: both input_column_names and label_column_name set to wrong values ("rgb_image", "class")
4. vit_wrong_label: input_column_names properly set, label_column_name set to wrong value ("image", "class,attribute")


assets/training/model_evaluation/components/model_prediction/spec.yaml
```diff
@@ -3,7 +3,7 @@ name: model_prediction
 display_name: Model Prediction
 description: Generate predictions on a given mlflow model for supported tasks.
 
-version: 0.0.28
+version: 0.0.29
```
rjaincc, line 6: We would need to bump up the version in the pipeline component and bump up the version of the pipeline component.

rdondera-microsoft: Don't have enough context here to understand. Is there another file that needs to be modified?

ankushbhatia2 (approved):

assets/training/model_evaluation/components/model_prediction/spec.yaml
rdondera-microsoft: Per Teams discussion, incremented version in pipeline_component/spec.yaml .

ankushbhatia2 (approved):

PR #3032, opened 2024-06-11
Hotfix import of AzureMLOnBehalfOfCredential to only credential-less datastore scenario.

alanpo1:
Hotfix for this PR here: https://github.com/Azure/azureml-assets/pull/2929

We saw some issue with synapse cache using OLD cache dependency instead of NEW cache which included azure-ai-ml. 

To remedy this need to move imports from new **azure.ai.ml** module to be inside of credential-less datastore code path to not impact old code path where we don't necessarily need to import from new conda module.

Note: for changing the try/catch from **CredentialUnavailableError** to an **Exception** here is an example job showing we still catch that error fine: https://ml.azure.com/runs/lime_branch_bz7qzy3b0s?wsid=/subscriptions/1aefdc5e-3a7c-4d71-a9f9-f5d3b03be19a/resourcegroups/momo-rampup-alanpo/workspaces/project-alantest-oai&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -96,20 +97,31 @@ def get_credential(self) -> Union[
...
                                     "only Azure Blob and Azure Data Lake Gen2 are supported.")
 
     def is_credentials_less(self) -> bool:
         """Check if the store url is credential less."""
-        credential = self.get_credential()
+        # TODO: remove after we figure out cache failure issues.
+        # Should be able to import AzureMLOnBehalfOfCredential to check the class w/o issues.
+        credential = None
+        try:
+            credential = self.get_credential()
```
ycheng35xo, line 112: can you print some log for enabling credentialess path and use the managed identity successfully (or maybe we already have that in your first PR)?  Because it can help us debug our managed identity works as expected.

alanpo1: I added some log statements when we return the credential to know what type it is, this should let us know if the StoreUrl is using credential-less path.

Also, from the first PR I added a helpful error message that is printed when the read/write fails and it is due to CredentialUnavailableError which afaik means the managed identity is not attached or the job was not submitted with "managed".

Otherwise, if the credential is OBO token and there are no errors saving/writing data then the momo job is using managed identity and is working as expected I think.

assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -77,39 +76,58 @@ def _get_url(self, scheme=None, store_type=None, relative_path=None) -> str:
         return url
 
     def get_credential(self) -> Union[
-            str, ClientSecretCredential, AzureSasCredential, AzureMLOnBehalfOfCredential, None]:
+            str, ClientSecretCredential, AzureSasCredential, None]:
         """Get credential for this store url."""
         if not self._datastore:
+            from azure.ai.ml.identity import AzureMLOnBehalfOfCredential
+            print("Using AML OBO credential from StoreUrl.get_credential().")
```
ycheng35xo, line 83: minor: line83, 94, 106 are same, can we distinguish them for better debugging purpose?

alanpo1: Yes that makes sense. Added.

assets/model_monitoring/components/src/model_monitor_create_manifest/run.py
```diff
@@ -27,10 +27,19 @@ def _generate_manifest(root_dir: str):
...
 def run():
     """Create Manifest."""
-    # init aml OBO class for supporting credential-less datastore scenarios.
-    # this class will init specific env variables required by Amlfs to get
-    # user token in case of credential-less data access scenario
-    AzureMLOnBehalfOfCredential()
+    # TODO: investigate why these aren't initialized by AzureML Filesystem itself
+    # init env-vars for supporting credential-less datastore scenarios.
+    # this env variable is required by Amlfs to get user token.
+    spark = init_spark()
```
RichardLi1437, line 33: we replace AmlOBOCredential() with below env var setup code? You copy some initialization code from there?

assets/model_monitoring/components/src/model_monitor_create_manifest/run.py
```diff
@@ -27,10 +27,19 @@ def _generate_manifest(root_dir: str):
...
+    spark_conf = spark.sparkContext.getConf()
+    spark_conf_vars = {
+        "AZUREML_SYNAPSE_CLUSTER_IDENTIFIER": "spark.synapse.clusteridentifier",
```
RichardLi1437, line 36: I suppose these are only place holder values? They must be real cluster Id and endpoint url to take effect? If so, why we need to add them here?

assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -77,39 +76,60 @@ def _get_url(self, scheme=None, store_type=None, relative_path=None) -> str:
...
     def is_credentials_less(self) -> bool:
         """Check if the store url is credential less."""
-        credential = self.get_credential()
+        # TODO: remove after we figure out cache failure issues.
+        # Should be able to import AzureMLOnBehalfOfCredential to check the class w/o issues.
+        credential = None
+        try:
+            credential = self.get_credential()
+            from azure.ai.ml.identity import AzureMLOnBehalfOfCredential
+        except ModuleNotFoundError:
```
RichardLi1437, line 122: suggest to arrange the try except as below:
```py
try:
  credential = self.get_credential()
  from ... import AmlOboCrednetial
  return credetnial is Non or isinstance(...)
except ModuleNotFoundError:
  print(...)
  return credential is None
```

assets/model_monitoring/components/src/model_monitor_create_manifest/run.py
alanpo1: No not place holder. We get the real values from the spark conf with the key on right-side and set the environment variable with key on left-side to the spark conf value.

assets/model_monitoring/components/src/model_monitor_create_manifest/run.py
alanpo1: Yes I grabbed this code from the internals of AML obo class. If we want to avoid the import issue caused by old cache this is the only way to still support the credential-less save scenario on this component

assets/model_monitoring/components/src/shared_utilities/store_url.py
alanpo1: Done.

assets/model_monitoring/components/src/model_monitor_create_manifest/run.py
alanpo1: In reality this is probably a simpler and clearer way than having the adhoc aml obo credential instantiation.

RichardLi1437: consider to make it a util function

RichardLi1437 (approved):
have you tested with azure.ai uninstalled and the credential scenario works?

alanpo1:
> have you tested with azure.ai uninstalled and the credential scenario works?

I have a test running where I manually removed azure-ai-ml from the pip dependencies of all components and ran the create manifest e2e test. It is not entirely comprehensive of the changes made but it covers a large percent of the changes.

credential-less code-path Job w/o azure-ai-ml (expect failure): https://ml.azure.com/runs/good_bridge_yg74c7hr7d?wsid=/subscriptions/1aefdc5e-3a7c-4d71-a9f9-f5d3b03be19a/resourcegroups/momo-rampup-alanpo/workspaces/Project-alantest-oai&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

credential code-path job w/o azure-ai-ml (expect success): https://ml.azure.com/runs/upbeat_farm_vb0b49b3n2?wsid=/subscriptions/5fca341e-4ec3-45bd-b006-a7648d9376c5/resourcegroups/modelmonitoring-e2e-rg/workspaces/momo-e2e-test-ws-eastus&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

PR #3080, opened 2024-06-21
Add prompt(y) for generating user questions

jfomhover:


assets/large_language_models/prompts/chat_user_questions/prompt/prompt.prompty
```diff
@@ -0,0 +1,64 @@
...
+  persona: a confused customer, not proficient in web interfaces
+  context: contoso outdoors shopping website
+  product: hiking boots
+  count: 5
+  questions: >
+    {"question":"How do I find the right size of hiking boots for my feet on this website?"}
+    {"question":"Are there instructions available for how to measure my foot for the correct boot fit?"}
+    {"question":"Can someone guide me through the process of selecting a pair of hiking boots?"}
+    {"question":"I'm not sure what the different boot terms mean, is there a glossary or help section?"}
+    {"question":"What is the return policy in case the hiking boots don't fit well when I receive them?"}
```
sethjuarez, line 45: this should look something like this:
```yaml:
questions:
   - How do I find the right size of hiking boots for my feet on this website?
   - Are there instructions available for how to measure my foot for the correct boot fit?
```
In general things should be yaml formatted (unless folks define things externally in files)

jfomhover: in this case, I'm asking the model to output the results as a JSONL string, so this {} structure is what the model actually outputs

jfomhover: modified the prompts to move example i/o into the description

wayliums (approved):

PR #3119, opened 2024-07-04
check folder existence before call get_paths()

RichardLi1437:
fix ResourceNotFound error for gen2 store

### DIFF
--- a/assets/model_monitoring/components/src/shared_utilities/store_url.py
+++ b/assets/model_monitoring/components/src/shared_utilities/store_url.py
@@ -223,8 +223,11 @@ class StoreUrl:
         if not container_client:  # local
             return any(glob.iglob(full_path_pattern))
         if isinstance(container_client, FileSystemClient):  # gen2
-            paths = container_client.get_paths(non_wildcard_path, True)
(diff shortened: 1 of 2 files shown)


assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -223,8 +223,11 @@ def any_files(file_names: list, pattern):
         if not container_client:  # local
             return any(glob.iglob(full_path_pattern))
         if isinstance(container_client, FileSystemClient):  # gen2
-            paths = container_client.get_paths(non_wildcard_path, True)
-            file_names = [path.name[len(non_wildcard_path):] for path in paths if not path.is_directory]
+            if container_client.get_directory_client(non_wildcard_path).exists():
```
ycheng35xo, line 226: Only gen2 needs this check? How about blob

ycheng35xo (approved):

assets/model_monitoring/components/src/shared_utilities/store_url.py
RichardLi1437: blob has no directory concept, it just like blobs starts with a specific prefix.

PR #3158, opened 2024-07-17
Add check for UAI has no permission to storage account and fix bug for valid_aml_obo_credential()

ycheng35xo:
The new error handling will be like this:
image


assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -91,7 +92,8 @@ def valid_aml_obo_credential():
                 aml_obo_credential = AzureMLOnBehalfOfCredential()
                 if validate_aml_obo_credential:
                     try:
-                        _ = aml_obo_credential.get_token()
+                        _ = aml_obo_credential.get_token("https://management.azure.com/.default")
+                        return aml_obo_credential
                     except CredentialUnavailableError as cue:
                         raise InvalidInputError(MISSING_OBO_CREDENTIAL_HELPFUL_ERROR_MESSAGE.format(message=str(cue)))
                 else:
```
RichardLi1437, line 99: shouldn't you remove `else` so that it will return aml_obo_credential after validation check?

assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -225,15 +227,21 @@ def any_files(file_names: list, pattern):
...
+            if isinstance(container_client, FileSystemClient):  # gen2
+                if container_client.get_directory_client(non_wildcard_path).exists():
+                    paths = container_client.get_paths(non_wildcard_path, True)
+                    file_names = [path.name[len(non_wildcard_path):] for path in paths if not path.is_directory]
+                else:
+                    return False
+            else:  # blob
+                blobs = container_client.list_blobs(name_starts_with=non_wildcard_path)
+                file_names = [blob.name[len(non_wildcard_path):] for blob in blobs]
+        except HttpResponseError as hre:
```
RichardLi1437, line 240: need to check if the credential is AmlOboCredential to raise this UMAI error msg, otherwise, it could be SAS token doesn't have enough permission/expired or incorrect account key/client_id/client_credential, etc.

assets/model_monitoring/components/src/shared_utilities/store_url.py
```diff
@@ -91,7 +92,8 @@ def valid_aml_obo_credential():
                 aml_obo_credential = AzureMLOnBehalfOfCredential()
                 if validate_aml_obo_credential:
                     try:
-                        _ = aml_obo_credential.get_token()
+                        _ = aml_obo_credential.get_token("https://management.azure.com/.default")
```
RichardLi1437, line 95: do we need the /.default for the scope?

ycheng35xo: I think either should be fine. I saw most codes in this repo has the /.default, so just keep for consistency

https://github.com/search?q=repo%3AAzure%2Fazureml-assets+get_token%28&type=code

assets/model_monitoring/components/src/shared_utilities/store_url.py
ycheng35xo: fixed

assets/model_monitoring/components/src/shared_utilities/constants.py
```diff
@@ -201,7 +201,17 @@
...
+)
+
+ACCOUNT_KEY_MISS_PERMISSION_ERROR_MESSAGE = (
+    "Failed to use account key to get access of default storage account. "
+    "Please double check if the account key is valid."
+)
+
+SAS_TOKEN_MISS_PERMISSION_ERROR_MESSAGE = (
+    "Failed to use sas token to get access of default storage account. "
+    "Please double check if the sas token is valid and has enough permissions to access the storage account."
```
RichardLi1437, line 216: 'valid' -> 'not expired'

ycheng35xo: fixed

RichardLi1437 (approved):

PR #3194, opened 2024-07-24
Add system environment for image build and sbom generation

vizhur:


assets/system/environment.yaml
```diff
@@ -0,0 +1,16 @@
+image:
+  name: azureml/curated/docker-tools
```
jitupad, line 2: so image build and docker-utils in one single env?

jitupad (approved):

assets/system/environment.yaml
vizhur: correct

jitupad (approved):

elesel (approved):

edebar01 (approved):

PR #3230, opened 2024-08-07
Add support for adding tags from filepaths

acsoylemezoglu:


scripts/azureml-assets/azureml/assets/model/model_utils.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, spec_path, model_config, registry_name, temp_dir, copy_update
         try:
             self._model = load_model(spec_path)
             self._model.description = model_config.description
+            self._model.tags.update(model_config.extra_tags_from_files)
```
vizhur, line 67: def resolve_from_file(value):
    if os.path.isfile(value):
        with open(value, 'r') as f:
            content = f.read()
        return content
    else:
        return content

self._model.tags = {i, resolve_from_file(v) for i, v in self._model.tags.items()}

we can use that for all the values, not just tags (out of the scope of these changes)

scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -1318,6 +1318,19 @@ def description_file(self) -> str:
...
         return self._yaml.get('description_file')
 
+    @property
+    def extra_tags_from_files(self) -> Dict[str, str]:
+        extra_tags_from_files = {}
+
+        extra_tags_to_filepaths = self._yaml.get('extra_tags_from_files')
+
+        if extra_tags_to_filepaths:
+            for (tagKey, fpath) in extra_tags_to_filepaths.items():
```
lykelly19, line 1328: maybe `tag_key` for python conventions

andscho-msft (approved):

vizhur (approved):

PR #3241, opened 2024-08-09
Fix the crash in the dq component

yunjie-hub:


assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
```diff
@@ -394,7 +394,8 @@ def impute_categorical_with_mode(df: pyspark.sql.DataFrame, categorical_columns:
         )
 
         # Impute the missing values with the most frequent value
-        df = df.fillna({i: most_frequent})
+        if most_frequent is not None:
```
ycheng35xo, line 397: what if is null? Will it crash (like have some null exception) later?

ycheng35xo (approved):

assets/model_monitoring/components/src/data_quality_compute_metrics/compute_data_quality_metrics.py
yunjie-hub: It can work with the null value

PR #3243, opened 2024-08-09
Removed promptflow_creation_job from component pipelines

clomeli-ms:
Removed promptflow_creation_job from component pipelines. This step is not explicitly necessary.


akshaya-a (approved):
were there no tests that would be broken by this change?

runhli (approved):

shpengmsft (approved):

shpengmsft:
> were there no tests that would be broken by this change?

I don't think this will break anything. This changed pipeline definition will be rolled out to azureml-stage registry. I'm going to test that before deploying to production (azureml registry).

PR #3268, opened 2024-08-14
chore: introduce minor telemetry to distillation component

ShobithNandakumar:
### Description
PR introduces minor logging to time execution of functions and code blocks. In particular, we want to monitor time taken for batches and network calls, in this case time taken for inferencing.


babu-namburi (approved):
LGTM

assets/training/distillation/components/data_generation/src/generate_data.py
```diff
@@ -212,8 +214,19 @@ def _invoke_endpoint(url: str, key: str, data: dict) -> Response:
...
     }
-    response = requests.post(url, headers=request_headers, data=json.dumps(data))
-    return response
+
+    log_entry = log_entry or {}
+    idx = log_entry.get("idx", -1)
+    turn = log_entry.get("turn" ,-1)
+
+    # We don't want to log every request. Conditionally log some, to avoid overwhelming logs.
+    if idx % 10 == 0 and turn % 2 == 0:
```
skasturi, line 223: How do we track requests that are specifically taking long time?

ShobithNandakumar: I had earlier introduced a custom class which allowed us to track code blocks that only went _beyond a threshold_. I can re-introduce that and log those instead of logging every 10th record/alternate turn. (@babu-namburi thoughts?)

From this [run](https://ml.azure.com/experiments/id/50b4a44c-50f8-4e1e-9dc2-5d28f618957d/runs/474cd334-956b-453a-9511-4473fa37d215?wsid=/subscriptions/75703df0-38f9-4e2e-8328-45f6fc810286/resourceGroups/rg-sasumai/providers/Microsoft.MachineLearningServices/workspaces/sasum-westus3-ws&tid=7f292395-a08f-4cc0-b3d0-a400b023b0d2#outputsAndLogs) here, it seems like some requests go way beyond 30 seconds (I'm guessing because of the requests being processed sequentially, even though we are batching them). 

@skasturi @babu-namburi - is there a particular threshold beyond which we want to monitor long-running inference calls?

PR #3276, opened 2024-08-19
Bump version for oss_distillation_pipeline and chat_completion_pipeline

sharvin2187:
Updated the env to latest version

### DIFF
--- a/assets/training/distillation/components/pipeline/spec.yaml
+++ b/assets/training/distillation/components/pipeline/spec.yaml
@@ -259,7 +259,7 @@ jobs:
 
   oss_chat_completion_finetune:
     type: command
-    component: azureml:oss_chat_completion_finetune:0.0.19
+    component: azureml:oss_chat_completion_finetune:0.0.21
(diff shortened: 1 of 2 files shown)


scravikiran (approved):
lgtm

assets/training/distillation/components/pipeline/spec.yaml
```diff
@@ -259,7 +259,7 @@ jobs:
 
   oss_chat_completion_finetune:
     type: command
-    component: azureml:oss_chat_completion_finetune:0.0.19
+    component: azureml:oss_chat_completion_finetune:0.0.21
```
novaturient95, line 262: lets update the data import one as well post oss ft component release

vikasagrawal-ms: why are we jumping 2 versions?

sharvin2187: ayush: holding on here based on offline discussions
vikas: distillation component was running an older version already, we are using 19 but the latest is 21

PR #3279, opened 2024-08-20
[Enhancement]: Multi-node-support for MaaS

yeshsurya:
* Add barrier 
* Single run flag to subprocess function call


assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +368,109 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
 
 
-def _run_subprocess_cmd(cmd: List[str], component_name: str, completion_files_folder: str):
+def is_main_process():
+    """
+    To check if current process running this is master or rank_0 process.
+
+    :return: Boolean for whether this process is master.
+    """
+    return os.environ.get('AZUREML_PROCESS_NAME', 'main') in {'main', 'rank_0'}
```
yeshsurya, line 377: Check if environment variable will be supported in future.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +368,109 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+    num_processes: Number of process which need to reach barrier point.
+    """
+    with open(barrier_file, 'a+') as f:
+        f.seek(0)
+        lines = f.readlines()
+        process_count = len(lines)
+        process_name = os.environ.get('AZUREML_PROCESS_NAME', 'main')
+        logger.info(f'Process {process_name} at barrier, count is {process_count}, in barrier file {barrier_file}')
+        if process_count < num_processes:
+            f.write(f"{os.getpid()} reached the barrier\n")
```
asubra-ms, line 394: There seems to be a potential race condition here. Think about it as discussed:  if process 1 opens the file as of lines 388/389, proceeds to finish executing line 393 then gets context switched. At this point lets say num_processes is 3 and the file desc points to the EOF which is line 4.
Now process 2 starts executing the function, runs through to completion, exhausting its time slice of execution. It also gets num_processes as 3 and goes on to write its pid at line 4.
If process 1 gets scheduled back again, now it starts executing at line 394 - so will write its own pid and overwrite the previous write.
Isn't that a classic case of race condition at the MP level?
This can be solved by a guard bit (use another guard file) to access this file with the pid.

yeshsurya: Thinking to use https://pypi.org/project/portalocker/ for file based lock. It gives portability across windows and Linux with MP support

asubra-ms: I am not familiar with portalocker. But if it seems reasonable, please try it. Raise a PR with the new code or usage then. Will take a look at the new implementation then.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: same used in acft accelerator

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+    Control will halt till number to process reaching the execution point is less than a given number.
+
+    barrier_file: File used to create execution barrier.
+    num_processes: Number of process which need to reach barrier point.
+    """
+    process_name = os.environ.get('AZUREML_PROCESS_NAME', 'main')
+    with open(barrier_file, 'a+') as f:
+        portalocker.lock(f, portalocker.LOCK_EX)
+        try:
+            f.seek(0)
```
vkedia, line 392: is this seek needed? Won't a freshly opened file already be at offset 0.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+            lines = f.readlines()
+            process_count = len(lines)
+            if process_count < num_processes:
+                f.write(f"{os.getpid()} reached the barrier\n")
+                f.flush()
```
vkedia, line 397: I think you will need to call os.fsync also to gurantee persistence.
See
https://docs.python.org/3.12/library/os.html#os.fsync
https://portalocker.readthedocs.io/en/latest/index.html#tips

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+def is_main_process():
+    """
+    To check if current process running this is master or rank_0 process.
+
+    :return: Boolean for whether this process is master.
+    """
+    return os.environ.get('AZUREML_PROCESS_NAME', 'main') in {'main', 'rank_0'}
+
+
+def wait_at_barrier(barrier_file, num_processes):
```
vkedia, line 381: if any of the file i/o code throws any error will that get logged somewhere?

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+            process_count = len(lines)
+            if process_count < num_processes:
+                f.write(f"{os.getpid()} reached the barrier\n")
+                f.flush()
+                portalocker.unlock(f)
+                logger.info(f'Process {process_name} at barrier, count is {process_count},')
+                logger.info(f'in barrier file {barrier_file}')
+        finally:
+            portalocker.unlock(f)
+        with open(barrier_file, 'a+') as f:
```
vkedia, line 403: here you can open as a readonly file instead of a+

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+                raise ACFTValidationException._with_error(
+                    AzureMLError.create(
+                        ACFTUserError,
+                        pii_safe_message=(
+                            f"{component_name} failed"
+                        )
+                    )
+                )
+            logger.info(f"{component_name} completed successfully")
+            Path(completion_files_folder).mkdir(parents=True, exist_ok=True)
```
vkedia, line 448: you already created this folder above so you can skip creating here.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+            Path(completion_file).touch()
+            logger.info(f"Created completion file: {completion_file}")
+        logger.info(f"Waiting for completion file: {completion_file}, by rank : {process_name}")
+        while not completion_file.exists():
```
vkedia, line 452: I am not able to understand why is this needed? It will anyways wait at the barrier below

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+        logger.info(f"Process {os.getpid()} has passed barrier with name {process_name}")
+
+
+def _run_subprocess_cmd(cmd: List[str], component_name: str, completion_files_folder: str,
+                        single_run=True, number_of_processes=1):
     """Run the subprocess command."""
     logger.info(f"Starting the command: {cmd}")
     completion_file = Path(completion_files_folder, f"{component_name}.complete.txt")
+    barrier_file = Path(completion_files_folder, f"{component_name}.barrier.txt")
+    Path(completion_files_folder).mkdir(parents=True, exist_ok=True)
```
vkedia, line 419: if I understand correctly, a single completion file is shared among all the processes of this component? In that case if one of them has created the completion file, it is possible that all others will end up doing nothing due to the if statement below?

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -429,15 +516,32 @@ def initiate_run():
...
-            preprocess_output, pytorch_model_folder, mlflow_model_folder)
+
+    if is_main_process():
+        cleanup(completion_files_folder, model_selector_output,
+                preprocess_output, pytorch_model_folder, mlflow_model_folder)
+
+
+def parse_to_int(s):
+    """
+    To parse string to integer with default value in case of failure as one.
```
vkedia, line 527: why do you want to default to 1? Isn't it better to fail fast if node counts or gpus is not set or set to an invalid value.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -429,15 +516,32 @@ def initiate_run():
...
+        return 1
 
 
 def _initiate_run(completion_files_folder: str, model_selector_output: str,
                   preprocess_output: str, pytorch_model_folder: str, mlflow_model_folder: str):
     """Run the model selector, preprocess, finetune and registration script."""
     # get task name
     task_name = decode_param_from_env_var("task_name")
+    nodes = parse_to_int(decode_param_from_env_var("Node_Count"))
+    gpus = parse_to_int(decode_param_from_env_var("number_of_gpu_to_use_finetuning"))
```
vkedia, line 543: who sets this number of gpus environment variable? And is it guaranteed that singularity will definitely start these many processes? Could it happen that it starts less than these in which case the barrier would never be released.

vkedia, line 543: What happens in case a worker, which was waiting on the barrier, gets preempted and started again. In that case it might again write to the barrier file thus releasing all the waiting workers (who were waiting for main to finish its work).

vkedia (changes requested):

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > who sets this number of gpus environment variable? And is it guaranteed that singularity will definitely start these many processes? Could it happen that it starts less than these in which case the barrier would never be released.

Process count will be same as GPU count. This is tested . It's behavior would not change dynamically for every run. yes , barrier need to know how many process to wait , even in case of torch dist. this will be know internally based on group size

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > is this seek needed? Won't a freshly opened file already be at offset 0.

At some point I was facing an issue where rank 0 was not reading contents property after adding seek(0) , it got fixed ; I feel it is good to specify explicitly

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > if any of the file i/o code throws any error will that get logged somewhere?

yes , while initial development phase of function , I was able to check exceptions in log. A folder called "user logs" will get created which will hold the std error and std out

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > I am not able to understand why is this needed? It will anyways wait at the barrier below

In few sets things are async , if we don't wait for completion file work might be in a background thread before completion of actual work process will reach barrier and cause issues based on incompletion. Hence waiting for completion is required. And for all ranks(process) to come out of this completion wait loop barrier is required.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > if I understand correctly, a single completion file is shared among all the processes of this component? In that case if one of them has created the completion file, it is possible that all others will end up doing nothing due to the if statement below?

yes, in this case , it is for creation of completion folder path. In case path already existing the Path library is hopefully not costly because of exist_ok argument ; If we don't use this way we need to either check if rank is 0 and get it created by it then if rank1 is scheduled first and it checks path it might get to error, we'll need to solve that scenario. Or we need an if condition to check for existence and create even in this case the present expression is locally of same cost.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
```diff
@@ -367,32 +369,117 @@ def add_task_specific_params(cmd: List[str], task_name: str, component_name: str
...
+    with open(barrier_file, 'a+') as f:
+        portalocker.lock(f, portalocker.LOCK_EX)
+        try:
+            f.seek(0)
+            lines = f.readlines()
+            process_count = len(lines)
+            if process_count < num_processes:
+                f.write(f"{os.getpid()} reached the barrier\n")
+                f.flush()
+                portalocker.unlock(f)
```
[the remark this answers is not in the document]
yeshsurya: > you don't need this unlock since you are unlocking in finally

Two reasons : 1.  I want to write info log after releasing the lock 2. I want to log within try block hence I kept it even though it looks redundant.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > here you can open as a readonly file instead of a+

Sure , I'll change this

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: sure , looks good to do this ; However it seem to not cause issue without it as well. But I'll make the change

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > you already created this folder above so you can skip creating here.

removing this , Thanks !

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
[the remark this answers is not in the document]
vkedia: ok and calling unlock twice works?

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
vkedia: My question was specifically around completion file itself (and not the folder). It seems a single completion file is shared among all the workers. In that case it is possible that one of the workers succesfully completes its work, writes out the completion file before the 2nd worker has started. When the 2nd worker starts, it will see that the completion file exists and not do anything.

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
vkedia: Why is this resolved?

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
vkedia: By preemption I meant eviction of the worker. Workers can be evicted and restarted. In which case they will everything from scratch including writing to the barrier.

vkedia (changes requested):

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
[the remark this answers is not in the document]
yeshsurya: > ok and calling unlock twice works?

yes !

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > My question was specifically around completion file itself (and not the folder). It seems a single completion file is shared among all the workers. In that case it is possible that one of the workers succesfully completes its work, writes out the completion file before the 2nd worker has started. When the 2nd worker starts, it will see that the completion file exists and not do anything.

I didn't see such thing happening but this check seems to be present in older code

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: sorry , I missed to reply this . Reason I thought was lowest possible value is 1 , so keep that as default in case of parsing failure

assets/training/finetune_acft_hf_nlp/environments/acpt/context/finetune_run.py
yeshsurya: > By preemption I meant eviction of the worker. Workers can be evicted and restarted. In which case they will everything from scratch including writing to the barrier.

I don't think this would happen

vkedia (approved):
Let us follow up with moving the barrier logic to its own package and writing unit tests for it.

amangupta26 (approved):

PR #3310, opened 2024-08-28
Change the logger message for get_credential.

scravikiran:


novaturient95 (approved):

assets/training/distillation/src/common/utils.py
```diff
@@ -77,14 +77,14 @@ def get_credential() -> Union[ManagedIdentityCredential, AzureMLOnBehalfOfCreden
...
-        logger.error("MSI auth failed")
+        logger.warning("MSI auth failed")
     try:
         credential = AzureMLOnBehalfOfCredential()
         credential.get_token("https://management.azure.com/.default")
         logger.info("Using OBO creds")
         return credential
     except Exception:
-        logger.error("OBO cred failed")
+        logger.warning("OBO cred failed")
```
skasturi, line 87: How about collating various auth options tried & log warning, if one of them is successful & if not, logging an error?

ShobithNandakumar: Agree with @skasturi - moreover, I don't think these need to be logged as warnings as well. 

I have a follow-up PR for this, can you please see if it makes sense to you - https://github.com/Azure/azureml-assets/pull/3314

PR #3311, opened 2024-08-29
ACFT MaaS Data Import Validations

skanakamedal:


Anubha98 (approved):
we can remove import script from acpt now as data import env is separately getting used from 0.0.21 release

PR #3314, opened 2024-08-29
chore: refactor credential method

ShobithNandakumar:
### Description
Minor refactor to try various auth methods while retrieving ML client.


assets/training/distillation/src/common/utils.py
```diff
@@ -68,32 +70,39 @@ def newfn(*args, **kwargs):
...
         return credential
-    except Exception:
-        logger.error("Azure CLI cred failed")
+    except Exception as e:
+        logger.info("Failed to get AzureCliCredential.")
+        auth_failures += f"AzureCli: {str(e)}. \n"
 
-    raise Exception("Error creating credentials.")
+    auth_failures = "One or more auth methods failed: \n" + auth_failures
+    logger.error(auth_failures)
```
skasturi, line 104: Could you please check if `auth_failures` contain any sensitive information?

ShobithNandakumar: Thanks for the call out @skasturi. They do not contain sensitive information. Individual methods only raise generic exceptions as listed below:

MSI - No managed identity endpoint found.
OBO - AzureML On Behalf of credentials not available in this environment.
AzureCLI - the credential invoked the Azure CLI but didn't receive an access token.

PR #3339, opened 2024-09-05
remove python 3.8

ycheng35xo:


alanpo1 (approved):

yunjie-hub (approved):

assets/model_monitoring/components/model_monitor/model_monitor_feature_selector/spec.yaml
```diff
@@ -46,7 +46,6 @@ conf:
     channels:
       - defaults
     dependencies:
-      - python=3.8
```
vizhur, line 49: why did you remove python version from here? we recommend to specify the interpreter version across AzureML. EMS reject any environment to be registered with no python version specified

vizhur (changes requested):

vizhur (approved):

PR #3353, opened 2024-09-09
Update batch score llm components in benchmark components

babu-namburi:
Benchmarking Pipeline failing with below error
`Unable to get image details : Specified image azureml/curated/model-evaluation @ sha256:b1d6981fe502c9abe2bed365bb6d2aa0ce0bb63cb9c42b57a83fd1d84a90c6a4 reached EOL 08/31/2024 00:00:00. Please use a newer image version or a different image. Please refer to aka.ms/azureml/environment/ce-brownouts..

This PR updates the component to the latest version from the correct registry.


guptha23 (approved):

babu-namburi:
@guptha23 
Info from Rahul:
```The problem is the fact that registry azureml-batch-score is not prod, it is an internal registry. We shouldn't publish components to prod registry that uses sub-components from a non-prod registry.
 
Ayush is working on releasing updated batch_score to azureml registry, we could use that. For the time being, please register this component in an internal registry, and use it for benchmarking runs.

Will wait till the new component is available.

FYI, @ShipraJain01

PR #3360, opened 2024-09-10
fix the HistogramBuilder input order bug

ycheng35xo:


imatiach-msft (dismissed):

assets/model_monitoring/components/src/model_monitor_output_metrics/entities/signals/data_drift_signal.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         self.feature_metrics: List[FeatureMetrics] = self._build_feature_metrics(
             monitor_name, signal_name, metrics, feature_importance
         )
-        self.histogram_builder = HistogramBuilder(baseline_histogram, target_histogram)
+        self.histogram_builder = HistogramBuilder(target_histogram, baseline_histogram)
```
RichardLi1437, line 40: we have such a bug in the code for quite a long time? So the datadrift we build before are all wrong?

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
```diff
@@ -4,7 +4,7 @@ type: pipeline
 name: data_drift_signal_monitor
 display_name: Data Drift - Signal Monitor
 description: Computes the data drift between a baseline and production data assets.
-version: 0.3.46
+version: 0.3.47
```
RichardLi1437, line 7: shouldn't weekly DRI to upgrade the versions?

RichardLi1437 (approved):

assets/model_monitoring/components/src/model_monitor_output_metrics/entities/signals/data_drift_signal.py
ycheng35xo: yes.

assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
ycheng35xo: We want to do a hotfix because there is a Css ticket. I do not think there will be more PRs this week so we can skip weekly release

PR #3378, opened 2024-09-16
Update model evaluation environment to remove dependency on transformers patch version.

guptha23:
Update model evaluation environment to remove dependency on transformers patch version.


assets/training/model_evaluation/environment/context/conda.yaml
```diff
@@ -36,7 +36,6 @@ dependencies:
   - pydantic=={{latest-pypi-version}}
   - mteb=={{latest-pypi-version}}
   - setuptools>=70.0.0
-  - rouge>=1.0.0,<2.0.0
```
guptha23, line 39: we are removing rouge as it is added as a dependency in latest version of azureml-metrics package.

ankushbhatia2 (approved):

ankushbhatia2 (approved):

PR #3389, opened 2024-09-18
feat: add designer-base image

RealKai42:
Due to the depreciation of the Curated Environment, we are trying to move some image to system registry.

This is the first pr, we’re working on getting the full process up and running, so any suggestions would be greatly appreciated.


assets/designer/environments/aml-designer/spec.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+name: "{{asset.name}}"
+version: "{{asset.version}}"
+
+build:
+  path: "{{asset.repo.url}}#{{asset.repo.commit_hash}}:{{asset.repo.build_context.path}}"
+  dockerfile_path: "{{image.dockerfile.path}}"
+
+os_type: linux
+
```
vizhur, line 14: python 38 is reaching eol in October, I would recommend to upgrade

assets/designer/environments/aml-designer/context/conda_dependencies.yaml
```diff
@@ -0,0 +1,31 @@
+name: aml-designer
+channels:
+- conda-forge
+dependencies:
+- python=3.8.10
```
vizhur, line 5: please upgrade python to at least 39

assets/designer/environments/aml-designer/context/conda_dependencies.yaml
```diff
@@ -0,0 +1,31 @@
...
+- pip=20.2
```
vizhur, line 6: please use a more recent pip resolver version

assets/designer/environments/aml-designer/asset.yaml
```diff
@@ -0,0 +1,6 @@
+name: aml-designer
```
vizhur: please don't use aml, AzureML, microsoft, etc words in the asset name. you can find naming conventions in wiki

RealKai42: Thanks, I just change it to designer-base

assets/designer/environments/aml-designer/context/conda_dependencies.yaml
RealKai42: The `azureml-designer-classic-modules` depend on the `pandas ==1.0.5` which is not support python 3.9 or higher .

I know the 3.8 will end of the support, maybe we can do this upgrade after the migration to system registry.

PR #3394, opened 2024-09-18
Update torch version and dependencies in AutoML NLP CE

raviskolli:
PyTorch updated to 2.2.2


jeff-shepherd (commented):
The pin-to-latest-mcr-version-tag sytax is for the old curated environments.
For the registry environments, you need to use latest-image-tag

The same applies to pin-to-latest-pypi-version, which should be latest-pypi-version

jeff-shepherd (commented):
This is downgrading aiohttp.
aiohttp==3.9.4 should be aiohttp==3.10.2

jeff-shepherd (approved):

PR #3427, opened 2024-09-25
refactor phi 3.5 moe model card to fit catalog template

jfomhover:
- align phi 3.5 moe model card with existing phi3 models surfaced in GH Models


jfomhover:
Waiting to see if I can turn freeplaygroud=true immediately

martincai:
Looks good. Since this is just splitting the model card into a few tabs, it's good to go.

PR #3434, opened 2024-09-26
Moved ALLaM-2-7b-instruct spec from azureml-asset

utkshukla:
Moved ALLaM-2-7b-instruct spec from azureml-asset


utkshukla:
> Lets add to registry files otherwise LGTM. 1 PR check is failing as well

As discussed.

lykelly19 (approved):

PR #3457, opened 2024-10-06
chore: introduce timeout to http calls

ShobithNandakumar:
### Description
Bad endpoints could lead to the request remaining in a hung state. This timeout should fix that.


sharvin2187 (approved):
LGTM

PR #3503, opened 2024-10-17
Update model cards + versions for HLS models

cassieesvelt:


tinaem (approved):
added a commit for medimageinsight - good to go!

PR #3537, opened 2024-10-30
Added tensorflow-2.16-cuda12

jeff-shepherd:


assets/training/general/environments/tensorflow-2.16-cuda12/context/Dockerfile
```diff
@@ -0,0 +1,257 @@
...
+    gnupg \
+    gnupg2 \
+    gpg \
+    libdpkg-perl \
+    dpkg \
+    libpcre3 \
+    libpcre2-8-0 \
+    sqlite3 \
+    uidmap \
+    libkrb5-26-heimdal \
```
vizhur, line 121: do you need all those? or just copied from the base images? nvcr is maintaining images well you don't need all of those, except tar,curl,wget,gcc

jeff-shepherd: They are just copied from a similar ubuntu 20.04 base image.
If they don't cause a problem, I would like to keep them.
I see this current definition as temporary until debug can support 22.04 images.
When that happens, we can use a much simpler definition.

vizhur (approved):

PR #3541, opened 2024-10-30
Add evaluationresult tags for cost, quality index and perf assets

guptha23:
Add evaluationresult tags for cost, quality index and perf assets


saikmicro (dismissed):

ankushbhatia2 (approved):

vizhur (changes requested):
you need to bump a version of the package for this to take effect

guptha23:
> you need to bump a version of the package for this to take effect

updated the change log and setup files as suggested to bump up the version.

vizhur (approved):

PR #3553, opened 2024-11-01
feat: rename designer-pytorch related image

RealKai42:
Rename 
`designer-pytorch-2.3-train` to `designer-pytorch-train`
`designer-pytorch-2.3` to `designer-pytorch`


vizhur (changes requested):
We are not bypassing the check or changing the naming conventions. We can look into that if that is an absolute blocker, if not, please proceed with a new images on MCR

PR #3583, opened 2024-11-08
Update owner for evaluationresult assets 

guptha23:
Update owner for evaluationresult assets (to remove aml-llm)


guptha23:
> Lets move eval results to ADO instead

sure, we will try to move the results to ADO as suggested. thank you!!

PR #3584, opened 2024-11-09
add score logic extraction from evaluation sdk

saikmicro:


assets/evaluation_on_cloud/environments/evaluations-built-in/context/online_eval/evaluate.py
```diff
@@ -176,6 +173,43 @@ def run_evaluation(command_line_args, evaluators, evaluator_configs):
         mlflow.log_artifact("instance_results.jsonl")
 
 
+def extract_score(data: Dict[str, Any]) -> Tuple[List[str], float]:
+    """Extract the float score value from the evaluation result."""
+    # Step 1: If the count of keys is 1, return that value
+    if len(data) == 1:
+        return list(data.keys()), list(data.values())[0]
```
ShipraJain01, line 180: Do we need to check for the return type? what if data length is 1 but the return value int double / numeric?

saikmicro: makes sense, made numeric value compulsory.

ShipraJain01 (approved):

vizhur (approved):

PR #3620, opened 2024-11-26
Specify langchain community version in rag images

clomeli-ms:
Specify langchain community version in rag images


shpengmsft:
I don't think we need this PR. because langchain-community will be installed via "azureml-rag[... langchain,...]"

PR #3632, opened 2024-12-01
Fix IndentationError

yunjie-hub:


chrjia (approved):

assets/model_monitoring/components/data_drift/data_drift_compute_metrics/spec.yaml
```diff
@@ -65,6 +65,7 @@ conf:
       - pip:
         - azure-storage-file-datalake~=12.8.0
         - azure-ai-ml~=1.13.0
+        - azureml-mlflow~=1.49.0
```
RichardLi1437, line 68: momo can't work with latest protobuf, so pin mlflow, thus pin the dependent protobuf to its old version, is that the fix?

assets/model_monitoring/components/feature_attribution_drift/feature_attribution_drift_compute_metrics/spec.yaml
```diff
@@ -2,7 +2,7 @@ $schema: http://azureml/sdk-2-0/SparkComponent.json
 type: spark
 
 name: feature_attribution_drift_compute_metrics
-version: 0.3.22
+version: 0.3.23
```
RichardLi1437, line 5: seems some spec.yaml files like this one doesn't need to upgrade, I suppose you use upgrade --all so all spec files are upgraded, but that's ok.

RichardLi1437 (approved):

assets/model_monitoring/components/feature_attribution_drift/feature_attribution_drift_compute_metrics/spec.yaml
yunjie-hub: Yes, I upgrade all

assets/model_monitoring/components/data_drift/data_drift_compute_metrics/spec.yaml
yunjie-hub: Actually I didn't figure out the protobuf version, from the call stack it's caused by mlflow. I tested in spark session seems I can fix it with this pin.

PR #3748, opened 2025-01-15
MatterSim model card update

abhishekMS2024:


roopavidhya (approved):

itzasifmd (approved):
Updated model card as we received from PM

PR #3763, opened 2025-01-16
Updated environments name

Tarun-Chevula:
Updated environments name


assets/inference/environments/mlflow-py312-inference/tests/requirements.txt
```diff
@@ -1,3 +1,4 @@
 azure-ai-ml==1.2.0
```
vizhur, line 1: don't you want to bump azure-ai-ml instead?

Tarun-Chevula: azure-ai-ml latest version doesn't support py3.13 yet. We will do this once it is supported.

vizhur (approved):

PR #3774, opened 2025-01-20
Add component upgrade script

iamrk04:
This PR introduces a script to auto upgrade components for release.

- Following features are supported in the script:
  - upgrade the component's version 
  - upgrade the component's environment version
  - upgrade the pipeline component's job version
  - prints the regular expression required by the asset build pipeline
  - uses caching to avoid duplicate search for assets
- A readme has been added to help folks run the script
- Easy onboarding of new components for auto upgrade by just adding the name in `components.yaml` file

### DIFF
--- a/assets/training/scripts/_component_upgrade/components.yaml
+++ b/assets/training/scripts/_component_upgrade/components.yaml
@@ -1,0 +1,72 @@
+# This file contains the list of component's name from spec.yaml that needs to be auto upgraded.
+
+# Instructions:
+# - Add the component name from the spec.yaml file that needs to be auto upgraded.
+# - pipeline component must be mentioned after all of its sub-components have been mentioned.
+
+component:
+  # Forecasting
+  - automl_hts_inference_setup_step
+  - automl_hts_prs_inference_step
+  - automl_hts_inference_collect_step
+  - automl_hts_inference
+  - automl_hts_training_setup_step
+  - automl_hts_automl_training_step
+  - automl_hts_data_aggregation_step
+  - automl_hts_training_collect_step
+  - automl_hts_training
+  - automl_many_models_inference_setup_step
+  - automl_many_models_inference_step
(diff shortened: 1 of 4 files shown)


assets/training/scripts/_component_upgrade/main.py
```diff
@@ -0,0 +1,245 @@
...
+            with open(asset_path, "r") as file:
+                spec = safe_load(file)
+            if spec.get("name", None) not in OWNED_COMPONENT_NAMES:
+                continue
+            component_paths.append(asset_path)
+    return component_paths
+
+
+def _get_bumped_version(version: str, increment: bool = True) -> str:
+    """
```
babu-namburi, line 54: This is for my own understanding, 
I have seen version to be string in general and have seen something like **0.0.1.dev**, won't this happen in prod registry 
 ? 
Should we look into a more robust version bump tools like below 
 https://pypi.org/project/bump2version/

iamrk04: None of the components that we own have this type of behavior in prod registry. Nevertheless, thanks for pointing this out. We can definitely consider this in the future if such a case happens.

assets/training/scripts/_component_upgrade/components.yaml
yeshsurya, line 1: Given there would be no component which is not used in pipeline we can have logic to get the pipeline and which need to scanned for components and look for update. To reduce future updated on this file. If this refactor is time consuming it can be backlogged to future enhancement.

yeshsurya (approved):

assets/training/scripts/_component_upgrade/components.yaml
iamrk04: We can only release components that we own and hence we have this file. If we, as a team, develop any new components(s) or start owning components not listed in this file, we simply need to add them here and that's it.

PR #3785, opened 2025-01-24
Spec for Phi-4-Mini and Phi-4-Omni

utkshukla:
Spec for Phi-4-Mini and Phi-4-Omni


anupamawalaus:
Safe to Publish. Approved.

PR #3793, opened 2025-01-29
FMI 64 release with 0.6.3.post1 support

utkshukla:
FMI 64 release with 0.6.3.post1 support.

Changes to support Phi-4 model. We have build the vLLM whl to support custom changes and associated it with a storage account, which can be added as a pip installer in PR.


pabhatia-ms (dismissed):

vizhur (changes requested):
any reason to install from a blob?

utkshukla:
> any reason to install from a blob?

@vizhur for the Phi-4 models we have custom vLLM code which is not yet published in public. We have built the code and installing it using blob as it was done for Phi-3.5-vision-instruct.

vizhur (approved):

pabhatia-ms (approved):

PR #3795, opened 2025-01-29
Removed Ubuntu 20.04 inference environments

Tarun-Chevula:
Removed Ubuntu 20.04 inference environments


vizhur (changes requested):
archive environments first

vizhur (approved):

PR #3798, opened 2025-01-30
Disable maap flag for ALLaM-2-7b-instruct for Dev and Staging

utkshukla:
Disable maap flag for ALLaM-2-7b-instruct for Dev and Staging


vizhur:
here are the changes along with the workflow updates https://github.com/Azure/azureml-assets/pull/3799

PR #3823, opened 2025-02-05
Updated Financial reports model card with inference examples

mgtk77:
Added some inference examples, for both chat completion and score endpoints.


vizhur (changes requested):
please make a PR from branch

PR #3826, opened 2025-02-07
update eci asset to v3

MilesHolland:
Updates the reference for the ECI evaluator to point to version 3 of the model. Version 2 is skipped because that was a faulty test version.


MilesHolland:
Added safe to publish label - this modal has already been tested in an internal testing registry.

lykelly19 (approved):

qusongms (approved):

PR #3846, opened 2025-02-19
Update account_uri when storage_name is updated for AzureBlobstoreAss…

lykelly19:
…etPath

### DIFF
--- a/scripts/azureml-assets/azureml/assets/config.py
+++ b/scripts/azureml-assets/azureml/assets/config.py
@@ -531,7 +531,6 @@ class AzureBlobstoreAssetPath(AssetPath):
             self._cloud_suffix = AzureBlobstoreAssetPath.AZURE_CLOUD_SUFFIX
         else:
             self._cloud_suffix = _get_storage_endpoint_from_metadata()
-        self._account_uri = f"https://{storage_name}.blob.{self._cloud_suffix}"
 
         # Its possible that the account URL may need additional tweaking to add a SAS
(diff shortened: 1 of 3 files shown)


scripts/azureml-assets/azureml/assets/config.py
```diff
@@ -528,10 +528,10 @@ def __init__(self, storage_name: str, container_name: str, container_path: str):
         if _get_default_cloud_name() in [AzureEnvironments.ENV_DEFAULT,
                                          AzureEnvironments.ENV_US_GOVERNMENT,
                                          AzureEnvironments.ENV_CHINA]:
-            cloud_suffix = AzureBlobstoreAssetPath.AZURE_CLOUD_SUFFIX
+            self._cloud_suffix = AzureBlobstoreAssetPath.AZURE_CLOUD_SUFFIX
         else:
-            cloud_suffix = _get_storage_endpoint_from_metadata()
-        self._account_uri = f"https://{storage_name}.blob.{cloud_suffix}"
+            self._cloud_suffix = _get_storage_endpoint_from_metadata()
+        self._account_uri = f"https://{storage_name}.blob.{self._cloud_suffix}"
```
elesel, line 534: Because self._account_uri is now being set in multiple places, what would you think about making an _account_uri (private) property that returns `f"https://{self._storage_name}.blob.{self._cloud_suffix}` ?

lykelly19: updated!

elesel (approved):

PR #3850, opened 2025-02-20
Support H100 GPU

matthiasblondeel:
The Health And Life Sciences models support the new H100 GPU


vivram (dismissed):

vivram (commented):
inference-recommended-sku in all these files too should be updated with these new skus.

Validation is failing with this error: skus in compute allowlists and not in recommended sku => {'Standard_NC80adis_H100_v5', 'Standard_NCC40ads_H100_v5', 'Standard_NC40ads_H100_v5'}

lykelly19 (approved):

PR #3863, opened 2025-02-25
Add default packages to mlflow environment

QianqianNie:


assets/inference/environments/mlflow-py312-inference/context/Dockerfile
```diff
@@ -21,6 +21,8 @@ COPY mlflow_hf_score_cpu.py /var/mlflow_resources/mlflow_hf_score_cpu.py
 COPY mlflow_hf_score_gpu.py /var/mlflow_resources/mlflow_hf_score_gpu.py
 
 # Create conda environment
+# Add default packages
+RUN conda config --add create_default_packages azureml-inference-server-http --add create_default_packages azureml-ai-monitoring --add create_default_packages azureml-contrib-services
```
vizhur, line 25: This is so wrong...

1. The intent of curated environment images is to use them as is. You should not expect anyone building on top of it. Customers do that, but that is their problem. For those scenarios we have base images. So, if you need those packages, pin them properly and move them to conda spec.
2. Contrib-services is a joke no a package. Whatever you need to do you can ship with inference server. Contrib packages **by design** are not allowed in any prod scenarios. Saying that as the author of contrib packages concept in AzureML
3. azureml-ai-monitoring besides the name of the package, there are many questions. Like where is the source. Why it is not maintainer. At all. you are installing it in py312 but classifiers say it only supports up to py310. This "Create environment with base image `mcr.microsoft.com/azureml/openmpi4.1.0-ubuntu20.04` and conda dependencies, then build the environment.\n\n```\nchannels:\n  - conda-forge\ndependencies:\n  - python=3.8\n  - pip=22.3.1\n  - pip:\n      - azureml-defaults==1.38.0\n      - azureml-ai-monitoring\nname: model-env" in the description of the package says hard no to any reference to this package. from deprecated base image, eol interpreter version, pip that is vulnerable, to 3+ years old version of azureml-defaults that brings a ton outdated and incompatible packages.


contrib-services is a joke package. None of the prod scenarios should depend on contrib, by design.

vizhur (changes requested):

PR #3868, opened 2025-02-25
Security patch 2025.02 for ai-studio-dev

YusakuNo1:


assets/aistudio/environments/ai_studio_dev/context/requirements.txt
```diff
@@ -11,3 +11,5 @@ langchain-openai
 semantic-kernel
 pytest
 keyring
+# Temporary fix version for security patch
```
vizhur, line 14: azure-ai-ml
# hardcoded the version of azureml-mlflow here for faster Docker image building speed
azureml-mlflow==1.57.0

Both v2 sdk and mlflow should use template or at least pinned. The latter is true for the rest of dependencies

why do you need to lowerbound cryptography?

vizhur (approved):

assets/aistudio/environments/ai_studio_dev/context/requirements.txt
```diff
@@ -11,3 +11,6 @@ langchain-openai
 semantic-kernel
 pytest
 keyring
+# dev dependencies
+setuptools>70.0.0
```
YusakuNo1, line 15: For some reasons, if we don't pin min version, it'll default to 69.x.x, which cause security issue

vizhur (approved):

PR #3897, opened 2025-03-03
fix: designer-related vulnerabilities

RealKai42:


assets/designer/environments/designer/context/conda_dependencies.yaml
```diff
@@ -9,5 +9,7 @@ dependencies:
   - scipy==1.10.0
   - pip:
     - azureml-designer-classic-modules==0.0.191
+    - cryptography==44.0.1
```
vizhur, line 12: there is no reason to have cryptography in the requirements

vizhur (approved):

PR #3922, opened 2025-03-07
Force rebuild for `tensorflow` env

iamrk04:


jeff-shepherd (dismissed):

assets/training/general/environments/tensorflow-2.16-cuda12/context/Dockerfile
```diff
@@ -255,3 +255,5 @@ RUN HOROVOD_WITH_TENSORFLOW=1 pip install horovod[tensorflow]
 ENV LD_LIBRARY_PATH=$CONDA_PREFIX/lib:$LD_LIBRARY_PATH
 
 ENV TF_USE_LEGACY_KERAS=1
+
```
vizhur, line 258: that's not going to merge

iamrk04: Hi @vizhur,

May I know the right way to go about it? We want to fix vulnerabilities, since we haven't pinned the vulnerable packages, just releasing a new version will fix the issue. But the build pipeline doesn't pick this image since there isn't any change in the artifacts for this env.

Thanks in advance!

vizhur: Use templates for your dependencies, but don't commit such comments to a prod environment

PR #3976, opened 2025-03-24
Vaibhj/deprecatespark33

vaibhj:


assets/model_monitoring/components/data_drift/data_drift_signal_monitor/spec.yaml
```diff
@@ -77,7 +77,7 @@ jobs:
         type: mltable
     resources:
       instance_type: ${{parent.inputs.instance_type}}
-      runtime_version: "3.3"
+      runtime_version: "3.4"
```
RichardLi1437, line 80: a suggestion, you may want to define a version const and use that const so that next time we upgrade to 3.5 we don't need to update everywhere, just need to update one single const.

RichardLi1437 (approved):

ycheng35xo (approved):

PR #4000, opened 2025-03-28
Openai-whisper-large-v3 and MedImageInsight release with mlflow-model-inference image

utkshukla:
Openai-whisper-large-v3 and MedImageInsight release with mlflow-model-inference image


pabhatia-ms (approved):
LGTM

PR #4004, opened 2025-03-28
Increment version for updated models

kumra-ms:
This PR changes the target file and bump model version for the updated models.


kumra-ms:
Opened the PR for testing

PR #4006, opened 2025-03-28
Increment version for updated models

Bishal-ms:
This PR changes the target file and bump model version for the updated models.


kumra-ms (approved):
Lgtm

PR #4091, opened 2025-04-14
Create MLmodel for MedImageParse3D

raghu-microsoft:
Added MLmodel file

### DIFF
--- a/assets/models/system/MedImageParse3D/MLmodel
+++ b/assets/models/system/MedImageParse3D/MLmodel
@@ -1,0 +1,26 @@
+flavors:
+  python_function:
+    artifacts:
+      model_dir:
+        path: artifacts/checkpoints
+        uri: /__w/1/s/checkpoints
+    cloudpickle_version: 3.1.1
+    code: code
+    env:
+      conda: conda.yaml
+      virtualenv: python_env.yaml
+    loader_module: mlflow.pyfunc.model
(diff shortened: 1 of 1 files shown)


assets/models/system/MedImageParse3D/MLmodel
```diff
@@ -0,0 +1,26 @@
+flavors:
+  python_function:
+    artifacts:
```
pabhatia-ms, line 3: we need to bump up the version as well in spec.yaml

pabhatia-ms (changes requested):
bump up the version

assets/models/system/MedImageParse3D/MLmodel
pabhatia-ms: so, you want to following steps.
1. add this MLModel file here.
2. run python script which will add the image_path for new inference image and also bump up the version ?
Is that correct ?
If yes, then you don't need the line number 17.. Seems you wld have added it manually by taking this MLModel from blob.

pabhatia-ms: If you want to run python script just to bump up the version, you can just have a spec.yaml with bump up version.
cc: @kumra-ms

pabhatia-ms (dismissed):

pabhatia-ms (approved):

PR #4112, opened 2025-04-21
Added Base MLFlow Image in Metadata for Import Model Component

Priyanka1329-coder:


assets/training/model_management/src/azureml/model/mgmt/processors/pyfunc/convertors.py
raghu-microsoft, line 1: @Priyanka, Why is only the mlflow-model-inference image information included here? Why not the foundation-model-inference as well?

PR #4134, opened 2025-04-30
Raghu/baseimage update

raghu-microsoft:
image
Tested locally It is working,


assets/training/model_management/components/convert_model_to_mlflow/spec.yaml
```diff
@@ -13,6 +13,7 @@ environment: azureml://registries/azureml/environments/model-management/versions
 
 code: ../../src/
 command: |
+  pip install transformers==4.48.0 torch==2.2.2 numpy==1.23.5 pandas==1.5.3 urllib3==1.26.19 --no-cache-dir --force-reinstall
```
vizhur, line 16: runtime dependencies installation? sweet...

raghu-microsoft: Thanks! Yes, the issue stemmed from a version mismatch between torch and transformers, specifically azureml-evaluate-mlflow library in the "import model pipeline"
To ensure compatibility, I locked the versions explicitly in the spec.yaml using --force-reinstall and --no-cache-dir to avoid conflicts from cached builds.

If preferred, I can move these dependencies into a requirements.txt file for cleaner separation and maintainability. Let me know what you’d suggest.

raghu-microsoft: I incorporated the required libraries directly into the Dockerfile rather than using the spec.yaml file. 
After adding those changes, I ran `import_model`, it is now running successfully.

raghu-microsoft: @vizhur, Could you please review this PR

vizhur (approved):

PR #4162, opened 2025-05-09
Create new local model versions for models with perf issues

sophies927:
Summary of changes:

1. Publish new models with improved perf (`deepseek-r1-distill-llama-8b`, `deepseek-r1-distill-qwen-1.5b`, `deepseek-r1-distill-qwen-14b`, `deepseek-r1-distill-qwen-7b`, `phi-4-mini-instruct`)
2. Fix issue with Qwen2.5-coder-1.5b generic GPU variant naming (missing `instruct`)
3. Fix issue with Phi-4-reasoning parent assetId (wrong version)

### DIFF
--- a/assets/models/system/deepseek-r1-distill-llama-8b/model.yaml
+++ b/assets/models/system/deepseek-r1-distill-llama-8b/model.yaml
@@ -1,6 +1,6 @@
 path:
   container_name: models
-  container_path: foundrylocal/foundry-local/deepseek-r1-distill-llama-8b
+  container_path: foundrylocal/fl-perf-improvements/deepseek-r1-distill-llama-8b
   storage_name: automlcesdkdataresources
   type: azureblob
 publish:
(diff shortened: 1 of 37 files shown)


assets/models/system/deepseek-r1-distill-llama-8b/model.yaml
```diff
@@ -5,4 +5,4 @@ path:
   type: azureblob
 publish:
   description: description.md
-  type: custom_model
\ No newline at end of file
+  type: custom_model
```
sophies927, line 8: This appears to indicate a change even though there isn't one. Could this create issues when publishing?

assets/models/system/deepseek-r1-distill-qwen-14b/model.yaml
```diff
@@ -5,4 +5,4 @@ path:
   type: azureblob
 publish:
   description: description.md
-  type: custom_model
\ No newline at end of file
+  type: custom_model
```
sophies927, line 8: Same thing as above - appears to indicate a change where there isn't one (that's visible at least).

vizhur (approved):

PR #4207, opened 2025-05-22
Detect if MLFlow Model was updated

lykelly19:


vizhur (approved):

sophies927 (commented):
This looks good to me from a Foundry Local perspective! Just to confirm that I'm fully understanding the changes, this will only cause an error if MLFlow model is updated, but it looks like nothing will happen if a custom model variant with an MLFlow parent model (e.g., some of the Phi models) is updated, correct? And are there any plans to fully delete/remove MLFlow models entirely?

If no to both, local models will be unaffected since we only update/publish new versions for model variants.

vizhur (approved):

PR #4208, opened 2025-05-22
Andscho/py3 9 fix

andscho-msft:


scripts/azureml-assets/azureml/assets/model/download_utils.py
```diff
@@ -94,7 +94,7 @@ def _onerror(func, path, exc_info):
         raise
 
 
-def run_cmd(cmd, cwd: Path | None = None, env: dict | None = None) -> int:
+def run_cmd(cmd, cwd: Path = None, env: dict = None) -> int:
```
nmecklenburg, line 97: nit: i mean, if we want to keep the spirit of the original change, we can always import Optional from typing and do `Optional[Path] = None` and the like, right?

vizhur (approved):

PR #4209, opened 2025-05-23
Update Dockerfile

raghu-microsoft:
Completely removed preinstalled Torch version (especially the Conda-installed torch==2.5.1) and cleanly installed torch==2.6.0 via pip inside a Docker image,

image

### DIFF
--- a/assets/training/model_management/environments/model-management/context/Dockerfile
+++ b/assets/training/model_management/environments/model-management/context/Dockerfile
@@ -25,7 +25,27 @@ RUN conda install python=3.9 -y
 RUN pip install -r requirements.txt --no-cache-dir
 
 # Vulnerability fix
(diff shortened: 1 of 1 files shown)


assets/training/model_management/environments/model-management/context/Dockerfile
```diff
@@ -25,7 +25,27 @@ RUN conda install python=3.9 -y
 RUN pip install -r requirements.txt --no-cache-dir
 
 # Vulnerability fix
-RUN pip install torch==2.6.0 torchvision --no-cache-dir
+# Remove Conda-installed Torch and related packages
+RUN conda remove -y --force \
```
vizhur, line 29: can you use torch26 image from aifx instead? it defeats the purpose of using that base image

PR #4235, opened 2025-06-03
Environment update for the delete_endpoint component

abhishekMS2024:


assets/common/components/delete_endpoint/spec.yaml
```diff
@@ -9,7 +9,7 @@ display_name: Delete Endpoint
 description:
   Deletes an endpoint resource.
 
-environment: azureml://registries/azureml/environments/python-sdk-v2/versions/31
+environment: azureml://registries/azureml/environments/model-management/versions/48
```
kumra-ms, line 12: Why did not we use previous version of python-sdk-v2 environment itself?

pabhatia-ms (approved):

PR #4264, opened 2025-06-13
Update FT and Eval components

novaturient95:


yeshsurya (changes requested):
Please add sanity runs links to description.

PR #4270, opened 2025-06-16
Update Dockerfile

raghu-microsoft:
image


vizhur (changes requested):
no dummy changes needed, just release with automated update that picked up aifx image update

PR #4339, opened 2025-07-20
Fix vulnerabilities in Training

SamGos93:


assets/training/automl/environments/ai-ml-automl-dnn-gpu/context/Dockerfile
```diff
@@ -13,10 +13,36 @@ ENV MLFLOW_MODEL_FOLDER="mlflow-model"
...
+        libnss-systemd=249.11-0ubuntu3.16 \
+        libpython3.10-stdlib=3.10.12-1~22.04.10 \
+        python3.10=3.10.12-1~22.04.10 \
+        libpython3.10-minimal=3.10.12-1~22.04.10 \
+        python3.10-minimal=3.10.12-1~22.04.10 \
+        libpam0g=1.4.0-11ubuntu2.6 \
+        libpam-modules-bin=1.4.0-11ubuntu2.6 \
+        libpam-modules=1.4.0-11ubuntu2.6 \
+        libpam-runtime=1.4.0-11ubuntu2.6 \
+        libarchive13=3.6.0-1ubuntu1.5
```
iamrk04, line 36: Is it possible to not pin the packages and let the latest being pulled. I am assuming that the latest version is same as being pinned here. Feel free to correct me. I see this happening at multiple places.

iamrk04 (changes requested):

assets/training/automl/environments/ai-ml-automl-dnn-gpu/context/Dockerfile
iamrk04: Problem with pinning is that we restrict the version which means someone need to manually clean them up or bump them in the next vulnerability fix. Unless required, we should avoid pinning packages so that it can auto pull the latest one and does need manual intervention. Pinning python packages is understandable, but we should avoid pinning ubuntu packages.

iamrk04 (approved):

PR #4464, opened 2025-09-30
Fixing training and project_24_25 vulns as on 29 Sep 2025

scravikiran:
Fixing training and project_24_25 vulns as on 29 Sep 2025

### DIFF
--- a/assets/training/automl/environments/ai-ml-automl-dnn-text-gpu/context/Dockerfile
+++ b/assets/training/automl/environments/ai-ml-automl-dnn-text-gpu/context/Dockerfile
@@ -99,8 +99,8 @@ RUN pip list && \
     pip install pyarrow==14.0.2 \
                 'transformers[sentencepiece,torch]==4.53.0' \
                 aiohttp==3.12.14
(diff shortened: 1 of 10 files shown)


assets/training/automl/environments/ai-ml-automl-dnn-text-gpu/context/Dockerfile
```diff
@@ -69,25 +69,26 @@ SHELL ["conda", "run", "-p", "$AZUREML_CONDA_ENVIRONMENT_PATH", "/bin/bash", "-c
...
-                azureml-defaults=={{latest-pypi-version}} \
                 'azure-identity>=1.16.1' \
                 'sentencepiece==0.1.97' \
                 'cryptography>=42.0.5'\
                 'requests>=2.32.4' \
                 'certifi>=2023.07.22' 
                 # end pypi dependencies
 
+# Install conflicting packages separately with --no-deps
+RUN pip install 'azureml-defaults=={{latest-pypi-version}}' --no-deps --force-reinstall
```
novaturient95, line 89: we should not try force install, if they are fixable in base packages. Could you add what conflict we are trying to avoid here?

PR #4941, opened 2026-04-18
Fix grpo, rft and image_mmdetection packages for vulnerabilities

sharvin2187:


assets/training/finetune_acft_image/environments/acft_image_mmdetection/context/Dockerfile
```diff
@@ -2,27 +2,36 @@
...
+    apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Install required packages from pypi
 COPY requirements.txt .
 RUN pip install -r requirements.txt --no-cache-dir
-RUN apt-get install -y expat
+RUN apt-get update && apt-get install -y expat && apt-get clean && rm -rf /var/lib/apt/lists/*
+# openmim still imports pkg_resources, which is removed from setuptools>=82.
+# Keep a temporary compatibility pin for mim installation, then restore setuptools.
+RUN pip install --no-cache-dir 'setuptools<82'
```
novaturient95, line 15: will this lead to any import issue?

2ndly, mim and mmdet GH does not have any contributions for past 3 years. We will have to patch up changes all the time. I will support deprecating env if they dont server any production use case. Could we discuss internally if any vision env or components are still in use?

sharvin2187: It may cause an import issue but it is still our best bet, open to any other options.

Started a thread internally to discuss the second concern.

yeshsurya (approved):

PR #4946, opened 2026-04-21
Rename trace to turn in evaluation level

salma-elshafey:
updating the multi turn contract to accept "conversation" and "turn" as supported Evaluation levels


assets/evaluators/builtin/customer_satisfaction/spec.yaml
```diff
@@ -10,7 +10,7 @@ tags:
   provider: "Microsoft"
   preview: "true"
```
AliMahmoudzadeh, line 11: don't we need to update the version as well?

AliMahmoudzadeh (approved):

ahibrahimm (approved):

PR #4947, opened 2026-04-21
Quality Grader Evaluator (Draft PR)

ahibrahimm:


assets/evaluators/tests/common/evaluator_mock_config.py
```diff
@@ -155,5 +244,9 @@ def get_flow_side_effect_for_evaluator(
             f"Evaluator '{evaluator_name}' not recognized. Available evaluators: {list(EVALUATOR_CONFIGS.keys())}"
         )
 
+    # Quality grader has a custom output format for stage 1
+    if evaluator_name == "quality_grader":
```
m7md7sien: Can you remove this check, and make the default flow side effect for quality grader is response_quality, and the groundedness step is already handled separately.

ahibrahimm: we talked about it offline, addressing it at the moment

m7md7sien (approved):

PR #4983, opened 2026-04-30
Handle empty evaluation_level for backwards compatibility

salma-elshafey:
Fix: Handle empty string evaluation_level in TaskCompletionEvaluator

Production failures occurred when evaluation_level is passed as '' (empty string) by the platform -not yet updated with the multi-turn changes- which isn't a valid EvaluationLevel enum value. The fix treats empty strings the same as None, defaulting to auto-detect mode instead of raising an error.

### DIFF
--- a/assets/evaluators/builtin/customer_satisfaction/evaluator/_customer_satisfaction.py
+++ b/assets/evaluators/builtin/customer_satisfaction/evaluator/_customer_satisfaction.py
@@ -3,7 +3,7 @@
 import math
 import os
 import logging
-from typing import Dict, Union, List, Optional
+from typing import Dict, Union, List, Optional, Tuple
 
 from typing_extensions import overload, override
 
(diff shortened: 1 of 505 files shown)


assets/evaluators/builtin/customer_satisfaction/evaluator/_customer_satisfaction.py
```diff
@@ -96,7 +96,7 @@ def _resolve_evaluation_level(
     :rtype: Optional[EvaluationLevel]
     """
     valid = [level.value for level in EvaluationLevel]
-    if evaluation_level is None:
+    if evaluation_level is None or evaluation_level == '':
```
imatiach-msft, line 99: nit: maybe add some tests to validate this edge case, can be in separate PR

imatiach-msft (dismissed):

assets/evaluators/builtin/customer_satisfaction/evaluator/_customer_satisfaction.py
salma-elshafey: added

imatiach-msft (dismissed):

m7md7sien (approved):

imatiach-msft (approved):

PR #5033, opened 2026-05-12
Update the storage account for Foundry Local models

hanbitmyths:
Foundry Local models are published via IPP pipeline, which requires a storage account under AME. Replication policy is set up from foundrylocalmodels to foundrylocalassetdata with SLA of 15 minutes.


rui-ren (approved):
👍

PR #5063, opened 2026-05-19
Amah/end with user

AliMahmoudzadeh:
handle this WI https://msdata.visualstudio.com/Vienna/_workitems/edit/5281684

clients might bring in data that ends with Assistant


assets/evaluators/builtin/coherence/evaluator/_coherence.py
salma-elshafey: we need to update this as well to be "last message", not specifically last assistant message

assets/evaluators/builtin/customer_satisfaction/evaluator/_customer_satisfaction.py
```diff
@@ -164,12 +164,14 @@ def __init__(
         self,
         error_target: ErrorTarget,
         requires_query: bool = True,
-        check_for_unsupported_tools: bool = False
+        check_for_unsupported_tools: bool = False,
+        evaluation_level: Optional[EvaluationLevel] = None,
```
salma-elshafey: There's an inconsistency between csat and the rest of the evaluators, csat adds evaluation_level to ConversationValidator, meanwhile the others add it to MessagesOrQueryResponseInputValidator

assets/evaluators/builtin/coherence/evaluator/_coherence.py
```diff
@@ -644,7 +659,7 @@ def validate_eval_input(self, eval_input: Dict[str, Any]) -> bool:
                     category=ErrorCategory.INVALID_VALUE,
                     target=self.error_target,
                 )
-            if messages[-1]["role"] != MessageRole.ASSISTANT:
+            if self._evaluation_level == EvaluationLevel.TURN and messages[-1]["role"] != MessageRole.ASSISTANT:
```
salma-elshafey: I think a cleaner way is to pop the "messages" field from the kwargs in the TURN branch of _real_call (after the messages are split into query and response)
and then we can remove this condition entirely, and the changes done to the validator's initialization.
This way, when the evaluation level is turn, it will not pass through these validation conditions (rather the existing ones for query/response), and when it is conversation, there will be no condition that validates that the last message must be assistant.

salma-elshafey (changes requested):

imatiach-msft (dismissed):

salma-elshafey (dismissed):

salma-elshafey (approved):

PR #5066, opened 2026-05-20
Add slime-train PATH shim for cwd-agnostic invocation

neilyan-msft:
## Summary

Adds a small ``slime-train`` shim at ``/usr/local/bin/slime-train`` so consumers can launch the slime training entrypoint from any working directory.

## Motivation

Slime's entrypoint is the repo-root ``train.py`` script (not a Python module), so ``python train.py`` resolves against the caller's cwd. AML/Singularity sets cwd to a per-job working directory (e.g. ``/scratch/azureml/cr/j/{uuid}/exe/wd/``) that does not contain ``train.py``. As a result, jobs that compose their command as ``python train.py ...`` against this curated environment fail immediately with:

```
python: can't open file '/scratch/azureml/cr/j/{uuid}/exe/wd/train.py': [Errno 2] No such file or directory
```

(observed end-to-end on canary, jobs ``tb-slime-{sft,grpo}-gsm8k-qwen05b-05191606``, after the ``/root/slime`` -> ``/opt/slime`` move in #5046 unblocked image readability.)

## Change

- **Dockerfile**: install ``/usr/local/bin/slime-train`` (mode 0755) that ``cd``s to ``/opt/slime`` and ``exec``s ``python /opt/slime/train.py "$@"``. The ``cd`` keeps any relative artefact paths the slime entrypoint resolves at runtime anchored to the slime source tree.
- **smoke_test.py**: assert the shim is present, world readable+executable, and references ``/opt/slime/train.py``.

## Why not change consumers instead

Every other curated training environment (TRL, VERL, OpenRLHF, NeMo-RL, TorchForge) uses ``python -m <module>`` and is therefore cwd-independent. Slime is the only one whose entrypoint is a bare script file, so the cwd contract has to be solved either by hard-coding the absolute path in every consumer or by giving the asset a PATH-resolvable name. The latter keeps the layout knowledge inside the asset that owns it.

## Validation

Local Python parse check on ``smoke_test.py``. Image build + smoke test will validate the shim during the asset CI build.

### DIFF
--- a/assets/training/finetune_acft_hf_nlp/environments/slime-pytorch-2.9-cuda12.8/context/Dockerfile
+++ b/assets/training/finetune_acft_hf_nlp/environments/slime-pytorch-2.9-cuda12.8/context/Dockerfile
@@ -115,6 +115,18 @@ RUN cd /opt/slime/slime/backends/megatron_utils/kernels/int4_qat && \
 # Megatron-LM import all succeed without requiring `--user root`.
 RUN chmod -R a+rX /opt/slime /opt/Megatron-LM
 
+# Install a PATH-resolvable `slime-train` shim. Slime's entrypoint is the
+# repo-root `train.py` script (not a Python module), so `python train.py`
+# must resolve against the cwd. AML/Singularity sets cwd to a per-job
+# working directory (e.g. /scratch/azureml/cr/j/{uuid}/exe/wd/) that does
+# not contain train.py, so consumers cannot invoke slime via the bare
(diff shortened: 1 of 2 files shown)


neilyan-msft:
Closing — switching to a consumer-side fix in Vienna MFE (hard-code /opt/slime/train.py in TrainingRecipeCompiler instead of adding a PATH shim in the asset). The slime asset stays cwd-coupled, but that contract already exists implicitly and slime is the only direct-script framework.

PR #5093, opened 2026-05-29
Resolve dependencies while keeping vulns out

sharvin2187:
resolve deps


SamGos93 (dismissed):

assets/training/automl/environments/ai-ml-automl/context/Dockerfile
vizhur, line 46: remove

assets/training/automl/environments/ai-ml-automl/context/Dockerfile
```diff
@@ -108,6 +108,11 @@ RUN pip install --upgrade \
     'onnx>=1.21.0' \
     'pillow>=12.2.0' \
     'python-dotenv>=1.2.2'
+RUN pip install --no-deps --force-reinstall 'setuptools==81.0.0'
```
vizhur: that is the 3rd setuptools installation in this dockerfile

vizhur (changes requested):

assets/training/automl/environments/ai-ml-automl/context/Dockerfile
sharvin2187: makes sense, updated in the above block.

assets/training/automl/environments/ai-ml-automl/context/Dockerfile
sharvin2187: The comment above makes sense to me:

```# Security: upgrade pip to fix CVE-2026-6357 (GHSA-jp4c-xjxw-mgf9). Two pips need
# upgrading because they live in independent prefixes:
#   1. /opt/miniconda/bin/pip — ships with the parent base image
#      (mcr.microsoft.com/azureml/openmpi5.0-ubuntu24.04) at version 26.0.1.
#      VCM SCA scans this site-packages tree at
#      opt/miniconda/lib/python3.10/site-packages/pip-26.0.1.dist-info/METADATA.
#   2. $AZUREML_CONDA_ENVIRONMENT_PATH/bin/pip — created by `conda create` above.
# pip is its own parent (no upstream package can pull in a fixed pip), so explicit
# upgrades are the only available remediation.
```

Are we sure this wont cause vulns ?

sharvin2187: This change was made 3 weeks ago, I doubt if that is safe enough timeframe to expect the source to have fixed it.

vizhur (approved):

PR #5042, opened 2026-05-15
Skipped status and standardize output

ashaabansoliman:
This PR is collaborative effort between Mohamed Hussein and Ahmed Shaaban to:
-Update and standardize evaluator output schema
-Provide new field "status" to indicate if the evaluation completed, resulted in an error or skipped
Note: The commits of this branch were created in multi-stages on multi-branches by both authors using Copilot

PR #5107, opened 2026-06-04
Fix TypeError when evaluator returns None score for skipped/not-applicable results

vebudumu:
When _return_not_applicable_result sets score=None (introduced in PR #5042), the math.isnan() check in _do_eval crashes with:
  TypeError: must be real number, not NoneType

Guard math.isnan() with a None check so skipped/not-applicable results pass through correctly instead of raising an unhandled TypeError.

Affected evaluators: groundedness, coherence, fluency, retrieval, similarity.

### DIFF
--- a/assets/evaluators/builtin/similarity/evaluator/_similarity.py
+++ b/assets/evaluators/builtin/similarity/evaluator/_similarity.py
@@ -389,8 +389,9 @@ class SimilarityEvaluator(PromptyEvaluatorBase):
         :rtype: Dict
         """
         result = await self._the_super_do_eval(eval_input)
-        # Check if base returned nan (invalid output case)
-        if math.isnan(result.get(self._result_key, 0)):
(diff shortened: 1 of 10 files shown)


ashaabansoliman (approved):

imatiach-msft (approved):

assets/evaluators/builtin/similarity/evaluator/_similarity.py
```diff
@@ -389,8 +389,9 @@ async def _do_eval(self, eval_input: Dict):  # type: ignore[override]
         :rtype: Dict
         """
         result = await self._the_super_do_eval(eval_input)
```
imatiach-msft, line 391: maybe not for this PR but please add tests - maybe as a separate PR given the urgency

aprilk-ms (approved):

PR #5108, opened 2026-06-04
Fix credscan issue 

ashaabansoliman:


assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
```diff
@@ -1053,9 +1053,10 @@ def _reformat_tool_calls_results(response, logger=None):
             # fallback to the original response in that case
             if logger:
                 logger.warning(
-                    f"Empty agent response extracted, likely due to input schema change. "
```
shrutiyer, line 1056: This change looks good. There is actually a second line also logging response

Agent response could not be parsed, falling back to original response:

PR #5117, opened 2026-06-05
Enable ToolCallAccuracy/InputAccuracy/CallSuccess on restricted-tool conversations

mmkawale:
## Summary

Lifts the unconditional rejection of conversations containing built-in restricted tools (`bing_grounding`, `bing_custom_search`, `azure_ai_search`, `azure_fabric`, `sharepoint_grounding`, `browser_automation`, `code_interpreter_call`, `computer_call`, `openapi_call`, `web_search`) on the three tool-call evaluators (`tool_call_accuracy`, `tool_input_accuracy`, `tool_call_success`).

These three evaluators grade the agent's **tool selection**, **input arguments**, and **call status** respectively. None of them consume the (redacted) tool output body, so the previous blanket rejection was over-broad. `tool_output_utilization` and `groundedness` continue to reject restricted-tool conversations because they *do* read the tool output body.

This is the registry-asset side of the change. The matching SDK change ships in [Azure/azure-sdk-for-python#47369](https://github.com/Azure/azure-sdk-for-python/pull/47369) (`azure-ai-evaluation 1.17.1`). Because the registry hosts forked copies of the evaluator source under `assets/evaluators/builtin/<name>/evaluator/`, the same one-line flip has to be duplicated here per-evaluator.

### Source (`check_for_unsupported_tools=True` → `False`)
- `assets/evaluators/builtin/tool_call_accuracy/evaluator/_tool_call_accuracy.py` — `ToolCallsValidator`
- `assets/evaluators/builtin/tool_input_accuracy/evaluator/_tool_input_accuracy.py` — `ToolDefinitionsValidator`
- `assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py` — `ToolDefinitionsValidator`

### Spec version bumps
| Asset | Before | After |
|---|---|---|
| `tool_call_accuracy/spec.yaml` | 11 | 12 |
| `tool_input_accuracy/spec.yaml` | 12 | 13 |
| `tool_call_success/spec.yaml` | 7 | 8 |

### Tests
- Flipped `check_for_unsupported_tools` `True → False` in the three behavior test suites so assertions match the new behavior (validator accepts → flow runs).
- Relaxed `base_tool_evaluation_test._run_tool_type_test`: when a tool's `expected_flow_inputs` is not yet populated (empty dict), assert only that the flow was invoked once instead of exact-argument matching. The full per-tool expected-flow constants for the newly-enabled tool types will land in a follow-up PR via a mechanical generation pass over the existing `_QUERY` / `_RESPONSE` / `_TOOL_DEFINITIONS` fixtures.

## Test results

228 / 228 behavior tests pass across the three suites:
- `test_tool_call_accuracy_evaluator_behavior`
- `test_tool_call_success_evaluator_behavior`
- `test_tool_input_accuracy_evaluator_behavior`

## Follow-ups (not in this PR)

1. **Port the `_ToolCallSuccess` status short-circuit** (the other half of [SDK #47369](https://github.com/Azure/azure-sdk-for-python/pull/47369)) into `assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py`, with `spec.yaml` v8 → v9 and a mirrored short-circuit test suite. Without that follow-up, the cloud-managed `tool_call_success` evaluator will accept restricted-tool conversations (after this PR) but will still send failed runs through the LLM rather than deterministically returning `fail`.
2. Backfill per-tool `expected_flow_inputs` fixtures for the newly-enabled tool types so `base_tool_evaluation_test` resumes exact-argument matching.


salma-elshafey:
Outputs for tools are needed for the `tool_call_accuracy`, `tool_input_accuracy`, and `tool_call_success`
For the first 2, the output of a tool can be used as an input for the next tool, so it would need the output to be able to ground the inputs of the next tool.
For the last one, the output of the tool is needed to know whether the tool failed or not.

mmkawale:
@salma-elshafey Thanks for your feedback. You have a good point, but here's my take:

1. Today we've disabled all these evaluators on unsupported tools, which is zero customer value. Instead, I think we should score on what we can see (user query + tool name + tool definition + args, and status for TCS).
2. Multi-tool grounding is partially captured by the inputs we always see. The agent will pass the relevant info from the prior tool as input to the subsequent tool, and TIA scores those args appropriately ΓÇö if the agent failed to ground correctly, the args will look incoherent and TIA scores them low, which is the desired behavior.
3. For TCS, we'll be passing in the status correctly ΓÇö that's what the evaluator keys off of to determine success/failure, no tool output body needed.
4. This is part of a series ΓÇö next change enables SP / Fabric / Azure AI Search end-to-end (they do expose tool outputs, so `tool_output_utilization` comes online too). Would love your feedback on the broader plan: https://dev.azure.com/msdata/Vienna/_git/Observability-Specs/pullrequest/2139214.

mmkawale:
@salma-elshafey Thanks for your feedback. You have a good point, but here's my take:

1. Today we've disabled all these evaluators on unsupported tools, which is zero customer value. Instead, I think we should score on what we can see (user query + tool name + tool definition + args, and status for TCS).
2. Multi-tool grounding is partially captured by the inputs we always see. The agent will pass the relevant info from the prior tool as input to the subsequent tool, and TIA scores those args appropriately — if the agent failed to ground correctly, the args will look incoherent and TIA scores them low, which is the desired behavior.
3. For TCS, we'll be passing in the status correctly — that's what the evaluator keys off of to determine success/failure, no tool output body needed.
4. This is part of a series — next change enables SP / Fabric / Azure AI Search end-to-end (they do expose tool outputs, so `tool_output_utilization` comes online too). Would love your feedback on the broader plan: https://dev.azure.com/msdata/Vienna/_git/Observability-Specs/pullrequest/2139214.

mmkawale:
Superseded by #5126, opened from a same-repo branch. CI on #5117 was blocked by the no-fork-PR policy in .github/workflows/check-execution-context.yaml; same two commits (81ec0accb + ef7bdd3d) cherry-picked cleanly onto current main. Please review there.

PR #5131, opened 2026-06-11
Rollback unreleased MMDetection component bumps

sharvin2187:
## What
Rollback the unreleased MMDetection component and pipeline component version bumps introduced by the recent MMDetection environment fixes.

This restores:
- `mmdetection_image_objectdetection_instancesegmentation_model_import` to `0.0.22`
- `mmdetection_image_objectdetection_instancesegmentation_finetune` to `0.0.24`
- image object detection / instance segmentation / MMDetection pipeline component refs to the currently released component versions

## What this does not change
The `acft-mmdetection-image-gpu` environment asset changes are kept intact. The environment has already been released, so this PR only rolls back component source versions/refs that have not been released.

## Why
The component versions introduced by the prior PRs have not been released. Rolling them back avoids publishing a large component/runtime jump while we mitigate the customer issue separately with a pipeline override to the last known good path.

### DIFF
--- a/assets/training/finetune_acft_image/components/finetune/mmd_od_is/spec.yaml
+++ b/assets/training/finetune_acft_image/components/finetune/mmd_od_is/spec.yaml
@@ -1,14 +1,14 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 type: command
 
-version: 0.0.25
+version: 0.0.24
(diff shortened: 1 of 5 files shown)


novaturient95 (approved):

vizhur (changes requested):
please justify why do you revert environment update and back to the vulnerable one

PR #5145, opened 2026-06-15
ToolCallSuccess: move runtime-status short-circuit from prompt into Python

mmkawale:
## Summary

Follow-up to #5126. Moves the runtime-status short-circuit out of the LLM rubric and into Python, so failed/incomplete tool calls return a deterministic fail result without spending a judge call. The prompty rubric is now consulted only on the success path.

## Why

- The LLM was being asked to interpret a `[STATUS] failed` / `[STATUS] incomplete` annotation that we already know deterministically at evaluator time.
- That made every failure scorable, judge-model-version dependent, and a token spend.
- The status annotation is also surfaced to the model when the upstream payload is the source of truth, which is back-compat noise in the formatted wire input.

## What changed

**`_tool_call_success.py` (both repos):**
- `_do_eval` now calls `_collect_failed_tool_calls(eval_input["response"])` after preprocessing. If non-empty, returns `_return_short_circuit_failure_result(failed_tools)` early — no LLM call, no prompty render.
- New `_collect_failed_tool_calls` helper: two-pass scanner over messages, treats `status in {"failed", "incomplete"}` on either a `tool_call` block or its corresponding `tool_result` as a failure. Handles flat (`name`, `tool_call_id`) and nested (`tool_call.function.name`, `tool_call.id`) shapes. Falls back to id or name when one is missing, dedupes by (id, name).
- New `_return_short_circuit_failure_result`: returns `score=0.0`, `result="fail"`, `status="completed"`, `properties={"failed_tools": ...}`. Assets variant additionally includes top-level token-metadata keys for back-compat with the existing LLM path result shape.
- `_get_tool_calls_results` rewritten — no `_format_status_suffix` calls; output is byte-identical to the pre-pass-through format.
- `_format_status_suffix` removed.

**`tool_call_success.prompty` (both repos):**
- Removed the `[STATUS] failed` / `[STATUS] incomplete` bullet from the rubric.
- Removed step 3 paragraph describing `[STATUS]` as optional.
- Removed three `[STATUS]`-flavored examples ("status annotation overrides bland payload", "status completed but payload still indicates an error", "parallel calls in one turn, one annotated failed").

**Tests:**
- Deleted `test_tool_call_success_status_passthrough.py`.
- Added `test_tool_call_success_status_shortcircuit.py`: 19 tests covering both helpers — failed-on-call, failed-on-result, incomplete-treated-as-failure, dedupe, parallel calls, nested function shape, missing-id-falls-back-to-name, malformed-blocks-skipped-silently, non-list-input, no-status-forwarded-in-LLM-input, etc.

## Validation

- `pytest test_tool_call_success_status_shortcircuit.py` — 19/19 passed.
- `pytest test_tool_call_success_evaluator_behavior.py` — 69/69 passed (no regression).

## Related

- SDK side: Azure/azure-sdk-for-python#47462
- Upstream aca status plumbing: Vienna pullrequest/2139056


assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
```diff
@@ -855,6 +855,40 @@ def _return_not_applicable_result(
...
+        """Return a deterministic fail result without invoking the LLM judge.
+
+        Used when the runtime explicitly marks one or more tool calls as
+        failed/incomplete via the ``status`` field on a ``tool_call`` or
+        ``tool_result`` content block. The LLM call is skipped because the
+        runtime signal is authoritative; token-metadata fields are emitted
+        with zero/empty values for schema compatibility with the LLM path.
+        """
+        failed_list = ",".join(failed_tools)
+        reason = (
```
aprilk-ms, line 963: Does user need to know we didn't invoke LLM judge? Seems unnecessary.

assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
```diff
@@ -855,6 +855,40 @@ def _return_not_applicable_result(
...
+            self._result_key: 0.0,
+            f"{self._result_key}_score": 0.0,
+            f"{self._result_key}_passed": False,
+            f"{self._result_key}_result": "fail",
+            f"{self._result_key}_reason": reason,
+            f"{self._result_key}_status": "completed",
+            f"{self._result_key}_threshold": self._threshold,
+            f"{self._result_key}_properties": {
+                "failed_tools": failed_list,
+                **token_metadata,
```
aprilk-ms, line 978: do we need to output tokens with LLM calls?

assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
```diff
@@ -1004,14 +1048,86 @@ def _format_value(v):
     return v
 
 
+_FAILED_RUNTIME_STATUSES = frozenset({"failed", "incomplete"})
```
aprilk-ms, line 1171: How complete is this list? Can our upstream add a new one without us knowing?

assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
mmkawale: I agree. I removed the details. The short-circuit message now only states that runtime status was non-success (failed or incomplete), without mentioning judge invocation.

assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
mmkawale: Agreed. On the short-circuit path we no longer emit LLM token/model metadata fields. Those are now only present for the actual LLM-evaluated path.

assets/evaluators/builtin/tool_call_success/evaluator/_tool_call_success.py
mmkawale: I had verified ToolCallStatus in the Foundry Conversations API: https://ai.azure.com/api-reference/conversations/get-conversation-item/. 
I am adding a warning when we observe any non-empty status outside that known set, so we’ll get telemetry/log signal if upstream extends the enum and can update handling promptly.

aprilk-ms (approved):

PR #5165, opened 2026-06-22
User/eswar/vuln fix 19062026 project24 25

eswar1996b:


assets/training/finetune_acft_hf_nlp/environments/acpt-rft/context/verl_rft_compat_patches
```diff
@@ -138,12 +138,12 @@ _TF_SAUX_NEW = (
 # is an optional acceleration; fall back to vLLM's native rotary when it is absent.
 # --------------------------------------------------------------------------------------
 _VLLM_ROTARY_OLD = '''        self.apply_rotary_emb_flash_attn = None
-        if find_spec("flash_attn") is not None:
+        if not current_platform.is_cpu() and find_spec("flash_attn") is not None:
```
yeshsurya: Why this is required for vulnerability fixes ?

yeshsurya (changes requested):

assets/training/finetune_acft_hf_nlp/environments/acpt-rft/context/verl_rft_compat_patches
eswar1996b: Here's the reply i got from copilot.

with the current package pins it's genuinely needed. Here's the evidence from the Dockerfile:

Line 46: pip install vllm==0.20.1
Line 51: pip install 'flash-attn-4>=4.0.0b9' → the image installs flash-attn 4, which registers the flash_attn namespace (flash_attn.cute) but does not have the flash-attn 2.x flash_attn.ops.triton.rotary module.
The failure chain at runtime:

vLLM 0.20.1's rotary_embedding/common.py does an unguarded if find_spec("flash_attn") is not None: from flash_attn.ops.triton.rotary import apply_rotary.
flash-attn 4 makes find_spec("flash_attn") truthy, so it enters the branch.
The import then raises ModuleNotFoundError because flash_attn.ops doesn't exist in v4.
That code runs when the RoPE class initializes during model load → GRPO rollout crashes.
The patch is anchor-verified: the script calls _replace("...rotary_embedding.common", _VLLM_ROTARY_OLD, ...) and, per its own docstring, fails the build loudly if the anchor text is absent. Since the image builds, that unguarded import is confirmed present in vLLM 0.20.1. So without the fallback patch, the image would break at runtime.

yeshsurya: It is runtime thing which model is trying to see by checking code, did it read actual vllm code for that version to come back with this analysis? or it is hallucinating. But this runtime behavior needs to be tested by running actual RL job. I see risk of having this check is import will get limited to only on CPU device and it might break other things. If image is building successfully let's remove this check for now. We can add this back when we do actual RL test using the image.

yeshsurya (approved):

PR #5247, opened 2026-07-30
fix: verify AutoML model signatures before loading

babu-namburi:
## Summary
- require detached RSA PKCS#1 v1.5 SHA-256 signatures before loading AutoML forecasting model artifacts
- deserialize only staged bytes that were successfully verified, preventing model replacement between verification and loading
- reject unsigned, tampered, unsupported, and malformed-signature artifacts before `pickle.load` or `torch.load`
- bump `automl_forecasting_inference` to `0.0.16` and document the trusted public-key requirement

## Security context
Remediates CWE-502 reported under IcM 842340960 while preserving compatibility with existing AutoML pickle and full-object PyTorch artifacts. Secure `skops` migration is currently blocked because `skops>=0.12` requires NumPy >=1.25 while `azureml-train-automl-runtime==1.62.0` requires NumPy <=1.23.5.

The trusted pipeline operator must set `AUTOML_MODEL_SIGNING_PUBLIC_KEY_PEM`; each model must include a sibling `<model>.sig` detached signature.

## Why the vulnerability is resolved
The model input no longer flows directly into an unsafe deserialization sink:

1. The model is copied to a temporary staging file while its SHA-256 digest is calculated.
2. The digest is verified against a detached RSA PKCS#1 v1.5 signature using the operator-controlled public key.
3. Only the exact staged bytes that passed signature verification are supplied to `pickle.load()` or `torch.load()`.
4. Missing, malformed, forged, or tampered signatures fail before deserialization.

Staging and loading the same verified copy also prevents a time-of-check/time-of-use replacement between verification and deserialization.

Regression coverage verifies that a malicious pickle `__reduce__` payload is not executed when:
- the signature is missing
- the model is modified after signing
- the signature has an invalid size or value
- the trusted public key is unavailable

A production verification can sign an existing model without changing its bytes:

```bash
openssl dgst -sha256 -sign private.pem -out model.pkl.sig model.pkl
```

Inference should succeed with the signed model. Modifying any model byte afterward must produce `Model signature verification failed` before the model-loading log and before any payload executes.

This guarantee depends on protecting the private signing key and ensuring the model supplier cannot override `AUTOML_MODEL_SIGNING_PUBLIC_KEY_PEM`. The trusted key must be configured by the pipeline or platform operator, not supplied by the same principal that supplies the model.

## Backward compatibility
The model serialization and inference contracts are preserved:
- existing `.pkl`, `.pt`, and `.pth` formats remain supported
- existing model bytes do not need retraining, conversion, or reserialization
- the existing `pickle.load()` and full-object `torch.load()` behavior is retained after provenance verification
- component inputs, outputs, and forecasting behavior are unchanged

This is intentionally not zero-touch compatible with unsigned artifacts. Existing models must receive a sibling detached signature and the trusted public key must be configured. Component version `0.0.15` remains available for existing pipelines, while `0.0.16` introduces signature enforcement, allowing a controlled migration.

## Validation
- `python -m pytest assets\training\forecasting_demand\automl-single-model-inference\tests\test_model_loader.py -q`
- `python scripts\validation\doc_style.py -i .`
- `python -m py_compile assets\training\forecasting_demand\automl-single-model-inference\src\model_loader.py assets\training\forecasting_demand\automl-single-model-inference\src\inference.py`

Remediates CWE-502 from IcM 842340960.

### DIFF
--- a/assets/training/forecasting_demand/automl-single-model-inference/src/model_loader.py
+++ b/assets/training/forecasting_demand/automl-single-model-inference/src/model_loader.py
@@ -1,0 +1,134 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Load AutoML forecasting models only after verifying their provenance."""
+
+import hashlib
+import os
+import pickle
+import tempfile
+
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import padding, rsa, utils
+
+try:
+    import torch
+
+    _torch_present = True
+except ImportError:
+    _torch_present = False
+
+
+MODEL_SIGNING_PUBLIC_KEY_ENV_VAR = "AUTOML_MODEL_SIGNING_PUBLIC_KEY_PEM"
+PTH_FILE_POSTFIX = ".pth"
+PT_FILE_POSTFIX = ".pt"
+PICKLE_FILE_POSTFIX = ".pkl"
+SIGNATURE_FILE_POSTFIX = ".sig"
+_SUPPORTED_MODEL_POSTFIXES = (PICKLE_FILE_POSTFIX, PT_FILE_POSTFIX, PTH_FILE_POSTFIX)
+_HASH_CHUNK_SIZE = 1024 * 1024
+
+
+def _load_signing_public_key():
+    public_key_pem = os.environ.get(MODEL_SIGNING_PUBLIC_KEY_ENV_VAR)
+    if not public_key_pem:
+        raise ValueError(
+            f"{MODEL_SIGNING_PUBLIC_KEY_ENV_VAR} must contain the trusted RSA public key "
+            "used to verify model artifacts."
+        )
+
+    try:
+        public_key = serialization.load_pem_public_key(public_key_pem.encode("utf-8"))
+    except (TypeError, ValueError) as ex:
+        raise ValueError(
+            f"{MODEL_SIGNING_PUBLIC_KEY_ENV_VAR} does not contain a valid PEM public key."
+        ) from ex
+
+    if not isinstance(public_key, rsa.RSAPublicKey):
+        raise ValueError("The model-signing public key must be an RSA public key.")
+
+    return public_key
+
+
+def _stage_and_verify_model(model_full_path):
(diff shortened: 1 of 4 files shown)


assets/training/forecasting_demand/automl-single-model-inference/src/model_loader.py
```diff
@@ -0,0 +1,134 @@
...
+PTH_FILE_POSTFIX = ".pth"
+PT_FILE_POSTFIX = ".pt"
+PICKLE_FILE_POSTFIX = ".pkl"
+SIGNATURE_FILE_POSTFIX = ".sig"
+_SUPPORTED_MODEL_POSTFIXES = (PICKLE_FILE_POSTFIX, PT_FILE_POSTFIX, PTH_FILE_POSTFIX)
+_HASH_CHUNK_SIZE = 1024 * 1024
+
+
+def _load_signing_public_key():
+    public_key_pem = os.environ.get(MODEL_SIGNING_PUBLIC_KEY_ENV_VAR)
```
msp8955, line 33: I'm not sure what is happening here..
where is this being set? is MODEL_SIGNING_PUBLIC_KEY_ENV_VAR used when training run is run and saved with, and you are trying to verify here during load time that this is trusted model output-ed from training time? I can't find the reference to this ENV_VAR anywhere though.

msp8955: also is this tested? ai agent is suggesting it may break forecasting scenario:


No—the branch does not fix the issue as implemented and breaks existing inference.

• The verification public key comes from  AUTOML_MODEL_SIGNING_PUBLIC_KEY_PEM . A caller can provide their own key, sign a malicious pickle, and pass verification.
• Existing AutoML  best_model  outputs have no  .sig  file, so normal forecasting pipelines fail.
• No AutoML producer-side signing or trusted key provisioning was added.
• After verification, it still uses unrestricted  pickle.load()  and  torch.load() .
• Tests use a synthetic dictionary, not real sklearn or TCN forecasting models.

The signature design works only if AutoML signs artifacts and the inference component uses a service-controlled, non-overridable public key. Do not merge this branch as-is. The smaller compatible fix is enforcing that  model_path  is directly bound to an AutoML forecasting job’s  best_model  output.

babu-namburi: This is not tested and work is in progress.

PR #5262, opened 2026-08-05
Optimize evaluator prompt layout for Azure OpenAI prefix caching

bonanza89:
Azure OpenAI caches the longest byte-identical prefix shared across requests, and only once that prefix reaches 1024 tokens. Several builtin evaluator prompts interleave their static rubric with substituted variables, so the first row-varying value appears within the first few hundred characters. That truncates the cacheable prefix far below the 1024-token floor and defeats prefix caching entirely.

This change reorders the system message of 19 evaluator prompts into:

    [static rubric] -> [run-constant variables] -> [row-varying variables]

Run-constant variables (tool_definitions, taxonomy, tool_descriptions, system_message) hold the same value for every row of an evaluation run, so placing them early extends the shared prefix. Row-varying variables (query, response, context, conversation, ...) change on every row and must come last.

The edits are strictly positional: no rubric wording, scoring criteria, few-shot example, output schema, or variable name is altered. Every non-blank line of each prompt is preserved and only the order of blocks changes, so the prompts are expected to be behaviourally identical.

Cacheable prefix, measured with tiktoken o200k_base (before -> after):

image

  * task_adherence_multi_turn crosses the 1024-token threshold at runtime, once the run-constant tool_definitions block is substituted.

Eleven of these prompts previously had a cacheable prefix below the 1024-token floor and so could never produce a cache hit at all. Ten now clear the floor statically and task_adherence_multi_turn clears it at runtime as noted above. The remaining eight already cached, and gain a longer cached prefix.

Deliberately not changed: fluency and similarity have static rubrics that are too short (895 and 803 tokens) to reach the 1024-token floor even when ordered optimally, so reordering them would add churn for no benefit.

Asset versions are bumped for the 13 affected evaluator packages so the registry republishes them.


assets/evaluators/builtin/coherence/evaluator/coherence.prompty
```diff
@@ -97,4 +92,9 @@ Your output must be a valid JSON object with exactly these keys:
       - "completed": evaluation was performed normally.
       - "skipped": evaluation was not performed because the QUERY or RESPONSE is empty or not provided. When skipped, set score to null.
 
-# Output
\ No newline at end of file
+# Output
```
AliMahmoudzadeh: I suggest we move "#Outout" to the end of the prompt. so, the flow would become: task, data, output. it should only move a couple of tokens , so, should still trigger cache.

AliMahmoudzadeh (approved):

bonanza89:
Superseded by #5276 (same changes, moved from fork to branch per repo policy).

PR #5275, opened 2026-08-10
Remove ends-with-user skip condition from CSAT and task_completion multi-turn prompts

AliMahmoudzadeh:
## Summary
Removes the `or doesn't end with the agent response` skip gating from the multi-turn prompts of the Customer Satisfaction and Task Completion evaluators. Conversations that end with a user message are now scored instead of returning `skipped`.

This completes the change started in #5063, which already removed the equivalent code-level check (`if messages[-1]["role"] != MessageRole.ASSISTANT: raise ...`) from all five evaluators. Verified that no ends-with-user gate remains in the evaluator code paths.

## Changes
- `customer_satisfaction_multi_turn.prompty`: skip only when CONVERSATION is empty or not provided
- `task_completion_multi_turn.prompty`: skip only when CONVERSATION is empty or not provided
- Bumped `customer_satisfaction` spec version 14 -> 15
- Bumped `task_completion` spec version 18 -> 19

## Notes
coherence, task_adherence, and groundedness prompts never had this condition and only skip when there is no assistant response at all, so they are unchanged.


assets/evaluators/builtin/task_completion/spec.yaml
```diff
@@ -1,6 +1,6 @@
 type: "evaluator"
 name: "builtin.task_completion"
-version: 17
+version: 19
```
salma-elshafey, line 3: Why are the versions incremented by 2?

vizhur (dismissed):

assets/evaluators/builtin/task_completion/spec.yaml
AliMahmoudzadeh: it shows as one to me. 18 to 19

salma-elshafey (approved):

imatiach-msft (approved):

PR #5282, opened 2026-08-12
Add the Tool Use Quality and the Conversation Quality Meta Evaluators

salma-elshafey:
**Summary**
Adds the ConversationQualityEvaluators and ToolUseQualityEvaluators meta-evaluators.

**1. Conversation quality**
- Batches six evaluators into a single LLM call:
      - Fluency
      - Coherence
      - Intent Resolution
      - Task Adherence
      - Groundedness
      - Task Completion

**2. Tool-use quality**
- Batches five tool-use evaluators into a single LLM call:
      - Tool Call Accuracy
      - Tool Call Success
      - Tool Input Accuracy
      - Tool Output Utilization
      - Tool Selection

Which both:
- Support single-turn and multi-turn evaluation levels.
- Return an any-fail aggregate conversation_quality/tool_use_quality results.
- Preserve raw results from each member evaluator.

### DIFF
--- a/assets/benchmarkspecs/builtin/bigbenchhard/spec.yaml
+++ b/assets/benchmarkspecs/builtin/bigbenchhard/spec.yaml
@@ -1,6 +1,6 @@
 type: "benchmarkspec"
 name: "builtin.bigbenchhard"
-version: 1
+version: 2
 display_name: "BIG-Bench Hard Benchmark"
 description: "BIG-Bench Hard (BBH) is a suite of 23 challenging tasks from the BIG-Bench benchmark where prior language models did not outperform average human performance. It evaluates diverse reasoning capabilities including algorithmic, causal, logical, spatial, and temporal reasoning, as well as language understanding and common-sense inference. This is a 1k sample, generated using IRT method."
 benchmarkType: "builtin"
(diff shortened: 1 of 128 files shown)


lykelly19 (approved):

AliMahmoudzadeh (approved):
These are two comments of the aggregation and typing. we can address them in version 2

1. Missing evaluator output can falsely pass the aggregate
 quality_composite/_quality_composite.py:653-665 
 tool_use_composite/_tool_use_composite.py:893-906 
Missing or malformed evaluator objects become  status="skipped" , while the aggregate passes if all returned scores pass. A truncated response containing one passing result could therefore mark the whole composite as passed. Missing/invalid required outputs should fail execution or make the aggregate indeterminate.
2. Published threshold schemas don’t match the implementation
 quality_composite/spec.yaml:20-24 
 tool_use_composite/spec.yaml:20-24 
The specs expose  threshold  as a number, but the constructors only use dictionary-based, per-evaluator thresholds. Accepted scalar values are silently ignored. Additionally, tool-use caps the schema value at  1 , despite  tool_call_accuracy  using a 1–5 scale. The schemas should expose an object matching the constructor contract, or scalar threshold behavior should be implemented.

m7md7sien (approved):

PR #5069, opened 2026-05-20
acft-medimageparse-3d-finetune: bump marshmallow + azure-ai-ml to cle…

sidkiblawi:
marshmallow 3.23.2 -> 3.26.2 
3.26.2 is the lowest 3.x release with the GHSA-428g-f7cq-pgp5 / CVE-2025-68480 fix. The previous pin (3.23.2) was the last 3.x to expose `marshmallow.fields._T`, which azure-ai-ml <=1.22.x imported from a private path. azure-ai-ml >=1.26.x no longer needs `_T`, so the joint pin
(azure-ai-ml 1.26.5 + marshmallow 3.26.2) clears the DoS finding.

azure-ai-ml 1.16.1 -> 1.26.5: required because marshmallow >=3.24 removed the private `marshmallow.fields._T` typevar that older azure-ai-ml (<=1.22) imports. 1.26.5 is the minimum azure-ai-ml that does not need _T (verified via binary search).

Dockerfile sanity-check updated: drop the now-impossible `from marshmallow.fields import _T` import and surface marshmallow / azure-ai-ml versions on successful build.

Validated:
- pip resolution clean (no broken requirements)
- vcm image vulnerabilities show -> {"vulnerabilities": []} (was 1 MED finding on marshmallow CVE-2025-68480)
- AML finetune run on hls-NDA100-ncus-oor-gpu entered training successfully using the rebuilt image

PR #5321, opened 2026-09-08
acft-medimageparse-3d-finetune: fix fine-tuned checkpoint filename for deployment + azure-ai-ml 1.33.0

sidkiblawi:
Two changes to the MedImageParse3D finetune asset. Revives #5076 (auto-closed for inactivity) and adds the deployment fix found while validating the finetune -> deploy path.

## Fine-tuned checkpoint filename (component `0.0.1` -> `0.0.2`)

The finetune entrypoint wrote the converted weights to
`artifacts/checkpoints/boltzformer_focal_all.safetensors` - the filename used by the **2D**
MedImageParse wrapper this entrypoint was cloned from. The **3D** base model expects a different
name: `python_model.pkl` loads the vision weights by
`vision_model_name = "biomedparse_3D_AllData_MultiView_edge.safetensors"` (and `copy_catalog()`
reuses that wrapper verbatim), and the 3D checkpoint loader config points at the same path. The
expected file was therefore absent from the fine-tuned model, so it loaded fine in training but
failed at online-deployment container start. The entrypoint now writes the expected filename;
component bumped `0.0.1` -> `0.0.2` so the corrected code republishes.

### Validation
- Base model catalog confirmed: `python_model.pkl` -> `vision_model_name = biomedparse_3D_AllData_MultiView_edge.safetensors`; 3D checkpoint loader (`biomedparse_checkpoint_loader.yaml`) loads from the same path
- Reproduced on a full finetune run (real AMOS abdomen CT, 3 volumes -> 122 slices, `hls-4xA100`, job `upbeat_gyro_qzltblnj9d`): the registered model contained only `boltzformer_focal_all.safetensors`, with `biomedparse_3D_AllData_MultiView_edge.safetensors` missing
- With the fix the entrypoint emits the expected filename, so the registered model matches the wrapper and loads without a post-hoc rename

### Diff
```diff
-                / "boltzformer_focal_all.safetensors"
+                / "biomedparse_3D_AllData_MultiView_edge.safetensors"
```

## azure-ai-ml 1.33.0 (revival of #5076)

Follow-up to #5069 (marshmallow CVE-2025-68480 fix), which pinned `azure-ai-ml` to the minimum
version compatible with `marshmallow>=3.24` (1.26.5). Bumping to the latest stable so we don't
sit on a stale SDK unnecessarily. `azure-ai-ml 1.33.0` requires `marshmallow>=3.5,<4.0`, so the
existing `marshmallow==3.26.2` pin remains compatible.

### Validation
- `pip` resolution clean (no broken requirements)
- Local build pushed to staging ACR; `vcm image vulnerabilities show` -> **0 findings**
- AML finetune smoke run on `hls-NDA100-ncus-oor-gpu` against the rebuilt image entered training successfully (`frank_tangelo_7vs8tx0h1x`)

### DIFF
--- a/assets/training/finetune_acft_image/components/finetune/medimage_parse_3d/spec.yaml
+++ b/assets/training/finetune_acft_image/components/finetune/medimage_parse_3d/spec.yaml
@@ -1,7 +1,7 @@
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
 
 name: medimageparse_3d_finetune
-version: 0.0.1
+version: 0.0.2
 
 
 type: command
--- a/assets/training/finetune_acft_image/environments/acft_image_medimageparse_3d_finetune/context/requirements.txt
+++ b/assets/training/finetune_acft_image/environments/acft_image_medimageparse_3d_finetune/context/requirements.txt
@@ -9,7 +9,7 @@ hydra-core==1.3.2
 marshmallow==3.26.2
 
 # Azure ML SDK
-azure-ai-ml==1.26.5
+azure-ai-ml==1.33.0
 azure-identity==1.16.1
 azureml-mlflow==1.62.0.post2
 azureml-dataprep==5.4.1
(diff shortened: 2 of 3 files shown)


sidkiblawi:
Superseded by #5322, which branches the checkpoint fix cleanly off main (this one was incorrectly stacked on the azure-ai-ml environment commit).

PR #5336, opened 2026-09-16
fix(benchmark): use native regex grader for TruthfulQA

imatiach-msft:
## Summary

- Publish TruthfulQA benchmark version 4 using the native `azure_ai_evaluator` path with `builtin.regex_match` version 4 instead of the OpenAI `string_check` grader.
- Preserve exact, case-sensitive, whole-response matching against `Best_Answer` using `\A{{ground_truth}}\Z`. The native evaluator escapes the reference answer before substitution.
- Keep the dataset, generation prompt, and inference parameters unchanged.
- Add regression coverage that loads the benchmark spec and executes the real native evaluator, covering incorrect answers, case, whitespace, trailing newlines, prefixes, regex metacharacters, and multiline answers.

## Motivation

Avoid the OpenAI grader execution dependency for a deterministic equality check. This does not remove AOAI calls used to generate benchmark responses. Results use the native `regex_match` metric rather than `string_check`; existing version-3 groups are unchanged.

## Validation

- All 44 regex evaluator behavior tests passed, including 10 new TruthfulQA cases.
- No live run against the unpublished benchmark version has been performed. Release requires `builtin.regex_match` version 4 to be available in the target registry.

Authored-by: GitHub Copilot CLI v1.0.84-3
Model: GPT-6 Astra (gpt-6-astra)


imatiach-msft:
Replaced by #5337 from an upstream Azure/azureml-assets branch, satisfying the repository's no-fork policy without modifying or bypassing any checks. The replacement also fixes the missing test docstring identified by the three-model review and confirmed by CI.

_This comment was generated by an AI assistant on behalf of the PR author._

PR #5266, opened 2026-08-06
fix: remediate Training image vulnerabilities

eswar1996b:
## Summary
- remediate reported vulnerabilities across seven Training environment images
- upgrade inherited Ubuntu PAM/glibc packages until the shared base image is patched
- upgrade inherited base-Miniconda msgpack to 1.2.1 or newer for GHSA-6v7p-g79w-8964
- preserve only build inputs; exclude vulnerability tracking, SBOM, scan, and rendered artifacts

## Validation
- focused asset validation completed with zero errors for all changed environments
- Dockerfile whitespace and staged-path audits passed
- PR image CI will rebuild, test, and Trivy-scan the changed images

PR #5100, opened 2026-06-01
Pin DNN forecasting setuptools for pkg_resources

sharvin2187:
## Summary
- pin `setuptools==81.0.0` in `ai-ml-automl-dnn-forecasting-gpu`
- keep `pkg_resources` available for AzureML packages that still import it during AutoML DNN forecasting training
- remove stale vulnerability override comment for `setuptools>=82.0.1`

## Validation
- confirmed `setuptools==81.0.0` exists on PyPI and includes `pkg_resources`
- reviewed Dockerfile diff

PR #5340, opened 2026-09-18
fix: temporarily restore forecasting pkg_resources compatibility (security exception required)

lusu-msft:
## Security approval required before merge or release

**DRAFT: This intentionally rolls the forecasting GPU environment back from `setuptools>=83.0.0` to `setuptools==81.0.0`. It leaves [GHSA-h35f-9h28-mq5c](https://github.com/advisories/GHSA-h35f-9h28-mq5c) unremediated. Explicit security-exception approval is required before merging or releasing this image. Preparing this draft is not approval of that exception.**

The advisory affects setuptools versions below 83.0.0: Unicode NFC/NFD differences can cause `MANIFEST.in` exclusions to fail when building a source distribution, potentially publishing files intended to be excluded. The documented exposure is particularly relevant to macOS APFS/HFS+. This Linux training image has a different use case, but this PR does not claim the advisory is inapplicable or resolved.

## Compatibility problem

#5100 restored setuptools 81 for AutoML's `pkg_resources` dependency. #5266 subsequently raised the minimum to 83 for security remediation. Setuptools 82 removed `pkg_resources`, while the latest published packages still import it:

- `azureml-automl-runtime==1.62.0.post1`: `azureml/automl/runtime/faults_verifier.py:10`, with `resource_filename` at line 157.
- `azureml-contrib-automl-dnn-forecasting==1.62.0`: `azureml/contrib/automl/dnn/forecasting/callbacks/_run_update_base.py:13`, with `resource_filename` at line 241.

Upgrading TensorBoard alone cannot remove these SDK imports.

## Changes

- Restore `setuptools==81.0.0` in both the initial install and subsequent constraints for `ai-ml-automl-dnn-forecasting-gpu` only.
- Document that this is a temporary compatibility exception, not a security floor.
- Remove the three setuptools-only `vendor.txt` substitutions that labeled vendored 70.3.0 as 83.0.0 without changing vendored code. Do not relabel it as 81 either. Other existing security fixes and metadata operations are unchanged.
- Extend the existing environment smoke job to import both affected AutoML modules before sklearn training. Import failures propagate and fail the job rather than being caught or skipped.

## Validation and limitations

- Existing targeted asset validator: 0 errors for 1 asset; existing warning that the environment name is missing a framework.
- Changed Python files parse successfully; diff whitespace checked with CRLF-aware settings matching the existing Dockerfile.
- Public wheel SHA256 values matched PyPI metadata. In isolated local Python 3.11, executing the unmodified published `faults_verifier.py` with setuptools 83 fails at line 10 with `ModuleNotFoundError: No module named 'pkg_resources'`. With unpacked setuptools 81, that import succeeds and execution advances to the next unavailable dependency (`pandas`). This is a narrow import reproduction, not a successful full SDK import or training run.
- Full Python 3.10 Linux image build, new Azure smoke job, GPU forecasting training, and vulnerability scan have **not** been run locally. Docker CLI is installed but its Linux daemon is unavailable. This PR must remain draft pending those results and security approval; other dependency incompatibilities may still exist.

## Required follow-up

- [ ] Obtain explicit security-exception approval covering the setuptools rollback before merge/release.
- [ ] Build and scan the complete image and run the extended environment smoke job.
- [ ] Run representative AutoML DNN forecasting training before release.
- [ ] Publish compatible SDK patches migrating both packages' resource access to `importlib.resources`; verify remaining transitive dependencies no longer require `pkg_resources`, then restore the setuptools security floor and retire the exception.


lusu-msft:
All three failed checks (assets-test, environments-ci, scripts-test) stop in the shared execution-context workflow at "Check PR labels if forked", before any image build or test executes. This fork PR has no `safe to test` label. Please review the patch and, if appropriate, add the existing `safe to test` label to trigger the supported fork CI path. The contributing account does not have upstream write/triage permissions. No workflow bypass is proposed. CI authorization is separate from the explicit GHSA-h35f-9h28-mq5c security-exception approval still required before merge/release; please keep this PR draft.

